PR #3, opened 2023-04-12
sysdetect: Add cuda header path to Rules file

anustuvicl:
This fixes configuring PAPI without cuda component. If 
`$ ./configure && make`
then `components/sysdetect/nvidia_gpu.o` complains that `"cuda.h"` not found.

If `./configure --with-components=cuda && make`
then the error goes away.

This PR fixes the above problem by adding include path to cuda header.


anustuvicl:
Closing. This was fixed by [https://bitbucket.org/icl/papi/pull-requests/458](https://bitbucket.org/icl/papi/pull-requests/458)

PR #4, opened 2023-04-12
papi.c: Fix documentation for PAPI_add_named_event

anustuvicl:
## Documentation
-	PAPI_event_code_to_name mentions PAPI_MAX_STR_LEN.
-	PAPI_overflow states args are EventSet, EventCode, threshold, flags, handler. The handler arg is a callback function that takes address, overflow_vector, *context (According to documentation).
But the example code shows that handler function takes EventSet and then the other three. **Updated**
-	PAPI_get_executable_info: Documentation says params fullname, name, text_start, text_end, data_start, data_end, bss_start, bss_end. Docs also says the return value could be PAPI_EINVAL. **Updated**
The actual function definition says it takes no args, and returns struct `PAPI_exe_info_t *`. The params described are members of this struct. How can this function return an an error code?
-	PAPI_lock, PAPI_unlock doc says user should pass an integer (PAPI_USR1_LOCK or PAPI_USR2_lock). What if he passing some other number >=0 and <= PAPI_MAX_LOCK?
o	Also says doesn’t return anything, but code can return PAPI_EINVAL
o	Added `@bug` to PAPI_lock documentation
-	`PAPI_set_debug`: mentions it is only used if the library was compiled with -DDEBUG. **Updated to** if the library was configure with `--with-debug=yes`?
o	It also says in a note: This is the only function that may be called BEFORE PAPI_library_init(). This is not true. PAPI_disable_component() can only be called before PAPI_library_init().
-	PAPI_get_virt/real_usec/nsec/cyc have example code copied across.
-	4 instances of `\bug none known` in docs. These are not shown in man pages, but shown in html. There are 15 occurrences of `@bug` not shown in man pages.
-	In most examples if blocks are not indented.
-	PAPI_get_multiplex: check FORTRAN code interface.
-	PAPI_add_event & PAPI_add_named_event: Removed `@bug` (same as PAPI_add_events)
-	PAPI_num_hwctrs – marked as deprecated, should’ve been removed in 7.0.0.0
-	Fixed examples: PAPI_get_real_usec, PAPI_get_virt_usec, 
-	Fixed example: PAPI_get_component_index

*Added all tracked files in repo including updated & built manpage files*

## To review
-	PAPI_get_cmp_opt says see table below. **Where is the table?**
-	All functions documented as `\class`, should be `\fn`.
-	PAPI_accum: Possible mistake when it says PAPI_accum resets values array to zero.
- Should add `@bug` to `PAPI_event_code_to_name` documentation.

### DIFF
--- a/src/papi.c
+++ b/src/papi.c
@@ -145,8 +145,8 @@ _internal_check_rate_state()
   *	@brief Simplified call to get Mflips/s (floating point instruction rate), real and processor time. 
   *
   *	@par C Interface: 
-  *	\#include <papi.h> @n
-  *	int PAPI_flips_rate( int event, float *rtime, float *ptime, long long *flpins, float *mflips );
+  *	`#include <papi.h>` @n
+  *	`int PAPI_flips_rate( int event, float *rtime, float *ptime, long long *flpins, float *mflips );`
   *
   *   @param event
   *     one of the three presets PAPI_FP_INS, PAPI_VEC_SP or PAPI_VEC_DP 
@@ -209,8 +209,8 @@ PAPI_flips_rate( int event, float *rtime, float *ptime, long long *flpins, float
   *	@brief Simplified call to get Mflops/s (floating point operation rate), real and processor time. 
   *
   *	@par C Interface: 
-  *	\#include <papi.h> @n
-  *	int PAPI_flops_rate ( int event, float *rtime, float *ptime, long long *flpops, float *mflops );
+  *	`#include <papi.h>` @n
+  *	`int PAPI_flops_rate ( int event, float *rtime, float *ptime, long long *flpops, float *mflops );`
   *
   *   @param event
   *     one of the three presets PAPI_FP_OPS, PAPI_SP_OPS or PAPI_DP_OPS
@@ -274,8 +274,8 @@ PAPI_flops_rate( int event, float *rtime, float *ptime, long long *flpops, float
   *	@brief Simplified call to get instructions per cycle, real and processor time. 
   *
   *	@par C Interface: 
-  *	\#include <papi.h> @n
-  *	int PAPI_ipc( float *rtime, float *ptime, long long *ins, float *ipc );
+  *	`#include <papi.h>` @n
+  *	`int PAPI_ipc( float *rtime, float *ptime, long long *ins, float *ipc );`
   *
   * @param *rtime
   *		realtime since the latest call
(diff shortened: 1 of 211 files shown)


src/papi.c
```diff
@@ -655,11 +655,16 @@ valid_ESI_component( EventSetInfo_t * ESI )
...
  *
  *	@code
-if ( PAPI_thread_init(pthread_self) != PAPI_OK )
-	exit(1);
+ *  if ( PAPI_thread_init(pthread_self) != PAPI_OK )
+ *  	exit(1);
  *	@endcode
  *
- *	@see PAPI_register_thread PAPI_unregister_thread PAPI_get_thr_specific PAPI_set_thr_specific PAPI_thread_id PAPI_list_threads
+ *	@see PAPI_register_thread
```
gcongiu, line 662: unneeded indentation?

src/papi.c
```diff
@@ -879,27 +884,30 @@ PAPI_list_threads( PAPI_thread_id_t *tids, int *number )
...
+ *     if (state == NULL) return (PAPI_ESYS);
+ *     memset(state, 0, sizeof(RateInfo));
+ *     state->EventSet = PAPI_NULL;
+ *     ret = PAPI_create_eventset(&state->EventSet);
+ *     if (ret != PAPI_OK) return (PAPI_ESYS);
+ *     ret = PAPI_set_thr_specific(PAPI_USR1_TLS, state);
+ *     if (ret != PAPI_OK) return (ret);
+ * }
+ *	@endcode
+ *	@see PAPI_register_thread
```
gcongiu, line 906: unneeded indentation?

src/papi.c
```diff
@@ -955,26 +963,29 @@ PAPI_get_thr_specific( int tag, void **ptr )
...
+ *     if (ret != PAPI_OK) return (PAPI_ESYS);
+ *     ret = PAPI_set_thr_specific(PAPI_USR1_TLS, state);
+ *     if (ret != PAPI_OK) return (ret);
+ * }
  *	@endcode
- *	@see PAPI_register_thread PAPI_thread_init PAPI_thread_id PAPI_get_thr_specific
+ *	@see PAPI_register_thread @n 
+ *  PAPI_thread_init @n 
+ *  PAPI_thread_id @n 
+ *  PAPI_get_thr_specific
```
gcongiu, line 988: `@see PAPI_thread_init ...`?

src/papi.c
```diff
@@ -6456,36 +6534,40 @@ PAPI_add_events( int EventSet, int *Events, int number )
...
  *  to a real error code.
  *
- *	@see PAPI_cleanup_eventset PAPI_destroy_eventset PAPI_event_name_to_code 
- *		PAPI_presets PAPI_add_event PAPI_add_events
+ *	@see PAPI_cleanup_eventset @n 
+ *  PAPI_destroy_eventset @n 
+ *  PAPI_event_name_to_code @n 
+ *	PAPI_presets @n 
+ *  PAPI_add_event @n 
+ *  PAPI_add_events
```
gcongiu, line 6570: Is `@see ...` needed or `@n` accounts for it?

src/papi.c
anustuvicl: All the tags must be at the same level of indentation for it to render correctly. Then I will have to dedent all the tags down.

PR #12, opened 2023-05-27
cuda: New component based on NVIDIA PerfWorks API.

anustuvicl:


src/components/cuda/cupti_profiler.c
```diff
@@ -0,0 +1,1576 @@
...
+            .structSize = NVPW_MetricsContext_GetMetricNames_End_Params_STRUCT_SIZE,
+            .pPriv = NULL,
+            .pMetricsContext = avail_events[gpu_id].metricsContextCreateParams.pMetricsContext,
+        };
+        NVPW_CALL(NVPW_MetricsContext_GetMetricNames_EndPtr((NVPW_MetricsContext_GetMetricNames_End_Params *) &getMetricNameEndParams), return PAPI_EMISC);
+
+    }
+    TOCK;
+    TIMEDBG("Time to get all metric names =");
+    TICK;
```
gcongiu, line 1187: What are those `TOCK`/`TICK`?

src/components/cuda/cupti_profiler.c
```diff
@@ -0,0 +1,1576 @@
...
+                goto fn_exit;
+            }
+            if (find_event_name(all_evt_names, evt_name, &find) == PAPI_ENOEVNT) {
+                errno = insert_event_record(all_evt_names, evt_name, i, 0);
+                if (errno != PAPI_OK) {
+                    goto fn_exit;
+                }
+            }
+            // LOGDBG("Name:\t\t%s\nDescription:\t%s\n\n",
+            //    all_evt_names->evts[all_evt_names->count-1].name, avail_events[gpu_id].nv_metrics->evts[i].desc);
```
gcongiu, line 1206: Remove commented code

src/components/cuda/cupti_profiler.c
```diff
@@ -0,0 +1,1576 @@
...
+        }
+        if (avail_events[gpu_id].nv_metrics)
+            free_event_name_list( &(avail_events[gpu_id].nv_metrics) );
+    }
+    papi_free(avail_events);
+    avail_events = NULL;
+}
+
+// CUPTI Profiler component API functions
+int cuptip_init(const char **pdisabled_reason)
```
gcongiu, line 1294: Can you return the disabled reason using another function? Something like `cuptip_error_get_last`?

src/components/cuda/cupti_profiler.c
```diff
@@ -0,0 +1,1576 @@
...
+    if (errno != CUDA_SUCCESS) {
+        *pdisabled_reason = "Failed to initialize CUDA driver API.";
+        goto fn_fail;
+    }
+    return PAPI_OK;
+fn_fail:
+    return PAPI_EMISC;
+}
+
+int cuptip_control_create(event_list_t *all_event_names, int event_count, int *evt_ids, void **pctl, void **pcu_ctx)
```
gcongiu, line 1330: Why do you pass event ids as well as event names?

src/components/cuda/cupti_profiler.c
```diff
@@ -0,0 +1,1576 @@
...
+                if (errno != PAPI_OK) {
+                    goto fn_exit;
+                }
+            }
+            // LOGDBG("Name:\t\t%s\nDescription:\t%s\n\n",
+            //    all_evt_names->evts[all_evt_names->count-1].name, avail_events[gpu_id].nv_metrics->evts[i].desc);
+        }
+    }
+    TOCK;
+    TIMEDBG("Time to transfer all metric names to component =");
```
gcongiu, line 1210: I feel you have too many log messages.

src/components/cuda/linux-cuda.c
```diff
@@ -1,5651 +1,442 @@
 /**
  * @file    linux-cuda.c
- * @author  Tony Castaldo tonycastaldo@icl.utk.edu (updated in 08/2019, to make counters accumulate.)
- * @author  Tony Castaldo tonycastaldo@icl.utk.edu (updated in 2018, to use batch reads and support nvlink metrics.
- * @author  Asim YarKhan yarkhan@icl.utk.edu (updated in 2017 to support CUDA metrics)
- * @author  Asim YarKhan yarkhan@icl.utk.edu (updated in 2015 for multiple CUDA contexts/devices)
- * @author  Heike Jagode (First version, in collaboration with Robert Dietrich, TU Dresden) jagode@icl.utk.edu
```
gcongiu, line 7: You should not remove others people contributions from this file.

src/components/cuda/lcuda_utils.h
gcongiu, line 1: These are functions common to event and profiler APIs. Maybe `cupti_common.h` would be a better name for this file?

src/components/cuda/linux-cuda.c
```diff
@@ -1,5651 +1,442 @@
...
-        // Here we init the values for this event.
-        switch (cuda11_AllEvents[idx]->treatment) {
-            case SpotValue: cuda11_AllEvents[idx]->cumulativeValue = 0; break;
-            case RunningSum: cuda11_AllEvents[idx]->cumulativeValue = 0; break;
-            case RunningMin: cuda11_AllEvents[idx]->cumulativeValue = DBL_MAX; break;
-            case RunningMax: cuda11_AllEvents[idx]->cumulativeValue = -DBL_MAX; break;
-        }
-
-        // Get the device we need to count it towards.                
-        int eventDeviceNum = c
```
gcongiu, line 315: Unecessary logging of information

gcongiu, line 316: Locking here might cause too coarse granularity. Maybe locking inside the backend code is better.

gcongiu, line 393: I feel like logging every single action is not very useful and only adds unnecessary code

gcongiu (changes requested):
This is a first round review

src/components/cuda/cupti_profiler.c
```diff
@@ -0,0 +1,1576 @@
...
+    return PAPI_OK;
+}
+
+int cuptip_enumerate_all_metric_names(event_list_t *all_evt_names)
+{
+    int gpu_id, i, found, listsubmetrics=1, errno = PAPI_OK;
+    if (avail_events[0].nv_metrics != NULL)  // Already eumerated for 1st device? Then exit...
+        goto fn_exit;
+    STOPWATCH;
+    TICK;
```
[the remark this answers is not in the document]
gulaki: Macros inside `lcuda_debug.h` that allow Matlab like stopwatch functionality.

src/components/cuda/linux-cuda.c
```diff
@@ -1,5651 +1,442 @@
...
-#include <limits.h>
-#include <float.h> // For DBL_MAX. 
-#include <sys/stat.h>
-#include <dirent.h>
+#include <string.h>
 
-// NOTE: We can't use extended directories; these include files have includes.
-#include <cupti.h>
-#include <cuda_runtime_api.h>
+#include "lcuda_common.h"
```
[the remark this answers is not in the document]
gulaki: Because it's being used.

src/components/cuda/linux-cuda.c
gulaki: ?

gulaki: ?

gulaki: ?

src/components/cuda/linux-cuda.c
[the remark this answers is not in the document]
gcongiu: Why is it being used?

src/components/cuda/linux-cuda.c
gcongiu: I think I might have commented a stale commit. Disregard this

gcongiu: I think I might have commented a stale commit. Disregard this

src/components/cuda/cupti_profiler.c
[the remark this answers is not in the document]
gcongiu: Get rid of them please

src/components/cuda/lcuda_dispatch.c
```diff
@@ -0,0 +1,232 @@
...
+
+    } else {
+        *pdisabled_reason = "CUDA configuration not supported.";
+        papi_errno = PAPI_ECMP;
+    }
+fn_exit:
+    return papi_errno;
+}
+
+int cuptid_thread_info_init(void **thread_info)
```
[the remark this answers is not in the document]
gcongiu: I would also say replace init with create and add an update function as well.

src/components/cuda/lcuda_debug.h
```diff
@@ -0,0 +1,66 @@
...
+#ifdef TIME_MEASUREMENT
+#    include <time.h>
+#    undef STOPWATCH
+#    undef TICK
+#    undef TOCK
+#    undef TIMEDBG
+#    define STOPWATCH clock_t _tick, _tock, _tdiff;
+#    define TICK _tick = clock();
+#    define TOCK _tock = clock(); _tdiff = _tock - _tick;
+#    define TIMEDBG(...) SUBDBG("TIMEDEBUG: " __VA_ARGS__ "%f s.\n", (double) _tdiff / CLOCKS_PER_SEC);
```
gcongiu, line 63: We don't need these do we?

gcongiu (changes requested):
Another set of comments

anustuvicl:
Pushed 3 commits with changes reflected by thumbs up to @gcongiu comments.

src/components/cuda/cupti_events.c
```diff
@@ -0,0 +1,65 @@
...
+{
+    return PAPI_ENOIMPL;
+}
+
+int cuptie_stop(void **pctl, void **pcu_ctx)
+{
+    return PAPI_ENOIMPL;
+}
+
+int cuptie_control_read(void **pctl, long long *values)
```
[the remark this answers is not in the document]
anustuvicl: `cupti_[e|p]_control_destroy` also takes double pointer as it sets the pointer to `NULL` after freeing memory. Others have been changed to take single pointer.

src/components/cuda/cupti_profiler.c
```diff
@@ -0,0 +1,1576 @@
...
+        }
+        papi_free(state->ctl[i].rmr);
+    }
+    papi_free(state->ctl);
+    papi_free(state);
+    *pctl = NULL;
+    return PAPI_OK;
+}
+
+int cuptip_start(void **pctl, void **pcu_ctx)
```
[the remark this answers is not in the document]
anustuvicl: `cuptip_start` has to be able to access the user set contexts for each gpu.

src/components/cuda/cupti_profiler.c
anustuvicl: `disabled_reason` is set at multiple points. How would one function accomplish this?

src/components/cuda/cupti_profiler.c
anustuvicl: `all_event_names` is the component global event table, not only the events being added. The `evt_ids` contains the indices of the entries in the global event table to be added.

src/components/cuda/linux-cuda.c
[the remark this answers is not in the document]
anustuvicl: `lcuda_common.h` contains API for dynamic event table. It is used by `linux-cuda.c` as well as `cupti_profiler.c`. `cupti_events.c` should also use this.

src/components/cuda/lcuda_config.h
```diff
@@ -0,0 +1,23 @@
...
+ *          anustuv@icl.utk.edu
+ */
+
+#ifndef __LCUDA_CONFIG_H__
+#define __LCUDA_CONFIG_H__
+
+#include <cupti.h>
+
+#define CUPTI_PROFILER_API_MIN_SUPPORTED_VERSION  (13)
+// #define CUPTI_EVENTS_API_MAX_SUPPORTED_VERSION (xx)  // TODO: Set to last version when CUPTI events API removed
```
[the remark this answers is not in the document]
anustuvicl: This was requested by @jagode. It's a guide for future developers. We should discuss this further.

src/components/cuda/lcuda_dispatch.c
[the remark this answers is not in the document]
anustuvicl: I dont think `attr_init` is a good name for this. Cuda contexts are threading specific functionality. If we dont want cuda context in the name, thread_info is the best alternative.

src/components/cuda/cupti_profiler.c
```diff
@@ -0,0 +1,1576 @@
...
+    NVPW_RawMetricsConfig_IsAddMetricsPossiblePtr = DLSYM_AND_CHECK(dl_nvpw, "NVPW_RawMetricsConfig_IsAddMetricsPossible");
+    NVPW_MetricsContext_GetCounterNames_BeginPtr = DLSYM_AND_CHECK(dl_nvpw, "NVPW_MetricsContext_GetCounterNames_Begin");
+    NVPW_MetricsContext_GetCounterNames_EndPtr = DLSYM_AND_CHECK(dl_nvpw, "NVPW_MetricsContext_GetCounterNames_End");
+
+    Dl_info info;
+    dladdr(NVPW_GetSupportedChipNamesPtr, &info);
+    LOGDBG("NVPW library loaded from %s\n", info.dli_fname);
+    return PAPI_OK;
+fn_fail:
+    return PAPI_ESYS;
```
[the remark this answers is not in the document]
anustuvicl: `PAPI_ESYS` is appropriate here as this function will return this error only `dlopen` or `dlsym` fails which are C system functions. This was verified by @jagode

src/components/cuda/cupti_profiler.c
anustuvicl: There should be enough log messages, so that if something fails for the user, the log messages tell us where and what the problem was, and how we can fix it.

src/components/cuda/linux-cuda.c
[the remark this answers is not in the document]
gcongiu: Sharing the a global event table between frontend and backend is not a good idea. The only interaction between frontend and backend should be the main interface.

src/components/cuda/lcuda_dispatch.c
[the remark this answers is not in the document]
gcongiu: My point is this is control related information. Control information is thread specific.

src/components/cuda/linux-cuda.c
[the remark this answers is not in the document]
anustuvicl: Have a look at the API you will understand. This header is not itself responsible for passing the global table to the backend. It only contains the data struct and accompanying functions to support dynamic event table creation, deletion, insertion, and search. That facility is shared by both the layers.

In case of enumeration, the frontend has to pass the pointer to the global event table to the backend, right? As in
`cuptid_event_enum(global_event_table)`. How else will the global table be populated?

I can change it so that we are not passing `global_event_table` and `event_ids` to `cuptid_control_create`.
Is that something you prefer?

anustuvicl:
This commit adds a function `select_by_idx` that is used by `cuda_update_control_state` to select the names from the `global_event_names` to add to the control state of a thread. Thus, it avoids passing the global table to the backend. `cuptid_control_create` now takes a copy of events that it needs to add and no event_ids array.

src/components/cuda/lcuda_dispatch.c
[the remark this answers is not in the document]
anustuvicl: Sure control state is already thread specific, but why can it not possess information related to that particular thread? Is there any limitation on that?

src/components/cuda/cupti_profiler.c
anustuvicl: this is fixed in commit https://github.com/icl-utk-edu/papi/pull/12/commits/eca91a2a7f6be7de66724fafee9eef2016ca8b38

src/components/cuda/lcuda_dispatch.c
```diff
@@ -0,0 +1,232 @@
...
+        goto fn_exit;
+    }
+
+    if (util_gpu_collection_kind() == GPU_COLLECTION_MIXED) {
+        *pdisabled_reason = "No support for systems with mixed compute capabilities, such as CC < 7.0 and CC > 7.0 GPUS.";
+        papi_errno = PAPI_ECMP;
+        goto fn_exit;
+    }
+
+    if (util_runtime_is_perfworks_api()) {
```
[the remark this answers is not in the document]
anustuvicl: Have a look at the `else` condition. If user system has gpus of CC > 7.0 and he loads cuda toolkit version 9.2 (that does not have perfworks), then at runtime it wont satisfy either profiler API or events API. Thus two separate runtime checks are necessary.

src/components/cuda/linux-cuda.c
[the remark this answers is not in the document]
gcongiu: I guess that is ok, you are right. My reasoning was that the backend and the frontend are completely independent. They share nothing and interact only through the cupti wrapper API. However, the backend already uses definitions from PAPI, like error codes. Could you just rename `event_list_t` to `ntv_event_table_t`? List in the name reveals implementation details for the data structure. `event_rec_t` can also be just `ntv_event_t`. So that the `ntv_event_table_t` is a collection of `ntv_event_t` elements.

src/components/cuda/linux-cuda.c
```diff
@@ -1,5651 +1,459 @@
 /**
  * @file    linux-cuda.c
+ *
+ * @author  Anustuv Pal   anustuv@icl.utk.edu (updated in 2023, redesigned with multi-threading support.)
  * @author  Tony Castaldo tonycastaldo@icl.utk.edu (updated in 08/2019, to make counters accumulate.)
- * @author  Tony Castaldo tonycastaldo@icl.utk.edu (updated in 2018, to use batch reads and support nvlink metrics.
- * @author  Asim YarKhan yarkhan@icl.utk.edu (updated in 2017 to support CUDA metrics)
- * @author  Asim YarKhan yarkhan@icl.utk.edu (updated in 2015 for multiple CUDA contexts/devices)
- * @author  Heike Jagode (First version, in collaboration with Robert Dietrich, TU Dresden) jagode@icl.utk.edu
+ * @author  Tony Castaldo tonycastaldo@icl.utk.edu (updated in 2018, to use batch reads and support nvlink metrics.)
```
gcongiu, line 6: Remove `)`

src/components/cuda/cupti_common.h
gcongiu, line 1: The name of the file should also suggest the name of the API. Could you namespace cupti common interfaces as `cupti_common_`? Same would apply to other files.

src/components/cuda/cupti_events.c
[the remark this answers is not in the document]
gcongiu: That is up to you

src/components/cuda/lcuda_dispatch.c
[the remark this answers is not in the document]
gcongiu: It possesses information about a particular thread like the control state in the frontend does. The point is: making explicit references to thread is superfluous and confusing because the backend deals with `control` (which is thread specific). I think using `thread` in the name is also inconsistent. If we pass that information to the control create function it makes sense to call the info object as `control_info` rather than `thread_info`.

gcongiu (changes requested):
Another set of comments.

src/components/cuda/cupti_common.h
anustuvicl: There is only 1 `cupti_common.h` API. What name space would you prefer? I think a better name is `cupti_utils.h`

anustuvicl:
`cuptid_control_create` and `cuptid_control_info_create`, these names are too close and might cause confusion. How about just `cuptid_info_create`, `cuptid_info_update` (to capture the CUcontext), and `cuptid_info_destroy` API?

anustuvicl:
Some of the basic ones I have changed to resolved as I made the appropriate changes. If you want I can make them unresolved. The ones remaining need to be resolved, awaiting your comment/review.

anustuvicl:
Updated `configure.in` and build `configure` script with added option to enable perfworks API profiling when all gpus are CC7.0 and CUDA toolkit version 11.0.

src/components/cuda/linux-cuda.c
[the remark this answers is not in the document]
anustuvicl: Another common field is `evt_pos`. It is used by the low end to transport measured values back to the correct position of the event that it is measuring for.

src/components/cuda/cupti_profiler.c
```diff
@@ -193,40 +193,46 @@ static int load_nvpw_sym(void)
...
-    int count, i, found = 0;
+    int count, i;
     char *papi_cuda_root = getenv("PAPI_CUDA_ROOT");
+
+    char lookup_path[PATH_MAX];
+    const char *standard_paths[] = {
+        "%s/extras/CUPTI/lib64/%s",
+        "%s/lib64/%s",
+    };
+    int num_standard_paths = sizeof(standard_paths) / sizeof(standard_paths[0]);
```
gcongiu: I am not sure we should have this functionality implemented in this PR. The reason is we already have an issue tracking this problem and ideally we want a PR that addresses that issue only. This commit could be made a separate PR and merged immediately after this one. The PR can also reference the issue in the body so that merging the PR automatically closes the issue as well

anustuvicl: Have a look at [my fork PR](https://github.com/anustuvicl/papi/pull/2). It implements the feature in question in a commit to be merged on top of this branch.

src/components/cuda/cupti_common.c
```diff
@@ -0,0 +1,465 @@
...
+{
+    int runtimeVersion;
+    CUDART_CALL(cudaRuntimeGetVersionPtr(&runtimeVersion), return PAPI_EMISC );
+    return runtimeVersion;
+}
+
+static int util_dylib_cupti_version(void)
+{
+    unsigned int cuptiVersion;
+    CUPTI_CALL(cuptiGetVersionPtr(&cuptiVersion), return PAPI_EMISC );
```
gcongiu, line 275: unrelated white space before `)`

src/components/cuda/cupti_common.c
```diff
@@ -0,0 +1,465 @@
...
+    return numDevs;
+}
+
+static int get_gpu_compute_capability(int dev_num)
+{
+    int cc_major, cc_minor;
+    int cc;
+    CUDART_CALL(cudaDeviceGetAttributePtr(&cc_major,
+        cudaDevAttrComputeCapabilityMajor, dev_num),
+            return PAPI_EMISC );
```
gcongiu, line 295: either have a white space after `(` and before `)` or no white space at all. Don't mix things up.

src/components/cuda/cupti_common.c
```diff
@@ -0,0 +1,465 @@
...
+    cuInitPtr                    = NULL;
+    cuGetErrorStringPtr          = NULL;
+    cuCtxPopCurrentPtr           = NULL;
+    cuCtxPushCurrentPtr          = NULL;
+    cuCtxSynchronizePtr          = NULL;
+    cuDeviceGetAttributePtr      = NULL;
+    return PAPI_OK;
+}
+
+void* load_dynamic_syms(const char *parent_path, const char *dlname, const char *search_subpaths[])
```
gcongiu, line 105: be consistent when placing the `*` (void *load_dynamic_syms).

src/components/cuda/cupti_common.c
```diff
@@ -0,0 +1,465 @@
...
+static int util_dylib_cupti_version(void)
+{
+    unsigned int cuptiVersion;
+    CUPTI_CALL(cuptiGetVersionPtr(&cuptiVersion), return PAPI_EMISC );
+    return cuptiVersion;
+}
+
+int get_device_count(void)
+{
+    static int numDevs = -1;
```
gcongiu, line 281: why `-1`?

src/components/cuda/cupti_common.c
```diff
@@ -0,0 +1,465 @@
...
+        cudaDevAttrComputeCapabilityMajor, dev_num),
+            return PAPI_EMISC );
+    CUDART_CALL(cudaDeviceGetAttributePtr(&cc_minor,
+        cudaDevAttrComputeCapabilityMinor, dev_num),
+            return PAPI_EMISC );
+    cc = cc_major * 10 + cc_minor;
+    return cc;
+}
+
+enum gpu_collection_e {GPU_COLLECTION_UNKNOWN, GPU_COLLECTION_ALL_PERF, GPU_COLLECTION_MIXED, GPU_COLLECTION_ALL_EVENTS, GPU_COLLECTION_ALL_CC70};
```
gcongiu, line 303: `_e` should be used with typedef to indicate this is an enum type. This is redundant here, please remove.

src/components/cuda/cupti_common.c
```diff
@@ -0,0 +1,465 @@
...
+
+static enum gpu_collection_e util_gpu_collection_kind(void)
```
gcongiu, line 305: Same comment as above

src/components/cuda/cupti_common.c
```diff
@@ -0,0 +1,465 @@
...
+    if (util_gpu_collection_kind() == GPU_COLLECTION_MIXED) {
+        *pdisabled_reason = "No support for systems with mixed compute capabilities, such as CC < 7.0 and CC > 7.0 GPUS.";
+        papi_errno = PAPI_ECMP;
+        goto fn_exit;
+    }
+fn_exit:
+    return papi_errno;
+}
+
+int util_runtime_is_perfworks_api(void)
```
gcongiu, line 361: all functions that are non-static should be namespaced. Use `cuptic_` or `cupti_common_` for those.

src/components/cuda/cupti_profiler.c
```diff
@@ -0,0 +1,1564 @@
...
+NVPA_Status ( *NVPW_CUDA_MetricsContext_CreatePtr ) (NVPW_CUDA_MetricsContext_Create_Params* params);
+NVPA_Status ( *NVPW_MetricsContext_DestroyPtr ) (NVPW_MetricsContext_Destroy_Params * params);
+NVPA_Status ( *NVPW_MetricsContext_GetMetricNames_BeginPtr ) (NVPW_MetricsContext_GetMetricNames_Begin_Params* params);
+NVPA_Status ( *NVPW_MetricsContext_GetMetricNames_EndPtr ) (NVPW_MetricsContext_GetMetricNames_End_Params* params);
+NVPA_Status ( *NVPW_InitializeHostPtr ) (NVPW_InitializeHost_Params* params);
+NVPA_Status ( *NVPW_MetricsContext_GetMetricProperties_BeginPtr ) (NVPW_MetricsContext_GetMetricProperties_Begin_Params* p);
+NVPA_Status ( *NVPW_MetricsContext_GetMetricProperties_EndPtr ) (NVPW_MetricsContext_GetMetricProperties_End_Params* p);
+NVPA_Status ( *NVPW_CUDA_RawMetricsConfig_CreatePtr ) (NVPW_CUDA_RawMetricsConfig_Create_Params*);
+
+NVPA_Status ( *NVPW_RawMetricsConfig_DestroyPtr ) (NVPW_RawMetricsConfig_Destroy_Params* params);
```
gcongiu, line 72: Be consistent with spaces

src/components/cuda/cupti_profiler.c
```diff
@@ -0,0 +1,1564 @@
...
+            ERRDBG("NVPA Error %d: Error in call to " #call "\n", _status);  \
+            EXIT_OR_NOT; \
+            handleerror;  \
+        }  \
+    } while (0);
+
+static int load_cupti_perf_sym(void)
+{
+    COMPDBG("Entering.\n");
+    int papiErr = PAPI_OK;
```
gcongiu, line 128: Please use `papi_errno`

src/components/cuda/cupti_profiler.c
```diff
@@ -0,0 +1,1564 @@
...
+    CUpti_Profiler_Initialize_Params profilerInitializeParams = { CUpti_Profiler_Initialize_Params_STRUCT_SIZE, NULL };
+    papi_errno = cuptiProfilerInitializePtr(&profilerInitializeParams);
+    if (papi_errno != CUPTI_SUCCESS) {
+        ERRDBG("CUPTI error %d: cuptiProfilerInitialize failed.\n", papi_errno);
+        return PAPI_EMISC;
+    }
+    return PAPI_OK;
+}
+
+static int deinitialize_cupti_profiler_api(void)
```
gcongiu, line 309: `finalize`?

src/components/cuda/cupti_profiler.c
```diff
@@ -0,0 +1,1564 @@
...
+    strcpy(chipName, getChipName.pChipName);
+    return PAPI_OK;
+}
+
+struct byte_array_s {
+    int size;
+    uint8_t *data;
+};
+
+struct cuptip_gpu_control_s {
```
gcongiu, line 358: `cuptip_gpu_control_s` and `cuptip_control_s` sounds kind of the same thing from the name. Maybe you could rename `cuptip_gpu_control_s` to something more distinctive? I don't really have a good name for that. Maybe you can think of something?

src/components/cuda/cupti_profiler.c
```diff
@@ -0,0 +1,1564 @@
...
+        .pPriv = NULL,
+        .pMetricsContext = pMetricsContext,
+    };
+    NVPW_CALL(NVPW_MetricsContext_GetMetricProperties_EndPtr(&getMetricPropertiesEndParams), return PAPI_EMISC);
+    return PAPI_OK;
+}
+
+static int get_event_names_rmr(NVPA_MetricsContext* pMetricsContext, cuptip_gpu_control_t *gpu_ctl)
+{
+    COMPDBG("Entering.\n");
```
gcongiu, line 466: This level of logging is not needed but if you really want it in you should have an exit log as well.

src/components/cuda/cupti_profiler.c
```diff
@@ -0,0 +1,1564 @@
...
+    getCounterAvailabilityParams.pCounterAvailabilityImage = gpu_ctl->counterAvailabilityImage.data;
+    papi_errno = cuptiProfilerGetCounterAvailabilityPtr(&getCounterAvailabilityParams);
+    if (papi_errno != CUPTI_SUCCESS) {
+        ERRDBG("CUPTI error %d: Failed to get bytes.\n", papi_errno);
+        return PAPI_EMISC;
+    }
+    return PAPI_OK;
+}
+
+// Adapted from `metric.cpp` -------------------------
```
gcongiu, line 656: Where is metric.cpp? This comment is not necessary

src/components/cuda/cupti_profiler.c
```diff
@@ -0,0 +1,1564 @@
...
+        .pCounterDataBuilder = counterDataBuilderCreateParams.pCounterDataBuilder,
+    };
+    NVPW_CALL(NVPW_CounterDataBuilder_DestroyPtr(&counterDataBuilderDestroyParams), goto fn_fail;);
+
+    return PAPI_OK;
+fn_fail:
+    return PAPI_EMISC;
+}
+
+// Adapted from 11.2/cupti-samples/userrange_profiling/simplecuda.cu:CreateCounterDataImage
```
gcongiu, line 793: Not really needed.

src/components/cuda/cupti_profiler.c
```diff
@@ -0,0 +1,1564 @@
...
+        .pConfig = configImage->data,
+        .configSize = configImage->size,
+        .minNestingLevel = 1,
+        .numNestingLevels = 1,
+        .passIndex = 0,
+        .targetNestingLevel = 1,
+    };
+    CUPTI_CALL(cuptiProfilerSetConfigPtr(&setConfigParams), goto fn_fail;);
+
+    /* User takes the resposiblity of replaying the kernel launches */
```
gcongiu, line 909: Be consistent with comments, you are mixing C99 (and onwards) and C89 comments. PAPI is officially C89 compliant. I don't really care about which one you use, but be consistent with the one you pick.

src/components/cuda/cupti_profiler.c
```diff
@@ -0,0 +1,1564 @@
...
+        .ctx = NULL,
+    };
+    CUPTI_CALL(cuptiProfilerBeginPassPtr(&beginPassParams), goto fn_fail;);
+
+    CUpti_Profiler_EnableProfiling_Params enableProfilingParams = {
+        .structSize = CUpti_Profiler_EnableProfiling_Params_STRUCT_SIZE,
+        .pPriv = NULL,
+        .ctx = NULL,
+    };
+    CUPTI_CALL(cuptiProfilerEnableProfilingPtr(&enableProfilingParams), goto fn_fail;);
```
gcongiu, line 922: you don't need the `;` after `goto fn_fail`. The macro expansion already adds the `;`

src/components/cuda/cupti_profiler.c
```diff
@@ -0,0 +1,1564 @@
...
+        }
+    }
+fn_exit:
+    return papi_errno;
+fn_fail:
+    papi_errno = PAPI_EMISC;
+    goto fn_exit;
+}
+
+int cuptip_get_event_description(const char *evt_name, char *description)
```
gcongiu, line 1216: `event_name_to_descriptor`. I think I got this wrong too in the rocm component. I need to fix the naming too

src/components/cuda/cupti_profiler.c
gcongiu: disabled reason is set at init private I think. init component temporarily sets it to a delay init message.

src/components/cuda/lcuda_common.h
```diff
@@ -0,0 +1,49 @@
...
+
+typedef struct event_record_s {
+    char name[PAPI_2MAX_STR_LEN];
+    char desc[PAPI_2MAX_STR_LEN];
+    int gpu_id;
+    unsigned int evt_code;
+    unsigned int evt_pos;  // index of added event
+    int num_dep;
+    double value;
+    void *info;  // API specific details
```
gcongiu, line 20: No comment in the same line. Comment in the line before please. We should fix this in other components too.

gcongiu (changes requested):

src/components/cuda/cupti_profiler.c
gcongiu: I will review that after we merge this

src/components/cuda/cupti_common.c
anustuvicl: I made it a typedef. Remove what (please be specific)? The type itself is useful to record the gpu organization of the system, can't remove the enum.

src/components/cuda/cupti_profiler.c
anustuvicl: This is to facilitate easier logging. If an error occurs 3 function calls down from this function, the debug log will tell exactly which path of function calls led to that error. Without it, that would be unclear. The "Exiting" log would serve no such purpose.

src/components/cuda/cupti_profiler.c
anustuvicl: changed to `cuptip_event_name_to_descr`. see similar prev comment.

src/components/cuda/cupti_common.c
anustuvicl: Do you mean follow the same convention throughout, or on a per statement basis?

gcongiu: throughout possibly. Unfortunately we don't enforce coding standards but we should do it eventually. Too much inconsistency across the code. So I am not picking about the specific style just pick one and stick with it. We can fix the style eventually to be consistent across the code base.

src/components/cuda/cupti_common.c
anustuvicl: `-1` means this function was not called before, any other value means it was called before and use that value instead. Have a look at the implementation where a static variable is used to track this.

src/components/cuda/cupti_profiler.c
anustuvicl: Function pointer names are within spaces, the signatures are not. How would you like it?

gcongiu: But when you call the function you use spaces. This seems inconsistent.

src/components/cuda/cupti_profiler.c
gcongiu: I think that might work better.

src/components/cuda/cupti_common.c
anustuvicl: Changed to `typedef` ... `gpu_collection_e`.

anustuvicl:
I have made most of the refactoring changes requested by @gcongiu. I would like to raise two points
1. The disabled_reason is a local `const char *` in `cuda_init_private` which gets passed down to the backend to be updated while also avoiding any `strcpy`. Having a `cuptic_err_get_last` will require a global variable, and exposing more API between the front end and backend.
2. The `cuptiu_` API that deals with the dynamic hashed event table is used by the front end and backend for their purposes. Exposing this API to the front end by wrapping it inside `cuptid_` wrapper functions is only an extra layer of indirection. The `cupti_utils.h` header is already exposed to `linux-cuda.c` through `cupti_dispatch.h`.

Do you think these points will be a hindrance for the events API component to connect with the front-end?

gcongiu:
> I have made most of the refactoring changes requested by @gcongiu. I would like to raise two points
> 
> 1. The disabled_reason is a local `const char *` in `cuda_init_private` which gets passed down to the backend to be updated while also avoiding any `strcpy`. Having a `cuptic_err_get_last` will require a global variable, and exposing more API between the front end and backend.
This is not a problem. It is, actually, how most APIs work.

> 2. The `cuptiu_` API that deals with the dynamic hashed event table is used by the front end and backend for their purposes. Exposing this API to the front end by wrapping it inside `cuptid_` wrapper functions is only an extra layer of indirection. The `cupti_utils.h` header is already exposed to `linux-cuda.c` through `cupti_dispatch.h`.
> 
I don't see the problem here.

> Do you think these points will be a hindrance for the events API component to connect with the front-end?
The events API was written with isolation in mind. The API you have reveals more stuff to the frontend so, to integrate with that, the events API will have to change to some degree for sure.

src/components/cuda/cupti_dispatch.h
```diff
@@ -0,0 +1,27 @@
+/**
+ * @file    cupti_dispatch.h
+ * @author  Anustuv Pal
+ *          anustuv@icl.utk.edu
+ */
+
+#ifndef __CUPTI_DISPATCH_H__
+#define __CUPTI_DISPATCH_H__
+
+#include "cupti_utils.h"
```
gcongiu, line 10: I think we should have this wrapped with `cuptid_` equivalents. This cupti_utils interface should be internal to the cupti wrapper layer.

anustuvicl:
Last commit implements the `cuptic_` disabled reason getter and setter functions. Also provides a `cuptid_` getter function for linux-cuda.c to get the disabled reason.

The second last commit is a half way towards removing the hard dependence on `struct cuptiu_event_t`. Currently, I have removed any unnecessary members. Records are private and external APIs cannot directly access internal arrays or members. A record getter function is provided `cuptiu_event_table_get_item`. It is now possible to extend event record type with further members in private APIs, as long as the struct header is common.

Currently there are 15 usages of `cuptiu_` facilities in `linux-cuda.c`. I will work on exposing / wrapping them through the `cuptid_` interface.

gcongiu (approved):

jagode (approved):

Issue #22, opened 2023-06-16
/usr/share/papi/components/pcp/tests/testPCP FAILED


wcohen:
When reviewing the papi tests results I noticed that the testPCP test failed even if the machine has PCP setup on the machine.  I can see a large number of PCP events listed by "pminfo".  However, when running testPCP on Fedora rawhide (papi-7.0.1) see the following:

```
$ /usr/share/papi/components/pcp/tests/testPCP 
PAPI Error: Couldn't open hw_instructions in exclude_guest=0 test
PAPI_Library_init run time =    5301.0 uS
Testing PCP Component with PAPI 7.0.1
Found PCP Component at id 11
PAPI_enum_cmp_event returned -17 [Component Index isn't set].
FAILED!!!
Line # 108 Error in PAPI_enum_cmp_event failed.
: Component Index isn't set
Some tests require special hardware, permissions, OS, compilers
or library versions. PAPI may still function perfectly on your 
system without the particular feature being tested here.   
```

I took a look at earlier versions of papi, 6.0.0 and see the same result.  Are others able to successful run testPCP?  Is there something I need to change to successfully run it?

jagode:
What information does "papi_component_avail" give you?

What's the perf_event_paranoid flag set to on your system?
It sounds like this is a paranoid issue.

wcohen:
Here is the output of papi_component_avail and it looks that the component is disabled.  If the component is disabled ,shouldn't the test skip using test_skip rather than failing?

```
$ papi_component_avail
Available components and hardware information.
--------------------------------------------------------------------------------
PAPI version             : 7.0.1.0
Operating system         : Linux 6.4.0-0.rc6.20230614gitb6dad5178cea.49.fc39.x86_64
Vendor string and code   : GenuineIntel (1, 0x1)
Model string and code    : Intel(R) Xeon(R) E-2276M  CPU @ 2.80GHz (158, 0x9e)
CPU revision             : 13.000000
CPUID                    : Family/Model/Stepping 6/158/13, 0x06/0x9e/0x0d
CPU Max MHz              : 2808
CPU Min MHz              : 2808
Total cores              : 1
SMT threads per core     : 1
Cores per socket         : 1
Sockets                  : 1
Cores per NUMA region    : 1
NUMA regions             : 1
Running in a VM          : yes
VM Vendor                : KVMKVMKVM
Number Hardware Counters : 11
Max Multiplex Counters   : 384
Fast counter read (rdpmc): yes
[...]

PR #23, opened 2023-06-18
pcp: skip test if component is disabled

dbarry9:
Previously, the test would fail if the PCP component was disabled. These changes check to see if it is disabled, and if so, skip the test.

These changes have been tested on the IBM POWER9 architecture.

Addresses issue #22


jagode (approved):

dbarry9:
Yes, we can.

PR #27, opened 2023-06-20
sysdetect: add include path for cuda headers to makefile

gcongiu:
## Pull Request Description
This PR adds explicit preprocessor flags to the sysdetect makefile in order to allow the component to find the cuda headers in the system.

 ## Author Checklist
 - [x] **Description**
 _Why_ this PR exists. Reference all relevant information including _background_, _issues_, _test failures_, etc
 - [x] **Commits**
 _Commit_ headers have the form: `module: short description`
 _Commit_ bodies contain a detailed description of what problem is addressed and how it is addressed
 - [ ] **Tests**


jagode (approved):
Tested on Saturn with the same setup that failed yesterday. It works fine with the fix.

PR #32, opened 2023-06-27
tests: disable components tests

gcongiu:
# Pull Request Description
By default the build system builds all tests. This causes problems with spack if some specific component is enabled (e.g. rocm). Moreover, tests should not be built unless `make fulltest` (or `make tests`) is ran. This PR fixes the behavior building tests only when `make fulltest` (or `make tests`) is called.

 ## Author Checklist
 - [x] **Description**
 - [x] **Commits**
 - [ ] **Tests**

### DIFF
--- a/.github/workflows/ci.sh
+++ b/.github/workflows/ci.sh
@@ -57,5 +57,5 @@ utils/papi_component_avail | grep -A1000 'Active components' | grep -q "Name:   
 
 if [ "$COMPONENT" != "cuda" ]; then
    echo Testing
-   ./run_tests.sh
+   make fulltest
 fi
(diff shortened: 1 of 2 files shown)


gcongiu:
This PR is not achieving the intended behaviour.

gcongiu:
This should work now. @G-Ragghianti @jagode can you have a look?

G-Ragghianti:
The tests are running.  They still produce a lot of `PAPI Error: Error Code -1,Invalid argument` and similar error messages, but the end result is a non-failure. If these are not "real" errors, can the output be reduced?

gcongiu:
@G-Ragghianti there are a bunch of non-critical errors like `PAPI_ECMP`, `PAPI_ENOEVNT`, `PAPI_EINVAL`. I need to see if we can filter out these non-critical errors from the output somehow.

G-Ragghianti (approved):

PR #33, opened 2023-06-27
rocm_smi: rename special cases to derived events

gcongiu:
Derived events are handled in the `handle_special_events` function. This name is not intuitive. Use the more descriptive `handle_derived_events` instead.


Treece-Burgess:
@adanalis This looks good to me, I tested locally as well with everything appearing to behave as expected.

Treece-Burgess (approved):

PR #36, opened 2023-06-28
Add smoke tests for spack install

G-Ragghianti:
We need to include the required "smoke test" files so that spack can use these to validate the installation of papi.  I put these into `src/smoke_tests`.  Is there a better location?

Once the location and require test environment are set, we can implement the smoke tests within the papi spack package.

### DIFF
--- a/.github/workflows/spack.sh
+++ b/.github/workflows/spack.sh
@@ -1,0 +1,24 @@
+#!/bin/bash -e
+
+COMPILER=$1
+
+[ -z "$COMPILER" ] && COMPILER=gcc@10 
+
+source /etc/profile
+set +x
+set -e
+trap 'echo "# $BASH_COMMAND"' DEBUG
+shopt -s expand_aliases
+
+export HOME=`pwd`
+git clone https://github.com/spack/spack spack || true
+source spack/share/spack/setup-env.sh
+
+SPEC="papi@master +lmsensors +powercap +sde +infiniband +rapl +cuda %$COMPILER"
+echo SPEC=$SPEC
+
+module load $COMPILER
+spack compiler find
+spack install --fresh --only=dependencies $SPEC
+spack dev-build -i --fresh --test=root $SPEC
+spack test run papi
--- a/src/smoke_tests/Makefile
+++ b/src/smoke_tests/Makefile
@@ -1,0 +1,22 @@
+
+EXECUTABLES = simple threads
+
+CC ?= gcc
+
+PAPIROOT ?= /path/to/papi/install/prefix
+CFLAGS ?= -O0 -pthread -I$(PAPIROOT)/include
+LIBS = -L$(PAPIROOT)/lib  -lm -ldl -lpapi -Wl,-rpath=$(PAPIROOT)/lib
+
+all: $(EXECUTABLES)
+
+clean:
+	/bin/rm -f core *.o $(EXECUTABLES)
+
+i.SUFFIXES: .c .o
+	.c.o:
+	    $(CC) $(CFLAGS) -c $*.c
+
+simple: simple.o 
+	$(CC) $(CFLAGS) -o simple simple.o $(LIBS)
+threads: threads.o 
+	$(CC) $(CFLAGS) -o threads threads.o $(LIBS)
--- a/src/smoke_tests/simple.c
+++ b/src/smoke_tests/simple.c
@@ -1,0 +1,69 @@
+#include <stdlib.h>
+#include <unistd.h>
+#include <stdio.h>
+#include <papi.h>
+#define NUM_EVENTS 2
+
+int main( int argc, char **argv )
+{
+    int retval, i;
+    long long       values[NUM_EVENTS];
+    int             EventSet = PAPI_NULL;
+    int             events[NUM_EVENTS];
+    char            *EventName[] = { "PAPI_TOT_CYC", "PAPI_TOT_INS" };
+    
+    retval = PAPI_library_init( PAPI_VER_CURRENT );
+    if ( retval != PAPI_VER_CURRENT ) {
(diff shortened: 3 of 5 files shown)


gcongiu:
Probably stupid question. Why don't we simply run the PAPI utilities instead? E.g., `papi_command_line`, `papi_component_avail`, etc?

G-Ragghianti:
The smoke tests satisfy a different need.  They are intended to replicate
what a user would do after the papi package has been installed.  In this
case, papi_component_avail and similar could be used, but the main interest
is being able to link other code against the installed papi library and
make sure that the code can minimarlly use the functionality of the library.

gcongiu:
Ok, since the utilities are programs statically linked to `libpapi.a` this makes sense. Would it also make sense to run, along with smoke tests, some of the utilities too? The case for `papi_component_avail`, for example, is that it will force the initialisation of all components. Thus, it is a further guarantee that all components will at least initialise correctly. The smoke tests don't do that as far as I can tell...

> The smoke tests satisfy a different need. They are intended to replicate what a user would do after the papi package has been installed. In this case, papi_component_avail and similar could be used, but the main interest is being able to link other code against the installed papi library and make sure that the code can minimarlly use the functionality of the library.
> […](#)

G-Ragghianti:
Yes, we can also develop a script that runs tests on the installed
utilities.

.github/workflows/spack.sh
```diff
@@ -0,0 +1,25 @@
...
+export HOME=`pwd`
+git clone https://github.com/spack/spack spack || true
+source spack/share/spack/setup-env.sh
+
+SPEC="papi@master +lmsensors +powercap +sde +infiniband +rapl +cuda %$COMPILER"
+echo SPEC=$SPEC
+
+module load $COMPILER
+spack compiler find
+#spack uninstall -a -y papi || true
```
gcongiu, line 22: Can you remove this comment?

src/smoke_tests/Makefile
```diff
@@ -0,0 +1,22 @@
+
+EXECUTABLES = simple threads
+
+CC ?= gcc
+
+PAPIROOT ?= /path/to/papi/install/prefix
+CFLAGS ?= -O0 -pthread -I$(PAPIROOT)/include
```
gcongiu, line 7: Applications will likely use -O2

src/smoke_tests/threads.c
```diff
@@ -0,0 +1,124 @@
...
+        printf( "%12lld \t\t --> %s  (thread %d) \n", values[i],
+                EventName[i], thread );
+    }
+
+    retval = PAPI_unregister_thread(  );
+    if ( retval != PAPI_OK ) {
+        printf("ERROR: PAPI_unregister_thread: %d: %s\n", retval, PAPI_strerror(retval));
+        exit(EXIT_FAILURE);
+    }
+
```
gcongiu, line 72: Should destroy the eventset too.

src/smoke_tests/simple.c
```diff
@@ -0,0 +1,69 @@
...
+    if ( retval != PAPI_OK ) {
+        printf("ERROR: PAPI_stop: %d: %s\n", retval, PAPI_strerror(retval));
+        exit(EXIT_FAILURE);
+    }
+                
+    for( i = 0; i < NUM_EVENTS; i++ ) {           
+        printf( "%12lld \t\t --> %s  \n", values[i],
+                EventName[i] );
+    }
+
```
gcongiu, line 65: Should destroy eventset too.

gcongiu (changes requested):
Looks good to me. Just a couple minor comments. For the utilities we should check with @jagode she is fine adding some (e.g., `papi_component_avail`). I would have them but maybe is not necessary.

When everything is finalised could you also cleanup the commits, i.e. squash the ones that need squashing, and adding editing comments using the format in the wiki (Authors checklist)?

src/smoke_tests/simple.c
jagode: AFAIK, PAPI_shutdown(), which I included, cleans up and frees event sets.

src/smoke_tests/threads.c
jagode: AFAIK, PAPI_shutdown(), which I included, cleans up and frees event sets.

jagode (approved):

jagode:
> Looks good to me. Just a couple minor comments. For the utilities we should check with @jagode she is fine adding some (e.g., `papi_component_avail`). I would have them but maybe is not necessary.

Not needed for the spack smoke tests.

.github/workflows/spack.sh
G-Ragghianti: yep

src/smoke_tests/Makefile
G-Ragghianti: This was provided by Heike.  I don't have a preference.

G-Ragghianti:
> Looks good to me. Just a couple minor comments. For the utilities we should check with @jagode she is fine adding some (e.g., `papi_component_avail`). I would have them but maybe is not necessary.
> 
> When everything is finalised could you also cleanup the commits, i.e. squash the ones that need squashing, and adding editing comments using the format in the wiki (Authors checklist)?

I rebased according to the wiki instructions.  Squash which commits?  Add the checklist to the original PR content or to a comment?  Do I need to go back and change the commit headers and bodies according to the checklist?

gcongiu:
I think we can squash commits into only two commits: one that adds the C smoke tests and Makefile (commits 6 and 7) and one that concerns the CI integration of those tests (all the remaining).

G-Ragghianti:
OK I'll see about that.  Just a detail: the smoke tests are not part of the CI.  They are to be used by the spack package for papi when tested after install via spack (mainly required by ECP projects).

gcongiu (approved):

PR #38, opened 2023-06-28
Correct cpuid_get_leaf4_mask to find number of physical cores

wcohen:
The cpuid_get_leaf4_mask function was using a logical and operation rather than a bitwise and operation to extract the count of physical cores.


gcongiu:
Thank you @wcohen, we just fixed the same issue with https://github.com/icl-utk-edu/papi/pull/34

wcohen:
Great.  Glad to hear there is a fix queued up for it. I  did the coverity scan last week, then went on vacation.  Just starting to go through and submit fixes for things listed in the coverity scan.

PR #39, opened 2023-06-28
don't use fast rdpmc counter reads in attach or syswide scenarios

deater:
With perf_event we can use fast rdpmc reads for low-overhead counter access.  This only works in self-monitoring situations where the thread being measured is in the same process context and same CPU as PAPI.  This means it cannot generally be used in the attach case, or if trying to do system-wide measurements (granularity anything other than PAPI_GRN_THR).

Ideally the Linux kernel would notice the request to use rdpmc in inappropriate circumstances and cause the mmap() read to fail and fallback to using the read() syscall.  However for various reasons the kernel devs did not want to support this, so it's up to PAPI to avoid using rdpmc in cases where Linux will silently fail and allow rdpmc to return invalid counter values.

This should fix the "attach_cpu_sys_validate" test failure.


gcongiu (approved):
This looks good to me.

gcongiu:
Addresses same problem as https://github.com/spack/spack/pull/38574

jagode (approved):

PR #41, opened 2023-06-29
Sde shutdown

adanalis:
This PR fixes the code that unregisters counters in the presence of groups. The basic idea behind the change is adding a reference count to each counter so we can keep track of the number of groups it belongs to. In the process of fixing this, I also abstracted the use of the hash-tables a bit, so we don't access their internal structure from multiple places in the code, and added some locking to prevent race conditions between reading and shutdown.

### DIFF
--- a/src/sde_lib/sde_lib_datastructures.c
+++ b/src/sde_lib/sde_lib_datastructures.c
@@ -54,6 +54,63 @@ void ht_insert(papisde_list_entry_t *hash_table, int ht_key, sde_counter_t *sde_
     return;
 }
 
+// This function serializes the items contained in the hash-table. A pointer
+// to the resulting serialized array is put into the parameter "rslt_array".
+// The return value indicates the size of the array.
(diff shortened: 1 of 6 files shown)


gcongiu:
Nitpick: can you use the format in the wiki's "Authors Checklist" for commits? it is easier to understand what part of the code the commits refers to when doing a `git log --oneline`.

I will do a full review soon.

gcongiu (approved):

src/sde_lib/sde_lib_datastructures.c
```diff
@@ -127,6 +127,8 @@ sde_counter_t *ht_delete(papisde_list_entry_t *hash_table, int ht_key, uint32_t
         item = list_head->item;
         if( NULL != list_head->next)
             *list_head = *(list_head->next);
+        else
+            memset(list_head, 0, sizeof(papisde_list_entry_t));
         return item;
```
gcongiu, line 132: Can you use braces for one liners if statements?

PR #45, opened 2023-07-07
Implementing clang static code analysis

G-Ragghianti:
## Pull Request Description
Implement clang static code analysis from LLVM "scan-build" utility.

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [ ] **Tests**

### DIFF
--- a/.github/workflows/clang_analysis.sh
+++ b/.github/workflows/clang_analysis.sh
@@ -1,0 +1,18 @@
+#!/bin/bash -e
+
+OUT=$1
+
+echo Analysis output to $OUT
+
+source /etc/profile
+set +x
+set -e
+trap 'echo "# $BASH_COMMAND"' DEBUG
+shopt -s expand_aliases
+
+module load llvm
+which scan-build
+cd src
+scan-build -o $OUT ./configure
+scan-build -o $OUT make
+
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -50,3 +50,15 @@ jobs:
       - uses: actions/checkout@v3
       - name: Build/Test/Install via Spack
         run: .github/workflows/spack.sh
+  papi_clang_analysis:
+    runs-on: cpu
+    steps:
+      - uses: actions/checkout@v3
+      - name: Run static analysis
+        run: .github/workflows/clang_analysis.sh clang-analysis-output
+      - name: Archive analysis results
+        uses: actions/upload-artifact@v3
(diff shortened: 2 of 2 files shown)


G-Ragghianti:
I'm currently waiting for the build of llvm to finish so that the action will have access to the clang-analyze tool.

gcongiu:
@G-Ragghianti I am not sure about the clang static analysis tool. Looking at the Coverity static analysis output I can see a lot of useful information for different parts of the code. The clang static analysis tool only reports one error and that's it. Maybe it can be configured differently to output more information.

G-Ragghianti:
It actually reports 5 bugs found, not one.  This is in the direct output from the run of the tool.  It also provides details in a zip file that is saved as an artifact.  I see this as complementary to the coverity scan.  This clang tool has to distinct advantages that I can see right away:

1. Clang's scan-build will return an error if bugs are found.  This is properly handled by github actions and can be used to prevent merging until the problems are resolved.  Then way coverity is running, it will only trigger the scan.  It won't return actionable results to github.
2. Clang's tool doesn't require login and management on a separate cloud service website. 

> @G-Ragghianti I am not sure about the clang static analysis tool. Looking at the Coverity static analysis output I can see a lot of useful information for different parts of the code. The clang static analysis tool only reports one error and that's it. Maybe it can be configured differently to output more information.

gcongiu:
@G-Ragghianti ok I missed the ones in the middle. Ok I will look at the artefacts as well. I think this is fine. If it is rebased onto master already we can merge it.

> It actually reports 5 bugs found, not one. This is in the direct output from the run of the tool. It also provides details in a zip file that is saved as an artifact. I see this as complementary to the coverity scan. This clang tool has to distinct advantages that I can see right away:
> 
> 1. Clang's scan-build will return an error if bugs are found.  This is properly handled by github actions and can be used to prevent merging until the problems are resolved.  Then way coverity is running, it will only trigger the scan.  It won't return actionable results to github.
> 2. Clang's tool doesn't require login and management on a separate cloud service website.
> 
> > @G-Ragghianti I am not sure about the clang static analysis tool. Looking at the Coverity static analysis output I can see a lot of useful information for different parts of the code. The clang static analysis tool only reports one error and that's it. Maybe it can be configured differently to output more information.

gcongiu (approved):

G-Ragghianti:
Yes, it is rebased.  Thanks.

PR #46, opened 2023-07-07
Fixes for coverity reported issues in various files

wcohen:
I reviewed coverity scan results for papi.  There were a number of
areas flagged by coverity.  This pull request addresses a small
portion of them.  Specifically it addresses:

ctests/thrspecific.c: cleaning up allocated memory
sysdetect: properly closing file on all paths through code
linux-net: ensuring that strings are properly terminated
coretemp: ensuring that strings are properly terminated
coretemp:  properly closing file on all paths through code

The commits are self contained and only do one thing for each commit. The also contain a description of what each commit is doing.  However, the headers do not precisely follow the format of 'module: short description'.  

I ran the tests for both the master git branch and this branch. I did not see any differences in the results.

## Author Checklist
- [x] **Description**
- [ ] **Commits**
- [ ] **Tests**


src/components/net/linux-net.c
```diff
@@ -343,8 +348,11 @@ _net_init_component( int cidx  )
     _net_native_events = (NET_native_event_entry_t*)
         papi_malloc(sizeof(NET_native_event_entry_t) * num_events);
     do {
-        strncpy(_net_native_events[i].name, t->name, PAPI_MAX_STR_LEN);
-        strncpy(_net_native_events[i].description, t->description, PAPI_MAX_STR_LEN);
+        int retlen;
+        retlen = snprintf(_net_native_events[i].name, PAPI_MAX_STR_LEN, "%s", t->name);
+        if (retlen <= 0 || retlen >= PAPI_MAX_STR_LEN) HANDLE_STRING_ERROR;
```
gcongiu, line 353: Probably, exiting the user program just because of a string truncation is not a good idea. We could replace HANDLE_STRING_ERROR to just log a message using, e.g., `SUDBG`. I know other part of the component do that but we should fix them too. Maybe we can address this problem in another PR though.

gcongiu (approved):
@wcohen Looks good. I only have a nitpick on the commit messages format. Can you reformat the commit messages using `model: short description` like in the checklist? That makes it easier to locate what part of the code the commit belongs to when doing a `git --oneline`. E.g., `coretemp: add closedir operation to function exit`.

src/components/net/linux-net.c
wcohen: I was thinking that exiting on truncation might have been a bit more than desired for those places too.

For components just use the component name.  What is the naming conventions for things in other locations like src/papi*.[ch],  testlib, ctests, utils, validation_tests, etc.?

I can work revising the patches this week.

gcongiu: I think it is safe to use the name of the directory for components (e.g., cuda, rocm, etc) and tests. For things like the PAPI framework, maybe, we can agree on some naming convention (e.g., framework). I think I used `memory` or `threads` in the past for modifying code that belongs to either of those subsystems. Generally, anything that can give a good hint of what part of PAPI is affected by the changes in the commit would be good for history readability.

gcongiu:
@wcohen do you have any updates here? Can we merge this one?

gcongiu:
Superseded by PR #111

Issue #61, opened 2023-08-01
Updated PAPI CUDA component returns default for number of counters


jrodgers-github:
Noticed that the updated PAPI CUDA component is returning the default number of counters for `<papi_vector>.cmp_info.num_cntrs`, specifically `PAPI_CUDA_MAX_COUNTERS` (30).

Example `papi_component_avail` behavior:
- New component:
```
Name:   cuda                    CUDA profiling via NVIDIA CuPTI interfaces
                                Native: 1052928, Preset: 0, Counters: 30
```
- Old component:
```
Name:   cuda                    CUDA events and metrics via NVIDIA CuPTI interfaces
                                Native: 1052928, Preset: 0, Counters: 1052928
```

Using the old component as a guide, I think this could be fixed by capturing the counter count along side the native counter count.
- Old component:
```
int _cuda_init_private(void)
....
    /* Export some information */
    _cuda_vector.cmp_info.num_native_events = global_cuda_context->availEventSize;
    _cuda_vector.cmp_info.num_cntrs = _cuda_vector.cmp_info.num_native_events;
....
```
- New component:
```
static int cuda_ntv_enum_events(unsigned int *event_code, int modifier)
...
    _cuda_vector.cmp_info.num_native_events = global_event_names->count;
    _cuda_vector.cmp_info.num_cntrs = global_event_names->count;     < ------ WHAT NEEDS TO BE ADDED
....
```

jrodgers-github:
Confirmed issue was resolved in the following PR: https://github.com/icl-utk-edu/papi/pull/62

Updated `papi_component_avail` behavior:
```
Name:   cuda                    CUDA profiling via NVIDIA CuPTI interfaces
                                Native: 1053696, Preset: 0, Counters: <Not supported by vendor.>
```

@anustuvicl & @gcongiu; will the `<Not supported by vendor>` counters note be adopted by the other GPU (e.g. `nvml`, `rocm`, `rocm_smi`) and networking related components?

gcongiu:
@jrodgers-github I think that is the intent, i.e., extend this to other GPU (non-GPU) components.

PR #62, opened 2023-08-04
utils: papi_component_avail does not support cuda component counters

anustuvicl:
This commit updates `utils/papi_component_avail` to display cuda component does not support counters.

### DIFF
--- a/src/components/cuda/linux-cuda.c
+++ b/src/components/cuda/linux-cuda.c
@@ -72,7 +72,6 @@ papi_vector_t _cuda_vector = {
         .version = "0.1",
         .description = "CUDA profiling via NVIDIA CuPTI interfaces",
         .num_mpx_cntrs = PAPI_CUDA_MPX_COUNTERS,
-        .num_cntrs = PAPI_CUDA_MAX_COUNTERS,
         .default_domain = PAPI_DOM_USER,
         .default_granularity = PAPI_GRN_THR,
         .available_granularities = PAPI_GRN_THR,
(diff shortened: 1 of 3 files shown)


jrodgers-github:
Related issue: https://github.com/icl-utk-edu/papi/issues/61

gcongiu:
I don't think `papi_component_avail` is the problem here. You should fix the component. Please have a look at `force_cmp_init` in `papi_component_avail`.

gcongiu:
To be more specific, don't make component specific hacks in `papi_component_avail`. The utility should not care about the name of the component to work correctly. I would add a `PAPI_CMP_NUM_COUNTERS_UNKNOWN` flag in the component itself (maybe set this to `0x8000000`) and set `cmpinfo->num_cntrs=PAPI_CMP_NUM_COUNTERS_UNKNOWN`. Then, in `papi_component_avail`, you can simply check for this flag in `cmpinfo->num_cntrs` to output the correct value.

anustuvicl:
Shouldn't `PAPI_CMP_NUM_COUNTERS_UNKNOWN` be defined in `papi.h` so that the component and the utility can both access the same constant?

anustuvicl:
If we are allowed to use -ve values, I would suggest setting the value to `PAPI_ECMP` (not supported by component). This can be easily checked by the utility.

anustuvicl:
Update: If the component sets `papi_vector_t.cmp_info.num_cntrs=PAPI_ECMP`, the utility displays the counter disabled reason from `papi_vector_t.cmp_info.disabled_reason`.

gcongiu:
> Update: If the component sets `papi_vector_t.cmp_info.num_cntrs=PAPI_ECMP`, the utility displays the counter disabled reason from `papi_vector_t.cmp_info.disabled_reason`.

I don’t understand why. disabled_reason will be unset if the component inits correctly.

anustuvicl:
Typedef `PAPI_component_info_t` introduces string member `misc_info` to communicate miscellaneous message from component to utilities. CUDA component sets `.num_cntrs=PAPI_ECMP` and sets `.misc_info` to tell `papi_component_avail` that vendor does not support that feature.

gcongiu (approved):

Issue #73, opened 2023-08-30
PAPI ROCm: Confusion with `HIP_VISIBLE_DEVICES`


bertwesarg:
When setting `HIP_VISIBLE_DEVICES` the id in the `:device=%d` event name suffix is still the hardware device index, not the HIP device index.

The `./sample_multi_kernel_monitoring` test always uses `:device=0`, so starting it with different `HIP_VISIBLE_DEVICES` values will result in 0-value results:

```console
$ HIP_VISIBLE_DEVICES=0 ./sample_multi_kernel_monitoring
rocm:::SQ_INSTS_VALU:device=0 : 191459309210
rocm:::SQ_INSTS_SALU:device=0 : 73288502349
rocm:::SQ_WAVES:device=0 : 526324
rocm:::SQ_WAVES_RESTORED:device=0 : 2032
$ HIP_VISIBLE_DEVICES=1 ./sample_multi_kernel_monitoring 
rocm:::SQ_INSTS_VALU:device=0 : 0
rocm:::SQ_INSTS_SALU:device=0 : 0
rocm:::SQ_WAVES:device=0 : 0
rocm:::SQ_WAVES_RESTORED:device=0 : 0
```

gcongiu:
Thank you for reporting this. If I understand it correctly HIP_VISIBLE_DEVICES is a list of physical devices that is visible to the process (and set by the resource manager) scheduled to run on the compute node. Let’s assume we have devices 4,5,6,7 set. Should papi_native_avail show events for those using device=0,1,2,3 and then remap them internally to 4,5,6,7? Would the behaviour be similar for cuda?

bertwesarg:
I'm currently not able to check CUDA. Your understanding is correct. But for ROCm it actually depends on the used runtime. It looks like `rocprofiler` is on the ROCm level, i.e., the same as ROCm SMI, but each higher-level runtime has there own [GPU isolation mechanism](https://rocm.docs.amd.com/en/latest/understand/gpu_isolation.html). Because PAPI does not know which runtime is used by the application, I think the only solution is, to document that the PAPI ROCm component expects ROCm SMI device indices, and the application using PAPI needs to take care of the mapping.

In case you are interested, here is how Score-P does this mapping via the device UUID:
1. [Create a mapping from UUID to ROCm SMI device index](https://gitlab.com/score-p/scorep/-/blob/master/src/adapters/hip/scorep_hip_callbacks.c?ref_type=heads#L225)
[...]

PR #75, opened 2023-08-30
rocm: extend README with device partitioning information

gcongiu:
## Pull Request Description
Clarify how the component accesses available devices through the runtime system and how this can be affected by resource partitioning at the resource manager level using selected environment variables.

Addresses issue #73 

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/rocm/README.md
+++ b/src/components/rocm/README.md
@@ -109,3 +109,29 @@ setting the ROCP\_TOOL\_LIB to the PAPI library as follows:
 
 ## Unusual installations
 For the ROCM component to be operational, it must find the dynamic libraries `libhsa-runtime64.so` and `librocprofiler64.so`. These are normally found in the above standard directories. If these libraries are not found (or are not functional) then the component will be listed as "disabled" with a reason explaining the problem. If libraries were not found, then they are not in the expected places.
+
+2. [Device isolation](#markdown-device-isolation)
+
+## Device isolation
+Compute clusters resource managers can isolate GPU devices, on compute nodes,
+into subgroups. This means that a job might only see part of the devices on the
+node. How many devices are visible depends on the value of a set of environment
+variables, configured by the resource manager (e.g. HIP\_VISIBLE\_DEVICES,
+ROCR\_VISIBLE\_DEVICES, etc).
+
+In order to detect available devices, the ROCm component relies on the HSA ROCm
+runtime functions (i.e. hsa\_iterate\_agents). The ROCR\_VISIBLE\_DEVICES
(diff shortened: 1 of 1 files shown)


src/components/rocm/README.md
```diff
@@ -85,3 +85,20 @@ setting the ROCP\_TOOL\_LIB to the PAPI library as follows:
...
+variables, configured by the resource manager (e.g. HIP\_VISIBLE\_DEVICES,
+ROCR\_VISIBLE\_DEVICES, etc).
+
+In order to detect available devices, the ROCm component relies on the HSA ROCm
+runtime functions (i.e. hsa\_iterate\_agents). The ROCR\_VISIBLE\_DEVICES
+environment variable establishes how many devices will be visible to the ROCm
+runtime. Therefore, by extension, the PAPI ROCm component will only see as many
+devices as allocated by the resource manager through the aforementioned
+environment variable. The component assigns them integer identifiers in the
+range [0, N-1], where N is the number of devices for the partition.
```
bertwesarg, line 104: Add something like:

Therefore, when using the component in an HIP context, the application would need to map the device index given by `hipGetDevice` to this index range and use this index in the event name, e.g., `rocm:::GPUBusy:device=X`. Preferable the UUID of the device should be used for this mapping (see `hipDeviceGetUuid` and `HSA_AMD_AGENT_INFO_UUID`).

gcongiu:
@bertwesarg I have updated the PR to reflect your comments.

src/components/rocm/README.md
bertwesarg: btw, that could also be a feature that the user uses the UUID in `:device=GPU-DEADBEEFDEADBEEF`. At least `ROCR_VISIBLE_DEVICES` does support this too.

bertwesarg: CUDA supports this in `CUDA_VISIBLE_DEVICES` too. But I'm not familiar with the CUDA metrics and components

gcongiu: @anustuvicl do you have any information for CUDA regarding this?

gcongiu: > btw, that could also be a feature that the user uses the UUID in `:device=GPU-DEADBEEFDEADBEEF`. At least `ROCR_VISIBLE_DEVICES` does support this too.

I fear this could confuse users though. Something to keep in mind. We could discuss this further with @jagode and @adanalis too

bertwesarg (approved):

PR #76, opened 2023-09-01
rocm: remove openmp dependency from tests

gcongiu:
## Pull Request Description

Spack installation of PAPI + rocm component have dependency issues with openmp caused by the AMD llvm compiler. Because component tests are always built in PAPI this prevents spack from installing PAPI in the system. Removing the openmp dependency and replacing with pthreads solves the issue.

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [ ] **Tests**


gcongiu:
@jagode I finally removed the openmp dependency in the rocm component tests. Now all multithreaded tests run with pthreads.

adanalis (approved):

PR #82, opened 2023-09-07
lmsensors: Add lib/ to explicit search path to .so loader

anustuvicl:
This behaviour observed on ICL Leconte

```
$ module load lm-sensors
$ export PAPI_LMSENSORS_ROOT=$ICL_LM_SENSORS_ROOT
$ ./configure --with-components="lmsensors" && make && make install
$ utils/papi_component_avail
```
Shows the output
```
...
Name:   lmsensors               Linux LMsensor statistics
   \-> Disabled: libsensors.so not found.
```

The problem happens because the component searches for `libsensors.so` in `$PAPI_LMSENSORS_ROOT/lib64` while the so exists in `$PAPI_LMSENSORS_ROOT/lib`

This PR fixes this problem by adding `lib` to the search path for the so file.

After this fix the output of `papi_component_avail` shows
```
...
Name:   lmsensors               lm-sensors provides tools for monitoring the hardware health
...
Name:   lmsensors               lm-sensors provides tools for monitoring the hardware health
                                Native: 170, Preset: 0, Counters: 170
```

### DIFF
--- a/src/components/lmsensors/linux-lmsensors.c
+++ b/src/components/lmsensors/linux-lmsensors.c
@@ -399,6 +399,12 @@ link_lmsensors_libraries ()
       dl1 = dlopen(path_name, RTLD_NOW | RTLD_GLOBAL);             // Try to open that path.
    }
 
+   // Step 4: Try another explicit install default.
+   if (dl1 == NULL && lmsensors_root != NULL) {                          // if root given, try it.
+      snprintf(path_name, PATH_MAX, "%s/lib/libsensors.so", lmsensors_root);   // PAPI Root check.
+      dl1 = dlopen(path_name, RTLD_NOW | RTLD_GLOBAL);             // Try to open that path.
+   }
+
    // Check for failure.
    if (dl1 == NULL) {
       snprintf(_lmsensors_vector.cmp_info.disabled_reason, PAPI_MAX_STR_LEN, "libsensors.so not found.");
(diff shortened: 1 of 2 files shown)


gcongiu:
This PR is missing any description and context. Please provide more information.

src/components/lmsensors/linux-lmsensors.c
```diff
@@ -399,6 +399,12 @@ link_lmsensors_libraries ()
       dl1 = dlopen(path_name, RTLD_NOW | RTLD_GLOBAL);             // Try to open that path.
    }
 
+   // Step 4: Try another explicit install default.
+   if (dl1 == NULL && lmsensors_root != NULL) {                          // if root given, try it.
+      snprintf(path_name, 1024, "%s/lib/libsensors.so", lmsensors_root);   // PAPI Root check.
```
gcongiu, line 404: Please use `PATH_MAX` instead of a raw value for string length

gcongiu (changes requested):

src/components/lmsensors/linux-lmsensors.c
gcongiu: You can update it in a separate commit

anustuvicl: Updated as requested with `PATH_MAX`, and added commit to update the same for previous statement.

gcongiu (approved):

PR #88, opened 2023-09-13
Set size of perf_attr_struct prior to getting pfm encoding

jishminor:
## Pull Request Description
Simple fix to ensure correct size of struct gets mem copied in `pfm_lib_perf_event_encode` in libpfm4.

Events from pfm4 which contain values in the config2 field of the `perf_event_attr` struct seem to not get copied in `pfm_lib_perf_event_encode` when opening up a PAPI event.

The general trace of this issue is:
1. ntv_evt->attr is zeroed out [here](https://github.com/icl-utk-edu/papi/blob/2c816bddca73893636a306c046cb1eb0f1dba693/src/components/perf_event/pe_libpfm4_events.c#L187)
2. attr.size is 0 [here](https://github.com/icl-utk-edu/papi/blob/2c816bddca73893636a306c046cb1eb0f1dba693/src/libpfm4/lib/pfmlib_perf_event.c#L112), resulting in `PERF_ATTR_SIZE_VER0` being used as the default size used to memcpy the `perf_event_attr` which the user gets back from the function [here](https://github.com/icl-utk-edu/papi/blob/2c816bddca73893636a306c046cb1eb0f1dba693/src/libpfm4/lib/pfmlib_perf_event.c#L285).
3. This results in additional fields required to open some events which additional fields added in perf such as config2 being set when requested.


gcongiu (approved):

gcongiu:
@jishminor I wrongly made an update using a merge commit. Could you rebase onto master and repush the commit?

PR #102, opened 2023-10-24
rocm: allow component to resolve device ambiguity

gcongiu:
## Pull Request Description
Sometimes users don't want to specify the device id right away as this might be unknown. PAPI can figure out what device the user intends to monitor by looking at the selected device (through hipGetDevice) when the event is resolved with name_to_code.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**


gcongiu:
Superseded by PR #123

PR #111, opened 2023-11-07
Fixes for coverity reported issues in various files (redo of pull#46)

wcohen:
## Rework of https://github.com/icl-utk-edu/papi/pull/46 commit messages to follow format guidelines

I reviewed coverity scan results for papi. There were a number of
areas flagged by coverity. This pull request addresses a small
portion of them. Specifically it addresses:

ctests/thrspecific.c: cleaning up allocated memory
sysdetect: properly closing file on all paths through code
linux-net: ensuring that strings are properly terminated
coretemp: ensuring that strings are properly terminated
coretemp: properly closing file on all paths through code

The commits are self contained and only do one thing for each commit. The also contain a description of what each commit is doing. However, the headers do not precisely follow the format of 'module: short description'.

I ran the tests for both the master git branch and this branch. I did not see any differences in the results.

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/components/coretemp/linux-coretemp.c
+++ b/src/components/coretemp/linux-coretemp.c
@@ -455,18 +455,18 @@ _coretemp_init_component( int cidx )
      }
 
      do {
-        char *strCpy;
-        strCpy=strncpy(_coretemp_native_events[i].name,t->name,PAPI_MAX_STR_LEN);
-        if (strCpy == NULL) HANDLE_STRING_ERROR;
+        int retlen;
+        retlen = snprintf(_coretemp_native_events[i].name, PAPI_MAX_STR_LEN, "%s", t->name);
+        if (retlen <= 0 || retlen >= PAPI_MAX_STR_LEN) HANDLE_STRING_ERROR;
 
-        strCpy=strncpy(_coretemp_native_events[i].path,t->path,PATH_MAX);
-        if (strCpy == NULL) HANDLE_STRING_ERROR;
+        retlen = snprintf(_coretemp_native_events[i].path, PATH_MAX, "%s", t->path);
+        if (retlen <= 0 || retlen >= PATH_MAX) HANDLE_STRING_ERROR;
 
-        strCpy=strncpy(_coretemp_native_events[i].units,t->units,PAPI_MIN_STR_LEN);
-        if (strCpy == NULL) HANDLE_STRING_ERROR;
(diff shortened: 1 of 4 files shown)


src/components/net/linux-net.c
```diff
@@ -51,6 +51,11 @@ papi_vector_t _net_vector;
 #define NET_INVALID_RESULT     -1
 
 
+// The following macro follows if a string function has an error. It should 
+// never happen; but it is necessary to prevent compiler warnings. We print 
+// something just in case there is programmer error in invoking the function.
+#define HANDLE_STRING_ERROR {fprintf(stderr,"%s:%i unexpected string function error.\n",__FILE__,__LINE__); exit(-1);}
```
gcongiu, line 57: I would not add `HANDLE_STRING_ERROR` anywhere. In fact, we should eventually remove it from everywhere it is used. Calling exit is not a good way of handling errors. Can you return `PAPI_EBUF` instead?

src/components/coretemp/linux-coretemp.c
```diff
@@ -455,18 +455,18 @@ _coretemp_init_component( int cidx )
      }
 
      do {
-        char *strCpy;
-        strCpy=strncpy(_coretemp_native_events[i].name,t->name,PAPI_MAX_STR_LEN);
-        if (strCpy == NULL) HANDLE_STRING_ERROR;
+        int retlen;
+        retlen = snprintf(_coretemp_native_events[i].name, PAPI_MAX_STR_LEN, "%s", t->name);
+        if (retlen <= 0 || retlen >= PAPI_MAX_STR_LEN) HANDLE_STRING_ERROR;
```
gcongiu, line 460: We should eventually replace `HANDLE_STRING_ERROR` with proper error handling. I think this patch looks good for the time being.

gcongiu (changes requested):

src/components/coretemp/linux-coretemp.c
wcohen: There is a lot of existing code that uses `HANDLE_STRING_ERROR` and very few places using `PAPI_EBUF`. Switching to `PAPI_EBUF` does sound like a good change for a future patch set.

gcongiu: Ok, I think you have a point. Let's keep this for consistency's sake.

gcongiu (approved):

PR #116, opened 2023-11-09
utils: Fix bogus "Disabled" message in papi_component_avail for the sde component.

adanalis:
## Pull Request Description
When the sde component is initialized, in the context of an application that uses PAPI, it looks for the availability of libsde symbols. The rationale is that if the application is not linked against libsde, there are no SDEs to read, so the component disables itself. Therefore, papi_component_avail, which does not export any SDEs itself, always reported the sde component as "Disabled". Adding the symbols to the utility resolves this problem.

## Author Checklist
- [x] **Description**
- [ ] **Commits**
- [ ] **Tests**


gcongiu:
@adanalis nitpick: could you change the commit message to the format: `utils: Fix bogus 'Disabled' message in papi_component_avail for sde comp`?

gcongiu (approved):

PR #121, opened 2023-11-15
PAPI: ARM Cortexx A76 support (Raspberry Pi 5)

melintea:
## Pull Request Description
ARM Cortexx A76 support (Raspberry Pi 5) - PAPI changes


src/papi_events.csv
```diff
@@ -1954,6 +1954,7 @@ PRESET,PAPI_L1_ICM,NOT_DERIVED,IFETCH_MISS
 #
 CPU,arm_ac15
 CPU,arm_ac57
```
gcongiu, line 1956: You can put this change in another commit or, even, a different PR. The first part of this patch does not need libpfm4 support and can be merged immediately. Moreover, there are no preset events for `arm_ac76` yet, so unless you have a list of well tested preset events you would like to suggest, I would not make any changes to this file.

gcongiu (changes requested):

gcongiu:
@melintea this looks good to me. Instead of adding another commit to remove the changes in the previous can you just squash the two?

gcongiu (approved):

PR #122, opened 2023-11-15
libpfm4: ARM Cortexx A76 support (Raspberry Pi 5)

melintea:
## Pull Request Description
ARM Cortexx A76 support (Raspberry Pi 5) - libpfm4 changes
(also sent the diff to libpfm4 maintainers)


gcongiu:
@melintea, you should send the libpfm4 maintainer a patch not a diff. You can do that as follows.
From your feature branch:
```
git format-patch HEAD~1
```
That will create a patch for the last commit in your branch. If you want the latest n commits, replace `HEAD~1` with `HEAD~n`.

melintea:
Yes, sent a patch

PR #129, opened 2023-12-01
ci: run tests with and without PAPI debug enabled

gcongiu:
## Pull Request Description
Tests should make sure real use cases work as expected. Some tests might not working correctly if -O0 is used as optimization level in the compiler. For example, the ROCm runtime submits a kernel of 4 waves if the tests are built using -O0, which makes the tests fail.

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/.github/workflows/ci.sh
+++ b/.github/workflows/ci.sh
@@ -1,7 +1,8 @@
 #!/bin/bash -e
 
 COMPONENT=$1
-COMPILER=$2
+DEBUG=$2
+COMPILER=$3
 
 [ -z "$COMPILER" ] && COMPILER=gcc@11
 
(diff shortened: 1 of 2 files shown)


gcongiu:
@G-Ragghianti the idea with this PR is to have PAPI tested with debug both on and off. I am not sure the way I have done things is the right though. Could you provide any feedback here? Thanks

G-Ragghianti:
I recommend against doing two builds within the same action because you need to make sure that the builds don't interfere with each other, when one fails, it will not run the other, and when there is a failure it isn't immediately clear which build caused the failure.  I recommend that we modify main.yaml to include a new axis "debug" in the actions matrix (with values "no" and "yes").  Then the existing ci.sh can be modified to reference this variable in the configure step.

.github/workflows/ci.sh
```diff
@@ -43,9 +44,9 @@ if [ "$COMPONENT" = "infiniband_umad" ]; then
 fi
 
 if [ "$COMPONENT" = "perf_event" ]; then
-  ./configure --with-debug=yes --enable-warnings
+  ./configure --with-debug=$DEBUG --enable-warnings
```
gcongiu, line 47: @G-Ragghianti will this work?

G-Ragghianti: Yes I think that will work.

G-Ragghianti (approved):

PR #132, opened 2023-12-06
2023.12.01 CAT DC benchmark opts

dbarry9:
## Pull Request Description
Various optimizations and logic fixes to the CAT data cache benchmarks.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**


src/counter_analysis_toolkit/main.c
```diff
@@ -320,10 +320,10 @@ static hw_desc_t *obtain_hardware_description(char *conf_file_name){
...
     hw_desc->split[2] = 1;
     hw_desc->split[3] = 1;
-    hw_desc->pts_per_reg[0] = 5;
-    hw_desc->pts_per_reg[1] = 5;
-    hw_desc->pts_per_reg[2] = 5;
-    hw_desc->pts_per_reg[3] = 5;
+    hw_desc->pts_per_reg[0] = 3;
+    hw_desc->pts_per_reg[1] = 3;
+    hw_desc->pts_per_reg[2] = 3;
+    hw_desc->pts_per_reg[3] = 3;
```
gcongiu: In the commit message do you mean "[...] increase the benchmark execution speed" rather than "[...] decrease the benchmark execution speed" ?

src/counter_analysis_toolkit/dcache.c
```diff
@@ -11,12 +11,14 @@ static void print_core_affinities(FILE *ofp);
 
 extern char* eventname;
 
-int min_size, max_size, is_core = 0;
+long min_size, max_size;
```
gcongiu, line 14: Maybe better use `int64_t` or `long long`. On some system `long` is same size as `int`

gcongiu (changes requested):
A few minor changes. Overall looks good to me. I will do another pass to be sure I haven't missed anything.

src/counter_analysis_toolkit/main.c
dbarry9: Yes, I meant to write "decrease the benchmark execution time." I will update this commit message.

src/counter_analysis_toolkit/dcache.c
dbarry9: Thank you for pointing this out. I will update it to 'long long.'

dbarry9: This has been updated per your suggestion.

src/counter_analysis_toolkit/main.c
dbarry9: This has been updated.

PR #133, opened 2023-12-07
2023.12.06 CAT MPI fixes

dbarry9:
## Pull Request Description
Various updates for MPI builds of CAT.

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**


src/counter_analysis_toolkit/main.c
```diff
@@ -181,6 +181,7 @@ int main(int argc, char*argv[])
     PAPI_shutdown();
 
 #if defined(USE_MPI)
```
gcongiu, line 183: Is there any communication still going on at this point? If not `MPI_Barrier` should not be needed. If yes can you clarify it?

dbarry9: In my testing environment, MPI_Finalize did not block. When a single rank reached MPI_Finalize before others, the others were terminated prematurely.

PR #139, opened 2023-12-14
Minor fixes for ROCm component qualifiers

gcongiu:
## Pull Request Description
The ROCm component `ntv_code_to_info` function does not behave consistently with other components (e.g., `perf_event`). This PR fixes the function behavior to align it to the CPU component.

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/components/rocm/roc_profiler.c
+++ b/src/components/rocm/roc_profiler.c
@@ -358,13 +358,13 @@ rocp_evt_code_to_info(uint64_t event_code, PAPI_event_info_t *info)
             }
             *(devices + strlen(devices) - 1) = 0;
             sprintf(info->symbol, "%s:device=%i", ntv_table_p->events[inf.nameid].name, inf.device);
-            sprintf(info->long_descr, "%s masks:Mandatory device qualifier [%s]",
+            sprintf(info->long_descr, "%s, masks:Mandatory device qualifier [%s]",
                     ntv_table_p->events[inf.nameid].descr, devices);
             break;
         }
         case INSTAN_FLAG:
             sprintf(info->symbol, "%s:instance=%i", ntv_table_p->events[inf.nameid].name, inf.instance);
-            sprintf(info->long_descr, "%s masks:Mandatory instance qualifier in range [0-%i]",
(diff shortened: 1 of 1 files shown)


src/components/rocm/roc_profiler.c
```diff
@@ -344,9 +344,20 @@ rocp_evt_code_to_info(uint64_t event_code, PAPI_event_info_t *info)
...
+            char devices[PAPI_MAX_STR_LEN] = { 0 };
+            for (i = 0; i < device_table_p->count; ++i) {
+                if (rocc_dev_check(ntv_table_p->events[inf.nameid].device_map, i)) {
+                    sprintf(devices + strlen(devices), "%i,", i);
+                }
+            }
+            *(devices + strlen(devices) - 1) = 0;
             sprintf(info->symbol, "%s:device=%i:instance=%i", ntv_table_p->events[inf.nameid].name, inf.device, inf.instance);
-            sprintf(info->long_descr, "%s", ntv_table_p->events[inf.nameid].descr);
+            sprintf(info->long_descr, "%s, masks:Mandatory device qualifier [%s]:Mandatory instance qualifier in range [0-%i]",
```
bertwesarg, line 357: I was now able to digest how `perf_event` is creating the mask description, and indeed it [looks](https://github.com/icl-utk-edu/papi/blob/fbd3230740ef86aafe17f527725e935a699e14bb/src/components/perf_event/pe_libpfm4_events.c#L386) like it is using `:` (colon) to separate multiple qualifiers. So we have two which are using a `,` (comma) and one which is using a `:` (colon). And you opted for `:` (colon) too.

gcongiu: Yes I looked at the `perf_event` output and found a `:` qualifier separator. I guess this is an inconsistency that ought to be fixed sooner or later.

gcongiu: Actually, the components you mentioned that use a `,` separator are no longer in use and should eventually be removed.

bertwesarg: thanks, old code makes the life harder for a programmer-archaeologist

adanalis (approved):

PR #145, opened 2023-12-19
cat: fix compile-time error

dbarry9:
On some older versions of GCC (10.3.0), not having at least a 'break' in a switch-case statement can yield the compiler warning: "label at end of compound statement"

These changes fix this error and have been tested on the AMD Zen3 architecture.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/counter_analysis_toolkit/flops.c
+++ b/src/counter_analysis_toolkit/flops.c
@@ -832,6 +832,7 @@ void exec_flops( int precision, int EventSet, FILE *fp ) {
 #endif
           break;
       default:
+          ;
     }
 
     return;


src/counter_analysis_toolkit/flops.c
```diff
@@ -832,6 +832,7 @@ void exec_flops( int precision, int EventSet, FILE *fp ) {
 #endif
           break;
       default:
+          break;
     }
```
gcongiu, line 836: There should not be any need for break in `default`

dbarry9: Should I use an empty statement instead; as in, just a semicolon?

gcongiu: Yes that would also work.

gcongiu (approved):

PR #147, opened 2023-12-27
sysdetect: fix bug where Makefile sets `-ffree-form` everywhere

tgamblin:
To fix ARM compiler builds, #108 added these lines to the `Makefile` for `sysdetect` tests:

```diff
+ FFLAGS += -ffree-form
+
  ifeq ($(notdir $(F77)),gfortran)
      FFLAGS +=-ffree-form -ffree-line-length-none
  else ifeq ($(notdir $(F77)),flang)
      FFLAGS +=-ffree-form
  else ifneq ($(findstring $(notdir $(F77)), $(intel_compilers)),)
      FFLAGS +=-free
  else ifneq ($(findstring $(notdir $(F77)), $(cray_compilers)),)
      FFLAGS +=-ffree
  endif
```

but this adds `-ffree-form` to every compiler. This breaks [cray compiler builds for PAPI 7.1.0](https://gitlab.spack.io/spack/spack/-/jobs/9684882):

```
     1307    /home/gitlab-runner-protected-3/builds/Cj9btwqn/0/spack/spack/lib/
             spack/env/cce/ftn -ffree-form -ffree -I../../.. -o query_device_si
             mple_f query_device_simple_f.F ../../../libpapi.a
     1308    ftn-78 ftn: ERROR in command line
     1309      The -f option has an invalid argument, "free-form".
```

You can see that CCE gets both `-ffree-form` and `-ffree` there.

This patch removes the global `-ffree-form` and makes the flang condition above a bit more permissive, so that both `flang`, `armflang`, `amdflang`, or really anything ending in `flang` adds `-free-form`:

```diff
       FFLAGS +=-ffree-form -ffree-line-length-none
-  else ifeq ($(notdir $(F77)),flang)
+  else ifeq ($(patsubst %flang,,$(notdir $(F77))),)  # compiler name ends with flang
       FFLAGS +=-ffree-form
```

This should fix the bug that #108 fixed and it should also work on other compilers.

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/sysdetect/tests/Makefile
+++ b/src/components/sysdetect/tests/Makefile
@@ -21,11 +21,9 @@ endif
 intel_compilers := ifort ifx
 cray_compilers := ftn crayftn
 
-FFLAGS += -ffree-form
-
 ifeq ($(notdir $(F77)),gfortran)
     FFLAGS +=-ffree-form -ffree-line-length-none
-else ifeq ($(notdir $(F77)),flang)
+else ifeq ($(patsubst %flang,,$(notdir $(F77))),)  # compiler name ends with flang
     FFLAGS +=-ffree-form
 else ifneq ($(findstring $(notdir $(F77)), $(intel_compilers)),)
     FFLAGS +=-free


tgamblin:
We're testing this on Cray here: https://github.com/spack/spack/pull/41886, so you can monitor that pipeline for the result.

We don't currently test PAPI 7.1.0 with ARM compilers, so maybe @gcongiu can verify whether this fixes the issue he tried to solve in #108.

tgamblin:
I'm not sure what to do with the CI failure here.  `papi_spack` is failing because this patch is already in Spack, and the patch fails to apply :). We don't currently have a good way to notice that. We've tried before:
* https://github.com/spack/spack/pull/26784

But unfortunately making patch application idempotent can cause us to silently fail to apply patches (unix patch does not give you a way to differentiate between the two cases) so we had to revert:
* https://github.com/spack/spack/pull/27625

Any suggestions?

gcongiu:
@tgamblin I would merge this PR anyway as the failure is not related to the patch (unless @jagode prefers otherwise). Maybe we could avoid this problem from happening again in the future by creating and testing PRs for the new PAPI versions in spack before releasing them? I am open to suggestions

tgamblin:
Actually, looking at your CI,  think I can fix this on the spack side if I just restrict the patch to 7.1.0. You’re building `papi@master` in CI, and this patch will be in 7.1.1, right?

gcongiu:
@tgamblin it will definitely be in the next release, but that will not necessarily be versioned as `7.1.1`.

tgamblin:
I should ask better questions. If I restrict the patch to 7.1.0, does that cover all the PAPIs we might see with the issue?

gcongiu:
@tgamblin if you restrict the patch to `papi@7.1.0` I think that should do it. The sysdetect component was introduced with papi@7.0.0 and papi@7.0.1 was released soon after to fix some gcc related link time optimisation problem with more recent gcc versions (IIRC).

tgamblin:
Ok, https://github.com/spack/spack/pull/41892 restricts the patch to 7.1.0 so CI should pass.

gcongiu (approved):

G-Ragghianti:
I've triggered a rerun of the papi-spack CI test.

G-Ragghianti:
It is fixed! A holiday miracle!

PR #155, opened 2024-01-30
SDE: Updated the C++ version of the tests.

adanalis:
The updated version of the example uses classes and looks like proper C++ much more than the previous version. However, the functionality has remained the same.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/sde/tests/Minimal/Minimal_Test++.cpp
+++ b/src/components/sde/tests/Minimal/Minimal_Test++.cpp
@@ -11,17 +11,17 @@
 //------- Library example that exports SDEs
 
 class MinTest{
-    private:
-        long long local_var;
+public:
+    MinTest();
+    void dowork();
 
-	public:
-        MinTest();
-        void dowork();
+private:
+    long long local_var;
 };
 
 MinTest::MinTest(){
-    papi_sde::PapiSde sde("Min Example Code in C++");
     local_var = 0;
+    papi_sde::PapiSde sde("Min Example Code in C++");
     sde.register_counter("Example Event", PAPI_SDE_RO|PAPI_SDE_DELTA, local_var);
 }
 
--- a/src/components/sde/tests/Minimal/Minimal_Test++.cpp
+++ b/src/components/sde/tests/Minimal/Minimal_Test++.cpp
@@ -6,27 +6,39 @@
 #include "sde_lib.h"
 #include "sde_lib.hpp"
 
-long long local_var;
 
-void mintest_init(void){
+////////////////////////////////////////////////////////////////////////////////
+//------- Library example that exports SDEs
+
+class MinTest{
+    private:
+        long long local_var;
+
+	public:
+        MinTest();
+        void dowork();
+};
+
+MinTest::MinTest(){
     papi_sde::PapiSde sde("Min Example Code in C++");
-
-    local_var=0;
+    local_var = 0;
     sde.register_counter("Example Event", PAPI_SDE_RO|PAPI_SDE_DELTA, local_var);
 }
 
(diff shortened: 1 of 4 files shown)


src/components/sde/tests/Minimal/Minimal_Test++.cpp
```diff
@@ -6,28 +6,40 @@
...
+
+	public:
+        MinTest();
+        void dowork();
+};
 
-    local_var=0;
+MinTest::MinTest(){
+    papi_sde::PapiSde sde("Min Example Code in C++");
+    local_var = 0;
```
Treece-Burgess, line 24: Very small suggestion, but I would change the order here of the code from:
```
papi_sde::PapiSde sde("Min Example Code in C++");
local_var = 0;
sde.register_counter("Example Event", PAPI_SDE_RO|PAPI_SDE_DELTA, local_var);
```
to 
```
local_var = 0;
papi_sde::PapiSde sde("Min Example Code in C++");
sde.register_counter("Example Event", PAPI_SDE_RO|PAPI_SDE_DELTA, local_var);
```
I believe it reads better. However, this is mainly just personal preference.

src/components/sde/tests/Minimal/Minimal_Test++.cpp
```diff
@@ -6,28 +6,40 @@
...
 
-long long local_var;
 
-void mintest_init(void){
-    papi_sde::PapiSde sde("Min Example Code in C++");
+////////////////////////////////////////////////////////////////////////////////
+//------- Library example that exports SDEs
+
+class MinTest{
+    private:
```
Treece-Burgess, line 14: Based on the [ICL style guide](https://github.com/icl-utk-edu/style/blob/master/ICL_C_CPP_Coding_Style_Guide.md) both `private:` and `public:` should not be indented.

src/components/sde/tests/Minimal/Minimal_Test++.cpp
```diff
@@ -6,28 +6,40 @@
...
+        long long local_var;
+
+	public:
```
Treece-Burgess, line 17: `public` comes before `private`.

src/components/sde/tests/Simple2/simple2.hpp
```diff
@@ -0,0 +1,19 @@
+#if !defined(SIMP2_LIB_H)
```
Treece-Burgess, line 1: Instead of `#if !defined(SIMP2_LIB_H)` you could just use `ifndef PAPI_SIMPLE2_H`.

src/components/sde/tests/Simple2/simple2.hpp
```diff
@@ -0,0 +1,19 @@
...
+#define SIMP2_LIB_H
+
+#include "sde_lib.h"
+#include "sde_lib.hpp"
+
+class Simple{
+    private:
```
Treece-Burgess, line 8: Same suggestion as with the class located in `Minimal_Test++.cpp`:

- `public` before `private`.
- `public` and `private` are not indented.

src/components/sde/tests/Simple2/simple2.hpp
```diff
@@ -0,0 +1,19 @@
...
+        double comp_value;
+        long long int total_iter_cnt, low_wtrmrk, high_wtrmrk;
+        static long long int counter_accessor_function( double *param );
```
Treece-Burgess, line 11: For the `private` section, methods go first and then attributes.

Treece-Burgess (approved):

PR #159, opened 2024-02-06
Adding support for optional individual file flag to papi_hl_output_writer.py

Treece-Burgess:
## Pull Request Description
This PR adds an additional flag to `papi_hl_output_writer.py` titled `source_file`. This allows a user the flexibility to pass either a directory such as `papi_hl_output` which contains one or more `.json` files or a single `.json` file such as `rank_082041.json`. 

A few examples will be outlined below.

- Case 1: Passing an individual file to `papi_hl_output_writer.py`:
```
## code 
python3 papi_hl_output_writer.py --source_file rank_082041.json --notation=derived --type=summary
## output
{
    "computation": {
        "Region count": 1,
        "Real time in s": 0.0,
        "IPC": 0.43
    }
}
```
- Case 2: Error handling if a user passes an individual file and a directory to `papi_hl_output_writer.py`:
```
## code 
python3 papi_hl_output_writer.py --source papi_hl_output --source_file papi_hl_output/rank_082041.json --notation=derived --type=summary
## error output 
Cannot pass values to both source_dir and source_file. Value must be passed to either source_dir or source_file.
usage: papi_hl_output_writer.py [-h] [--source_dir SOURCE_DIR] [--source_file SOURCE_FILE] [--format FORMAT] [--type TYPE] [--notation NOTATION]

options:
  -h, --help            show this help message and exit
  --source_dir SOURCE_DIR
                        Measurement directory of raw data.
  --source_file SOURCE_FILE
                        Individual file containing measurements of raw data.
  --format FORMAT       Output format, e.g. json.
  --type TYPE           Output type: detail or summary.
  --notation NOTATION   Output notation: raw or derived.
```
- Case 3: Error handling if a user does not pass an individual file or a directory to `papi_hl_output_writer.py`:
```
## code
python3 papi_hl_output_writer.py --notation=derived --type=summary
## error output
Path to either a JSON file (--source_file) or a dictionary (--source_dir) which contains a JSON file is required.
usage: papi_hl_output_writer.py [-h] [--source_dir SOURCE_DIR] [--source_file SOURCE_FILE] [--format FORMAT] [--type TYPE] [--notation NOTATION]

options:
  -h, --help            show this help message and exit
  --source_dir SOURCE_DIR
                        Measurement directory of raw data.
  --source_file SOURCE_FILE
                        Individual file containing measurements of raw data.
  --format FORMAT       Output format, e.g. json.
  --type TYPE           Output type: detail or summary.
  --notation NOTATION   Output notation: raw or derived.
```

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/high-level/scripts/papi_hl_output_writer.py
+++ b/src/high-level/scripts/papi_hl_output_writer.py
@@ -78,7 +78,7 @@ def merge_json_files(source):
   return json_object
 
 def parse_source_file(source_file):
-    """Take JSON file (source_file) and convert it to a python dictionary.
+    """Take a JSON file (source_file) and convert it to a python dictionary.
     Additionally add extra metadata such as mpi rank.
 
     Arguments:
--- a/src/high-level/scripts/papi_hl_output_writer.py
+++ b/src/high-level/scripts/papi_hl_output_writer.py
@@ -31,12 +31,12 @@ event_rate_names = OrderedDict([
       ('PAPI_DP_OPS','Double precision MFLOPS/s')
     ])
 
-def merge_json_files(source):
+def merge_json_files(source_dir):
   json_object = {}
   events_stored = False
 
   #get measurement files
-  file_list = os.listdir(source)
+  file_list = os.listdir(source_dir)
   file_list.sort()
   rank_cnt = 0
   json_rank = OrderedDict()
@@ -51,7 +51,7 @@ def merge_json_files(source):
       rank = rank_cnt
 
     #open measurement file
-    file_name = str(source) + "/" + str(item)
+    file_name = str(source_dir) + "/" + str(item)
 
     try:
       with open(file_name) as json_file:
@@ -78,14 +78,6 @@ def merge_json_files(source):
   return json_object
 
 def parse_source_file(source_file):
-    """Take a JSON file (source_file) and convert it to a python dictionary.
-    Additionally add extra metadata such as mpi rank.
-
-    Arguments:
-        source_file -- JSON file generated from PAPI high level API functions.
-    Returns:
-        Python dictionary.
-    """
     json_data = {}
     json_rank = OrderedDict()
     events_stored = False
@@ -506,10 +498,10 @@ def write_json_file(data, file_name):
     outfile.write(to_unicode(str_))
     print (str_)
 
-def main(format, type, notation, source = None, source_file = None):
+def main(format, type, notation, source_dir = None, source_file = None):
     if (format == "json"):
-        if source != None:
-            json = merge_json_files(source)
+        if source_dir != None:
+            json = merge_json_files(source_dir)
         else:
             json = parse_source_file(source_file)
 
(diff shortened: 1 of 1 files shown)


src/high-level/scripts/papi_hl_output_writer.py
```diff
@@ -75,10 +75,45 @@ def merge_json_files(source):
   
   json_object['ranks'] = json_rank
 
-  # print json.dumps(json_object,indent=2, sort_keys=False,
-  #                  separators=(',', ': '), ensure_ascii=False)
   return json_object
 
+def parse_source_file(source_file):
+    """Take JSON file (source_file) and convert it to a python dictionary.
```
Treece-Burgess, line 81: I added a Python doc string to the function I created titled `parse_source_file`, I think adding them to the other individual functions could be beneficial for two reasons:

1. Adding the doc strings can help speed up development. A doc string would help with understanding what each function is doing rather than having to walk through a function with no prior background.
2. If a user wanted to look at the contents of a function without going to the actual file they could run the following: 
```
# code 
from papi_hl_output_writer.py import parse_source_file
help(parse_source_file)
# output
Take a JSON file (source_file) and convert it to a python dictionary.
Additionally add extra metadata such as mpi rank.
    
Arguments:
    source_file -- JSON file generated from PAPI high level API functions.
Returns:
    Python dictionary.
```

This is nothing major, but I wanted to mention it.

src/high-level/scripts/papi_hl_output_writer.py
```diff
@@ -471,45 +506,71 @@ def write_json_file(data, file_name):
...
+                write_json_file(sum_json_object(json, True), 'papi_sum.json')
+            else:
+                write_json_file(sum_json_object(json), 'papi_sum.json')
+    else:
+        print("Format not supported!")
 
 
 def parse_args():
   parser = argparse.ArgumentParser()
-  parser.add_argument('--source', type=str, required=False, default="papi_hl_output",
```
Treece-Burgess, line 498: One suggestion is we could change `--source` to `--source_dir`. This would add more distinction to what a user is passing and it would keep inline with `--source_file`.  Nothing major, and we definitely could keep it as `source` if preferred.

Treece-Burgess: Changed `--source` to `--source_dir` in latest commit.

src/high-level/scripts/papi_hl_output_writer.py
Treece-Burgess: It does look possible to add `doxygen` to a Python file, I will work on creating an example.

src/high-level/scripts/papi_hl_output_writer.py
```diff
@@ -471,46 +498,72 @@ def write_json_file(data, file_name):
...
+                write_json_file(derive_json_object(json), 'papi.json')
+            else:
+                write_json_file(json, 'papi.json')
 
-  else:
-    print("Format not supported!")
+        #summarize data over regions with the same name, threads and ranks
+        if type == 'summary':
+            if notation == 'derived':
+                write_json_file(sum_json_object(json, True), 'papi_sum.json')
```
Treece-Burgess, line 517: Currently the output file is hardcoded to just be `papi_sum.json`, we possibly could add another flag that allows a user to pass an output filename.

jagode (approved):

PR #176, opened 2024-03-28
SDE_LIB: Allow PAPI_reset() to reset a CountingSet.

adanalis:
## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/sde_lib/sde_lib_misc.c
+++ b/src/sde_lib/sde_lib_misc.c
@@ -501,6 +501,14 @@ sdei_read_and_update_data_value( sde_counter_t *counter, long long int previous_
         SDEDBG("Reading %s by calling registered function pointer.\n", event_name);
         tmp_int = counter->u.cntr_cb.callback(counter->u.cntr_cb.param);
         tmp_data = &tmp_int;
+    }else if( IS_CNTR_CSET(counter) ){
+        if( 0 == previous_value ){
+            SDEDBG("Resetting CountingSet %s by freeing all the elements it contains.\n", event_name);
+            return cset_delete(counter->u.cntr_cset.data);
+        }else{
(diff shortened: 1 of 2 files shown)


src/sde_lib/sde_lib_misc.c
```diff
@@ -501,6 +501,14 @@ sdei_read_and_update_data_value( sde_counter_t *counter, long long int previous_
         SDEDBG("Reading %s by calling registered function pointer.\n", event_name);
         tmp_int = counter->u.cntr_cb.callback(counter->u.cntr_cb.param);
         tmp_data = &tmp_int;
+    }else if( IS_CNTR_CSET(counter) ){
+        if( 0 == previous_value ){
```
Treece-Burgess, line 505: @adanalis For `sdei_read_and_update_data_value` to reset for a CountingSet, a user would have to explicitly pass a value of 0 when calling `sdei_read_and_update_data_value`? E.g:
```
sdei_read_and_update_data_value( counter, 0, &(counter->previous_data) )
```

adanalis: sdei_read_and_update_data_value() is an internal function (that's why there is the "i" in sdei), so it should not be called by end-users. The check of "previous_value" being zero is a sanity check and future-proofing of the code. In the current version of the code it is impossible for sdei_read_and_update_data_value() to be invoked for a counter of type CSET with a non-zero previous_value.

Treece-Burgess: I see, thank you for clarifying the i within `sdei_read_and_update_data_value()`, I am familiar with using  `_` for internal functions but not i.

Treece-Burgess (approved):

PR #187, opened 2024-04-29
Update Allowed Modifiers for PAPI_enum_event 

Treece-Burgess:
## Pull Request Description
Pull requests updates `PAPI_enum_event(int * EventCode, int  modifer)` to: 

1. Update documentation to show modifiers that are allowed to be passed by a user
2. Add conditional check to make sure a valid modifier is provided

No internal code utilizes the ITANIUM/Power modifiers. Checked with Github's search functionality.

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/papi.c
+++ b/src/papi.c
@@ -1664,34 +1664,6 @@ PAPI_event_name_to_code( const char *in, int *out )
  *	The following values are implemented for preset events
  *	<ul>
  *         <li> PAPI_PRESET_ENUM_AVAIL -- enumerate only available presets
- *         <li> PAPI_PRESET_ENUM_MSC   -- Miscellaneous preset events
- *         <li> PAPI_PRESET_ENUM_INS   -- Instruction related preset events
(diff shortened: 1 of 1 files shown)


Treece-Burgess:
@jagode @adanalis I currently added the modifier conditional check to be inside 
`if (IS_PRESET(i)) { ... }` due to prior discussion only being about preset modifiers. 

From today's PICL meeting it was mentioned briefly to have the modifiers conditional check at the very beginning, so right before `if (IS_PRESET(i)) { ... }`. This could then capture a modifier check for generic, native, and preset modifiers. However, we still would need to account for itanium and power modifiers.

How do we want to handle itanium and power modifiers? We could possibly remove them or have a modifier check inside both `if (IS_PRESET(i)) { ... }` and `if (IS_NATIVE(i)) { ... }`.

adanalis (approved):

PR #195, opened 2024-06-25
Add Note in Doxygen Regarding Values Array in PAPI_accum() 

Treece-Burgess:
## Pull Request Description
Unlike for `PAPI_read()` and `PAPI_stop()` where the `values` array is only initialized (written) in these functions. `PAPI_accum()` must have the `values` array initialized, because its values are read in.

Due to not being able to check to see if the `values` array is initialized, we will document in the Doxygen that the `values` array must be initialized.

Programmer's Manual User Perspective (Doxygen):
Screenshot 2024-06-27 at 2 37 00 PM

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/papi.c
+++ b/src/papi.c
@@ -3055,8 +3055,10 @@ PAPI_reset( int EventSet )
  *
  *  The counters continue counting after the read. 
  *
- *  Note the differences between PAPI_read() and PAPI_accum(), specifically
- *  that PAPI_accum() resets the values array to zero.
+ *  Note the differences between PAPI_read() and PAPI_accum(). Specifically,
+ *  PAPI_accum() adds the counter values to the values stored
+ *  in the array (the second parameter in PAPI_accum()) and then resets the counters
+ *  to zero.
  *
  *  PAPI_read() assumes an initialized PAPI library and a properly added 
  *  event set. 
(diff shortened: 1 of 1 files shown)


src/papi.c
```diff
@@ -3056,7 +3056,9 @@ PAPI_reset( int EventSet )
  *  The counters continue counting after the read. 
  *
  *  Note the differences between PAPI_read() and PAPI_accum(), specifically
- *  that PAPI_accum() resets the values array to zero.
+ *  that PAPI_accum() adds the current counter values to the existing counter
+ *  values held in the provided array (second parameter in PAPI_accum()) and resets the
```
jagode, line 3060: Note that the values stored in the array don't have to be "existing counter values" but can be any arbitrary values provided by the user.

I suggest changing the wording to the following:

"Note the differences between PAPI_read() and PAPI_accum(). Specifically, PAPI_accum() adds the counter values to the values stored in the array (the second parameter in PAPI_accum()) and then resets the counters to zero."

src/papi.c
```diff
@@ -3246,7 +3248,12 @@ PAPI_read_ts( int EventSet, long long *values, long long *cycles )
  *	PAPI_accum adds the counters of the indicated event set into the array values. 
  *	The counters are zeroed and continue counting after the operation.
  *	Note the differences between PAPI_read and PAPI_accum, specifically 
- *	that PAPI_accum resets the values array to zero. 
+ *	that PAPI_accum adds the current counter values to the existing counter values
```
jagode, line 3251: Same comment as above:

Note that the values stored in the array don't have to be "existing counter values" but can be any arbitrary values provided by the user.

I suggest changing the wording to the following:

"Note the differences between PAPI_read() and PAPI_accum(). Specifically, PAPI_accum() adds the counter values to the values stored in the array (the second parameter in PAPI_accum()) and then resets the counters to zero."

jagode (changes requested):

src/papi.c
Treece-Burgess: I have gone ahead and updated both sections of text with your suggested changes!

jagode (approved):

jagode (approved):

jagode (approved):

PR #199, opened 2024-06-27
utils: Include the compiler flags when linking the PAPI utilities

wcohen:
Newer Linux distributions are using LTO to generate the binaries rather than standard linking.  Should include the flags used to generate the .o files to make it clearer what optimization should be done in LTO when generating the final binaries.

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**


adanalis:
Will, can you please rebase this PR?

wcohen:
Fat fingers. It looks like it did a merge rather than a rebase. for this pull request. Let me rework this to properly rebase the patch.

wcohen:
Okay reapplied the patch to the current master branch.  Could you see if this looks okay to you once the checks are done.

adanalis (approved):

Issue #200, opened 2024-06-28
perf_event_uncore doesn't consider capabilities


voidbert:
The `perf_event_uncore` component checks if the user has permissions for reading uncore counters before trying to do so. It does so by running the [this code](https://github.com/icl-utk-edu/papi/blob/224be80dbaa7be512ab156704b3c2640772cb023/src/components/perf_event_uncore/perf_event_uncore.c#L666), abridged below:

```c
// read /proc/sys/kernel/perf_event_paranoid
// ...

if ((paranoid_level>0) && (getuid()!=0)) {
    // fail due to lack of permissions
}

// Success
```

However, checking for `root` is too strong of a check, as it's possible for a regular user to be able to read counters even with `perf_event_paranoid > 0`. To achieve that, the current thread must have one of the following capabilities: `CAP_SYS_ADMIN` or `CAP_PERFMON`.

jagode:
@deater : Vince, I reviewed the PR #201 , and it looks very reasonable. However, before approving it---and since you are most familiar with the perf_event_uncore component---could you please take a look as well and share your feedback?  Thanks!

deater:
this change looks reasonable.

for future work, maybe a similar change is needed for the perf_event component too.  The perf_event and perf_event_component code has drifted apart over the years even though they started out with very similar code (both components deal directly with the underlying Linux perf_event interface).  I started working on a project to re-sync or even merge some of the code involved to make it easier to maintain but haven't had a chance to finish it.

PR #201, opened 2024-06-28
perf_event_uncore: consider capabilities for permissions

voidbert:
## Pull Request Description

Fixed issue #200 by reading capabilities using the `capget` system call. Now, users other than `root` can use PAPI to read uncore counters, provided they have one of the following capabilities: `CAP_SYS_ADMIN` or `CAP_PERFMON`.

#### Some rationale and future possibilities

 - I opted to use the `capget` system call directly, not to add `libcap` as another dependency.

 - I'd like to ask if there's any reason why system call numbers are manually defined. Would there
   be any issue arising from the use of the `sys/syscall.h` header?

## Author Checklist
- [X] **Description**
- [X] **Commits**
- [X] **Tests**

### DIFF
--- a/src/components/perf_event_uncore/perf_event_uncore.c
+++ b/src/components/perf_event_uncore/perf_event_uncore.c
@@ -22,6 +22,7 @@
 #include <sys/utsname.h>
 #include <sys/mman.h>
 #include <sys/ioctl.h>
+#include <linux/capability.h>
 
 /* PAPI-specific includes */
 #include "papi.h"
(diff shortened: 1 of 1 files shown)


jagode:
@deater : Vince, I reviewed the PR that addresses issue #200, and it looks very reasonable. However, before approving it---and since you are most familiar with the perf_event_uncore component---could you please take a look as well and share your feedback?  Thanks!

jagode (approved):

PR #219, opened 2024-08-21
rocm: add reason for disabled component

dbarry9:
## Pull Request Description
Previously, in the absence of a ROCm device, the rocm component did not set the string containing the reason that the component was disabled. This corresponds to Issue https://github.com/icl-utk-edu/papi/issues/220.

These changes have been tested with ROCm 6.3.1 on Frontier and ROCm 6.3.2 on a system with no ROCm devices.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/rocm/roc_common.c
+++ b/src/components/rocm/roc_common.c
@@ -35,6 +35,11 @@ rocc_init(void)
 
     hsa_status_t status = hsa_init_p();
     if (status != HSA_STATUS_SUCCESS) {
+        int errMsgLen = snprintf(error_string, PAPI_MAX_STR_LEN, "%s",
+            "Call to hsa_init() failed. This could be due to ROCm devices failing to be detected.");
+        if( errMsgLen < 0 || errMsgLen >= PAPI_MAX_STR_LEN ) {
+            SUBDBG("error_string was truncated.\n");
+        }
         papi_errno = PAPI_EMISC;
         goto fn_fail;
     }
--- a/src/components/rocm/rocm.c
+++ b/src/components/rocm/rocm.c
@@ -221,16 +221,22 @@ rocm_init_private(void)
     papi_errno = evt_get_count(&count);
     _rocm_vector.cmp_info.num_native_events = count;
     _rocm_vector.cmp_info.num_cntrs = count;
+    _rocm_vector.cmp_info.initialized = 1;
+    int strLen = snprintf(_rocm_vector.cmp_info.disabled_reason, PAPI_MAX_STR_LEN, "%s", "");
+    if (strLen < 0 || strLen >= PAPI_MAX_STR_LEN) {
+        SUBDBG("Failed to fully write disabled_reason.\n");
+    }
+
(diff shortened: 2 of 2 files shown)


Treece-Burgess (approved):
I personally would add a format specifier to `snprintf`, but I don't believe it is actually necessary in this case or beneficial since we are providing the error message. 

The rest of the file seems to follow suit with no format specifier when we directly provide the error message as well.

src/components/rocm/roc_common.c
```diff
@@ -35,6 +35,7 @@ rocc_init(void)
 
     hsa_status_t status = hsa_init_p();
     if (status != HSA_STATUS_SUCCESS) {
+        snprintf(error_string, PAPI_MAX_STR_LEN, "Call to hsa_init() failed.");
```
Treece-Burgess, line 38: Two things:

- The format specifier is missing
- I would check the return value of `snprintf` as an additional safety net e.g: 
```
int len = snprintf(error_string, PAPI_MAX_STR_LEN, "%s", "Call to hsa_init() failed.");
if (len < 0 || len >= PAPI_MAX_STR_LEN) {
    SUBDBG("Failed to write error string.\n");
}
```

src/components/rocm/rocm.c
```diff
@@ -225,7 +225,6 @@ rocm_init_private(void)
   fn_exit:
     _rocm_vector.cmp_info.initialized = 1;
     _rocm_vector.cmp_info.disabled = papi_errno;
-    strcpy(_rocm_vector.cmp_info.disabled_reason, "");
```
Treece-Burgess, line 228: Why remove this line?

Treece-Burgess (changes requested):

src/components/rocm/rocm.c
dbarry9: This line was overwriting the error message with an empty string; I am not sure why it was here.

src/components/rocm/roc_common.c
dbarry9: I have updated the PR according to this feedback.

Treece-Burgess (approved):
I have verified the behavior on a machine with and without ROCm devices using ROCm 6.3.2.

Without:
```
./papi_component_avail
Name:   rocm                    GPU events and metrics via AMD ROCm-PL API
   \-> Disabled: Call to hsa_init() failed. This could be due to ROCm devices failing to be detected.

```
```
Application code test with PAPI_get_component_info:
Disabled reason: Call to hsa_init() failed. This could be due to ROCm devices failing to be detected.
Disabled: -14
Initialized: 0
```

With (1 * MI210):
```
./papi_component_avail
Name:   rocm                    GPU events and metrics via AMD ROCm-PL API
                                Native: 412, Preset: 0, Counters: 412
```
```
Application code test with PAPI_get_component_info:
Disabled reason: 
Disabled: 0
Initialized: 1
```

PR #228, opened 2024-08-30
Beta support for AMD ROCprofiler-SDK events.

adanalis:
## Pull Request Description

This component offers early support for AMD GPU/APU events through the new ROCprofiler-SDK library. Due to limitations in the vendor library, the component has the following limitations:
A) Only the dispatch mode is currently supported, which means that events can only be read after a kernel has completed, not during the execution of the kernel.
B) Event dimensions, such as "instance", or "shader_engine" are currently aggregated and can't be accessed individually.

### DIFF
--- a/src/components/rocp_sdk/Rules.rocp_sdk
+++ b/src/components/rocp_sdk/Rules.rocp_sdk
@@ -1,0 +1,14 @@
+COMPSRCS += components/rocp_sdk/rocp_sdk.c
+
+COMPOBJS += rocp_sdk.o sdk_class.o
+
+ROCP_SDK_INCL=$(PAPI_ROCP_SDK_ROOT)/include
+
+CFLAGS  += -g -I$(ROCP_SDK_INCL) -D__HIP_PLATFORM_AMD__
+LDFLAGS += $(LDL)
+
+rocp_sdk.o: components/rocp_sdk/rocp_sdk.c $(HEADERS)
+	$(CC) $(LIBCFLAGS) $(OPTFLAGS) -c $< -o $@
+
+sdk_class.o: components/rocp_sdk/sdk_class.cpp $(HEADERS)
+	$(CXX) $(LIBCFLAGS) $(OPTFLAGS) -c $< -o $@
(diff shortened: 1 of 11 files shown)


Treece-Burgess (approved):

bertwesarg:
Sorry for being late, but what is the long term plan regarding the component name? `rocp_sdk` is probably hard to argument to users. IIRC the metric names should not change between the `rocm` (rocprofiler V1 API) component and the `rocp_sdk` (rocprofiler SDK API) component. Is it planned to rename the `rocp_sdk` component back to `rocm` after the rocprofile V1 API was phased out (i.e., deprecated by AMD)? Or is it even possible to use the `rocp_sdk` component as in drop in without the need to change the `<component>:::` prefix to metric names? Maybe rename the current `rocp_sdk` already to something more similar to `rocm`, so that the finale rename is not that disturbing to the users?. Something like `rocm-ng`, `rocm-new`, `rocm-beta`, …?

bertwesarg:
> Or is it even possible to use the `rocp_sdk` component as a drop in without the need to change the `<component>:::` prefix to metric names?

Would something like this be accepted?

```diff
diff --git i/src/components/rocp_sdk/rocp_sdk.c w/src/components/rocp_sdk/rocp_sdk.c
index 1a3e107d5..a12edffab 100644 src/components/rocp_sdk/rocp_sdk.c
--- i/src/components/rocp_sdk/rocp_sdk.c
+++ w/src/components/rocp_sdk/rocp_sdk.c
@@ -97,6 +97,11 @@ static int check_n_initialize(void);
 int
 rocp_sdk_init_component(int cid)
 {
+    if (getenv("PAPI_RPSDK_FAKE_ROCM_COMPONENT")) {
+        _rocp_sdk_vector.cmp_info.name = "rocm";
+        _rocp_sdk_vector.cmp_info.short_name = "rocm";
+    }
+
     _rocp_sdk_vector.cmp_info.CmpIdx = cid;
     _rocp_sdk_vector.cmp_info.num_native_events = -1;
     _rocp_sdk_vector.cmp_info.num_cntrs = -1;

adanalis:
Bert, why is it problematic if the new component has a different name? Doing so enables users to access events through both components during the transition period since rocprofiler SDK is still immature, and the old API will still work for older GPUs.

bertwesarg:
Sure, I was talking about the long term plan. Eventually, the old rocprofiler API wont exists anymore. Will the component be renamed then?

PR #123, opened 2023-11-16
ROCm event id refactor

gcongiu:
## Pull Request Description
This PR introduces event qualifier support in the ROCm component.

Also, solves issue #100

Main changes introduced by this PR are for `papi_component_avail`, that now shows a significantly smaller number of ROCm component events (these are the unique event names without any `:device=` or `:instance=` addition to the name) and `papi_native_avail`, that now shows `:device=` and `:instance=` as event qualifiers. Following an example for a system with two AMD GPUs:

```
$ utils/papi_component_avail
Available components and hardware information.
--------------------------------------------------------------------------------
PAPI version             : 7.0.1.0
Operating system         : Linux 6.1.62-1.el9.elrepo.x86_64
Vendor string and code   : AuthenticAMD (2, 0x2)
Model string and code    : AMD EPYC 7413 24-Core Processor (1, 0x1)
CPU revision             : 1.000000
CPUID                    : Family/Model/Stepping 25/1/1, 0x19/0x01/0x01
CPU Max MHz              : 3630
CPU Min MHz              : 1500
Total cores              : 96
SMT threads per core     : 2
Cores per socket         : 24
Sockets                  : 2
Cores per NUMA region    : 48
NUMA regions             : 2
Running in a VM          : no
Number Hardware Counters : 5
Max Multiplex Counters   : 384
Fast counter read (rdpmc): yes
--------------------------------------------------------------------------------

Compiled-in components:
Name:   perf_event              Linux perf_event CPU counters
Name:   perf_event_uncore       Linux perf_event CPU uncore and northbridge
Name:   rocm                    GPU events and metrics via AMD ROCm-PL API
Name:   sysdetect               System info detection component

Active components:
Name:   perf_event              Linux perf_event CPU counters
                                Native: 145, Preset: 27, Counters: 5
                                PMUs supported: perf, perf_raw, amd64_fam19h_zen3

Name:   perf_event_uncore       Linux perf_event CPU uncore and northbridge
                                Native: 4, Preset: 0, Counters: 7
                                PMUs supported: amd64_rapl, amd64_fam19h_zen3_l3

Name:   rocm                    GPU events and metrics via AMD ROCm-PL API
                                Native: 401, Preset: 0, Counters: 401

Name:   sysdetect               System info detection component
                                Native: 0, Preset: 0, Counters: 0


--------------------------------------------------------------------------------

$ utils/papi_native_avail
...
 Native Events in Component: rocm
===============================================================================
[...]

PR #236, opened 2024-09-17
Cuda Component Refactor to Add Device Qualifier

Treece-Burgess:
## Pull Request Description
This pull request addresses adding a device qualifier to the Cuda component. This refactor is similar to the ROCm refactor which added an instance and device qualifier (PR #123).

### Sample Output for `./papi_component_avail`
```
Name:   cuda                    CUDA profiling via NVIDIA CuPTI interfaces
                                Native: 141834, Preset: 0, Counters: 30
```

### Sample Output for `./papi_native_avail`
```
===============================================================================
| cuda:::dram__bytes.avg                                                       |      
|            # of bytes accessed in DRAM. Units=(bytes) Numpass=1              |      
|            Mandatory device qualifier [0,1,2,3,4,5,6,7]                      |      
```

Note: As of Cuda Version 12.6, the MetricsContext API has been removed and replaced with the MetricsEvaluator API. Due to this change, the Cuda component will only work with Cuda versions < 12.6. The PAPI team will be working to support both the MetricsContext API and MetricsEvaluator API. 

To prevent the above note from occurring a conditional has been placed inside `Rules.cuda`, where if a user uses a Cuda version >= 12.6, then the build will error out and a message will print. From my testing this will not interfere with `make clean`. Output can be seen below:
```
components/cuda/Rules.cuda:11: *** In Cuda 12.6, the MetricsContext API was replaced with the MetricsEvaluator API. Due to this, the Cuda component is currently being refactored to support Cuda versions >= 12.6.  Stop.
```

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**


Treece-Burgess:
I am doing final cleanups/checks, but I will squash all the above commits into a single commit before merge. The merge commit message will have a note regarding the Cuda component not currently working with versions >= 12.6.

Treece-Burgess:
As stands, I have tested with:

1. Personal scripts to start, stop, read, reset, cleanup, and more.
2. The following tests in the test directory: simpleMultipGPU, HelloWorld, and test_multi_read_and_reset.

From the above testing, no errors were encountered.

dbarry9 (approved):
I have tested these changes on the NVIDIA Hopper architecture using floating-point benchmarks. I monitored the correct event counts for half-, single-, and double-precision FMA operations on the device.

jagode:
@Treece-Burgess : Since it has been approved by Daniel, please go ahead and merge it so the community can test it before the next release.

PR #238, opened 2024-09-17
Rdpmc topdown

### LINKED ISSUE #207: Negative/large values in topdown perf counters
- System: Intel Sapphire Rapids 8480+
- OS: Red Hat Enterprise Linux 9.2
- PAPI Version: 7.1.0

I am trying to read topdown counters that are reported to be available in the system above:
```
$ papi_native_avail | grep topdown
| perf::topdown-retiring                                                       |
| perf::topdown-bad-spec                                                       |
| perf::topdown-fe-bound                                                       |
| perf::topdown-be-bound                                                       |
```
Going by the description of these metrics (`must be used in a group with the other topdown- events with slots as leader`), I created an EventSet with `perf::slots` as the first event in the set. However, the values I obtain through `PAPI_read` are negative. I assumed this could be due to an overflow based on the granularity of the workload.

Investigating further, I found that the perf utility seems to give reasonable results when polling these counters:
```
$ perf stat --topdown -I1000 --td-level=1 --no-metric-only
#           time             counts unit events
     1.001028171          670042878      slots
     1.001028171          102988906      topdown-retiring                 #     15.4% Retiring
     1.001028171           18685968      topdown-bad-spec                 #      2.8% Bad Speculation
     1.001028171           49436879      topdown-fe-bound                 #      7.4% Frontend Bound
     1.001028171          499046991      topdown-be-bound                 #     74.5% Backend Bound
     1.001028171           14145047      topdown-heavy-ops
     1.001028171           15961096      topdown-br-mispredict
     1.001028171           28771199      topdown-fetch-lat
     1.001028171          406127949      topdown-mem-bound
```
However when attempting to read them through the `papi_command_line` utility, I get similar results compared to when I poll them manually through my library. For instance, trying to read the 5 events at once reports the following:
```
$ papi_command_line -x perf::slots perf::topdown-bad-spec perf::topdown-retiring perf::topdown-be-bound perf::topdown-fe-bound

Successfully added: perf::slots
Successfully added: perf::topdown-bad-spec
Successfully added: perf::topdown-retiring
Successfully added: perf::topdown-be-bound
Successfully added: perf::topdown-fe-bound

perf::slots : 	0XFFFF8000394F75D3
perf::topdown-bad-spec : 	0X5CA000035
perf::topdown-retiring : 	0X5CA000035
perf::topdown-be-bound : 	0X5CA000035
perf::topdown-fe-bound : 	0X5CA000035
----------------------------------
```
What's strange is if I only poll `perf::slots`, the reported numbers seem to correspond to the lower part of the number reported when polling all of them at the same time:
```
$ papi_command_line -x perf::slots

Successfully added: perf::slots

perf::slots : 	0X3951DCF8
```
Which leads me to believe perhaps this counter is reading extra data. Moreover, the 4 categories seem to consistently give the same value using the utility, and it seems rather large. Are these reported values supposed to be percentages if PAPI obtains them directly through the processed metrics given by perf? Or are they metrics that have to be combined with masks to obtain the percentages of each category?

Based on the event descriptions (for instance, for the `perf::topdown-retiring` event: "topdown useful slots retiring uops"), I expected integer values such as the ones reported by the `perf stat` utility that I could then correlate to the `slots` value to obtain the percentages myself.

willowec:
## Pull Request Description

This PR fixes #207

Intel's topdown events, if opened using perf_event's `perf_event_open()` behave differently when read using `read()` than when they are read using `rdpmc()`. This has a few unfortunate consequences. When a topdown event is read using `read()` you get a value that is the number of slots recorded for that topdown category. So to get, for example, the percentage of slots that were retiring, you divide the value of `TOPDOWN:RETIRING_SLOTS` by the value of `TOPDOWN:SLOTS` and multiply the result by 100.

However, when any topdown event is read using `rdpmc()`, the result is always one value containing eight 8-bit percentages, no matter which event was read:

```
     | mem_bound | fetch_lat | br_mispred | heavy_ops | be_bound | fe_bound | bad_spec | retiring
0x   | XX        | XX        | XX         | XX        | XX       | XX       | XX       | XX
```
In order to extract a percentage from this, you would instead do some shifting and masking as described in [tools/perf/Documentation/topdown.txt](https://github.com/torvalds/linux/blob/master/tools/perf/Documentation/topdown.txt)

This means that storing and accumulating the values of topdown events makes sense only when `rdpmc` is disabled in PAPI. In order to account for this, an additional `metric` field was added to `perf_event` component's `pe_event_info_t` struct, and is set during initialization for events matching the `TOPDOWN` base event. Events with this field set no longer have their reset flag cleared when `rdpmc` is enabled.

A new validation test, `topdown_validation.c` is added to verify that the results PAPI obtains agree with a direct call to `rdpmc()`

Because this patch was developed on an Intel Raptorlake machine, some other additions were also necessary to enable the development of this PR:
- Missing topdown events were added to the adl_glc events header file in libpfm4 (we are still waiting to hear back having this patch added to libpfm4 officially)
- Preset events were added to papi_events.csv for the adl_glc architecture

## Author Checklist
- [X] **Description**
- [X] **Commits**
- [X] **Tests**


willowec:
Converting to a draft as not all of the tests are passing for raptorlake

willowec:
Un-drafting as the failing tests are in the sysdetect component and are due to the heterogeneity of the Raptorlake architecture, unrelated to the contributions of this patch.

PR #249, opened 2024-09-24
ROCP_SDK: Enabling agent profiling mode.

adanalis:
The agent profiling mode (sampling) is now fully supported by the component. To activate it the user must set the environment variable RPSDK_MODE_AGENT_PROFILE=1.
However, the counter values returned by Rocprofiler-SDK when agent profiling is activated appear to be spurious.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/rocp_sdk/sdk_class.cpp
+++ b/src/components/rocp_sdk/sdk_class.cpp
@@ -184,7 +184,7 @@ obtain_function_pointers()
     if( !first_time )
         return NULL;
 
-    const char* pathname = std::getenv("ROCP_SDK_LIB");
+    const char* pathname = std::getenv("PAPI_ROCP_SDK_LIB");
 
     // If the user gave us an explicit path to librocprofiler-sdk.so, use it.
     if ( nullptr != pathname && strlen(pathname) <= PATH_MAX ) {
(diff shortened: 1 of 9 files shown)


src/components/rocp_sdk/sdk_class.cpp
```diff
@@ -973,6 +942,12 @@ int setup() {
     char *error_msg = NULL;
     int status = 0;
 
+    rpsdk_profiling_mode = RPSDK_MODE_CALLBACK_DISPATCH;
+    if( NULL != getenv("RPSDK_MODE_AGENT_PROFILE") ){
```
bertwesarg, line 946: this seems to be a PAPI only env var, better to prefix it with `PAPI_`?

adanalis:
In the latest version of the code, the default mode is sampling. To enable dispatch mode, the user must set the environment variable PAPI_RPSDK_DISPATCH_MODE.

bertwesarg:
> In the latest version of the code, the default mode is sampling. To enable dispatch mode, the user must set the environment variable PAPI_RPSDK_DISPATCH_MODE.

The only other PAPI rocprofiler-sdk related env var is named `PAPI_ROCP_SDK_ROOT`. Can they use the same prefix?

src/components/rocp_sdk/sdk_class.hpp
```diff
@@ -2,8 +2,10 @@
 #define __VENDOR_PROFILER_V1_H__
```
bertwesarg, line 2: this and `sdk_class.h` are still using the **same** component template include guard

PR #254, opened 2024-10-04
high-level: Explicitly use python3

wcohen:
Python3 has been released for over 15 years.  Fedora, Debian, and other distributions state that python scripts should avoid using unversioned python intepreter:

https://fedoraproject.org/wiki/FinalizingFedoraSwitchtoPython3 https://www.debian.org/doc/packaging-manuals/python-policy/

The RPM build process will flag unversioned python interpreter use as an error and the build will fail.  Making the papi_hl_output_writer.py explicitly use /usr/bin/python3.

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/high-level/scripts/papi_hl_output_writer.py
+++ b/src/high-level/scripts/papi_hl_output_writer.py
@@ -1,10 +1,10 @@
+#!/usr/bin/env python3
 ## 
 # @file papi_hl_output_writer.py
 # @brief Converts HL output to be more comprehensible. 
 # Output is enhanced by creating derived metrics like IPC,
 # MFlop/s, and MFlips/s. As well as real and processor time.
 
-#!/usr/bin/python
 from __future__ import division
 from collections import OrderedDict
 


Treece-Burgess:
@wcohen I think updating the shebang does make sense. What is your reasoning behind using `#!/usr/bin/python3` instead of `#!/usr/bin/env python3`? 

`#!/usr/bin/env python3` has the pros of looking through the PATH for the Python directory, which makes the script overall more portable. Virtual environments have became increasingly more popular with even Oak Ridge pushing users to create a virtual environment if they have specific Python packages they want to use. With `#!/usr/bin/python3` it assumes the Python interpreter is located there on each machine.

wcohen:
I am open to  `#!/usr/bin/env python3`.  The original patch was a minimal change to ensure that things build on Fedora without patching. Fedora (https://fedoraproject.org/wiki/FinalizingFedoraSwitchtoPython3) mentions  `#!/usr/bin/env` _name_ and is okay with it.  However, Debian (https://www.debian.org/doc/packaging-manuals/python-policy/) cautions against it in section 3.5.2:

> Maintainers should not override the Debian Python interpreter using /usr/bin/env name. This is not advisable as it bypasses Debian’s dependency checking and makes the package vulnerable to incomplete local installations of Python.

wcohen:
I have a branch with the `#!/usr/bin/env python3`.  I could update the pull request with that patch if it is preferred.

Treece-Burgess:
@wcohen I have access to a machine with a Debian release version 12 and I am working on getting the correct permissions to build PAPI. Once I do, I will run a few tests with the shebang `#!/usr/bin/env python3` to see the behavior. If there are no apparent issues then we will go with `#!/usr/bin/env python3`. However, if there are issues then we go with your initial suggestion of `#!/usr/bin/env`.

Treece-Burgess:
@wcohen I have tested on the machine with a Debian release version 12 and did not receive any warning output or errors when utilizing the `#!/usr/bin/env python3` shebang with the `papi_hl_output_writer.py`. So we will move forward with that option. I noticed that you already have `#!/usr/bin/env python3` as the change, but could you rebase one last time before I merge?

PR #257, opened 2024-10-10
perf_event: Eliminate paranoia error for check_exclude_guest()

willowec:
## Pull Request Description

On systems with the kernel setting `perf_event_paranoid`>=2, userspace calls to `perf_event_open()` must set `attr.exclude_kernel=1`. Set `exclude_kernel=1` in `check_exclude_guest()` to allow the function to execute successfully on systems with `perf_event_paranoid`>=2;

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/components/perf_event/perf_event.c
+++ b/src/components/perf_event/perf_event.c
@@ -285,6 +285,7 @@ check_exclude_guest( void )
 	/* First check that we can open a plain instructions event */
 	memset(&attr, 0 , sizeof(attr));
 	attr.config = PERF_COUNT_HW_INSTRUCTIONS;
+	attr.exclude_kernel=1;
 
 	ev_fd = sys_perf_event_open( &attr, 0, -1, -1, 0 );
 	if ( ev_fd == -1 ) {
@@ -296,6 +297,7 @@ check_exclude_guest( void )
 	/* Now try again with excude_guest */
 	memset(&attr, 0 , sizeof(attr));
 	attr.config = PERF_COUNT_HW_INSTRUCTIONS;
+	attr.exclude_kernel=1;
 	attr.exclude_guest=1;
 
 	ev_fd = sys_perf_event_open( &attr, 0, -1, -1, 0 );


Treece-Burgess:
Hello @willowec, what error are you getting when you do not have `exclude_kernel=1` set and when does it occur?

willowec:
@Treece-Burgess I apologize for excluding that information previously. Without the fix introduced in this PR, the error occurs for nearly every test invoked by `make fulltest`. It does not appear to be affecting the tests, but I figure it probably shouldn't be happening. For example, here is some trimmed output from `make fulltest`:

```
...
Running ctests/calibrate:                                  PAPI Error: Couldn't open hw_instructions in exclude_guest=0 test
SKIPPED
Running ctests/case1:                                      PAPI Error: Couldn't open hw_instructions in exclude_guest=0 test
PASSED
Running ctests/case2:                                      PAPI Error: Couldn't open hw_instructions in exclude_guest=0 test
PASSED
Running ctests/child_overflow:                             PAPI Error: Couldn't open hw_instructions in exclude_guest=0 test
SKIPPED
...
```
This occurs on my machine running Debian 6.7.12-1~bpo12+1, kernel 6.7.12+bpo-amd64 with `perf_event_paranoid`=2, but not for `perf_event_paranoid`<2.

Treece-Burgess:
@willowec I tested the proposed changes on a machine where my paranoid level is set to 4 and I still get the output of `Couldn't open hw_instructions in exclude_guest=0`:
```
Running ctests/attach2:                                    PAPI Error: Couldn't open hw_instructions in exclude_guest=0 test
SKIPPED
Running ctests/attach3:                                    PAPI Error: Couldn't open hw_instructions in exclude_guest=0 test
SKIPPED
Running ctests/attach_cpu:                                 PAPI Error: Couldn't open hw_instructions in exclude_guest=0 test
SKIPPED
```
I am on an Ubuntu machine with a kernel version of 5.15.0-87-generic.

willowec:
Interesting! The same thing happens for me on my Debian machine with `perf_event_paranoid=4`. Looking into it some more, it looks like the master branch of the linux kernel only defines paranoia levels up to 2 (see [kernel/events/core.c](https://github.com/torvalds/linux/blob/8faabc041a001140564f718dabe37753e88b37fa/kernel/events/core.c#L446)). The comment documenting the levels is mirrored here:
```
/*
 * perf event paranoia level:
 *  -1 - not paranoid at all
 *   0 - disallow raw tracepoint access for unpriv
 *   1 - disallow cpu events for unpriv
 *   2 - disallow kernel profiling for unpriv
 */
```
It also looks like some distros like Debian and Ubuntu include patches that add an extra level that disallows all use of `perf_event_open()` for unprivileged users. For Debian, that patch is [security,perf: Allow further restriction of perf_event_open](https://sources.debian.org/src/linux/6.10.11-1/debian/patches/features/all/security-perf-allow-further-restriction-of-perf_event_open.patch/).

Since paranoia level 2 only disallows kernel profiling for unprivileged users, from my understanding shouldn't `check_exclude_guest()` be able to execute without error when `perf_event_paranoid=2`? Either way, editing the error message to indicate the error is probably a paranoia issue would be helpful I think.

Treece-Burgess:
@willowec On a system where my `perf_event_paranoid` level is set to 2, I do get the output of `PAPI Error: Couldn't open hw_instructions in exclude_guest=0 test` when I run `./run_tests.sh`. Even though the tests do seem to run as expected. 
With your proposed changes this does not occur. 

I think adding your suggested change and updating the error message would probably be the best course of action. Let me speak with another team member about it more.

adanalis (approved):

Treece-Burgess:
@willowec Could you rebase your branch and then I will go ahead with merging.

willowec:
@Treece-Burgess Sorry for the delay, re-based!

PR #259, opened 2024-10-11
counter_analysis_toolkit: Automatically detect platform in Makefile

willowec:
## Pull Request Description

When compiling the Counter Analysis Toolkit (CAT) if the wrong architecture is selected for the ARCH variable, relatively confusing compilation errors can occur. For example, compiling on an ARM machine with the default `ARCH=X86` setting produces the following GCC error: `gcc: error: unrecognized command-line option ‘-mfma’`.

To prevent this from happening, make the CAT Makefile use 'uname -m' to detect if the architecture is arm or x86, and otherwise assume powerpc. If ARCH is already set, leave it alone. This change streamlines the workflow of using the CAT on new devices.

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/counter_analysis_toolkit/Makefile
+++ b/src/counter_analysis_toolkit/Makefile
@@ -16,7 +16,7 @@ ifeq ($(USEMPI),true)
 endif
 
 ## Try to auto-detect architecture if it is not set.
-ifeq ($(origin ARCH), undefined)
+ifndef ARCH
     UNAME_M:=$(shell uname -m)
 
     ARCH=POWER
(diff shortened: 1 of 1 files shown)


Treece-Burgess:
I think this would be a good addition. Personally I think using `ifndef ARCH` is better than `ifeq ($(origin ARCH), undefined)`. Mainly due to how it is defined in the [GNU documentation](https://www.gnu.org/software/make/manual/html_node/Origin-Function.html). 

I have tested on an x86 machine and it did work properly. @dbarry9 @adanalis You both have worked much more on this than myself, what do you two think?

willowec:
Thank you for the feedback! `ifndef ARCH` is definitely more readable as well. I've gone ahead and updated that.

adanalis (approved):

willowec:
Done!

PR #243, opened 2024-09-19
Update Makefile - CUDA 12.2

ametere-ssi:
Adding these two extra options to compiler and linker will produce successful compilation for my machine with gcc 9.4.0 and CUDA 12.2. Compilation fails otherwise.

## Author Checklist
- [1] Alfredo Metere, Samsung Semiconductor, Inc.
This fix addresses compilation failure of CUDA module tests on GCC v.9.4.0 and CUDA 12.2.
- changed two lines in the Makefile of CUDA module test.
The PR needs to pass all the tests

PR #262, opened 2024-10-15
Updating Cuda Tests Makefile with -pthread

Treece-Burgess:
## Pull Request Description
In GLIBC version 2.34, libpthread as a separate library has been removed. Most components that used to be in a separate shared object have been placed into the main libc object, `libc.so.6`. Due to this, the `-lpthread` flag is not needed in GLIBC >= 2.34. See the following [webpage ](https://developers.redhat.com/articles/2021/12/17/why-glibc-234-removed-libpthread#) for more discussion on this topic. 

Even with versions of GLIBC >= 2.34 or GLIBC < 2.34, we still should place the `-pthread` flag when needed. `-pthread` encompasses `-lpthread`, but it also defines extra macros. From the man pages for gcc the following is stated for the `-pthread` flag: You should use this option consistently for both compilation and linking.

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**


Treece-Burgess:
@ametere-ssi I was able to finally recreate the `pthread_create` error when building PAPI with the Cuda component you first mentioned in PR #243. Please let me know if this PR resolves this bug.

adanalis (approved):

PR #265, opened 2024-10-17
Sysdetect: allow users to disable component.

adanalis:
Added a flag (--with-sysdetect=<yes,no>) in configure that allows users to disable the Sysdetect component. The default value is "yes", so if the user does not specify anything, the component is included.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**


Treece-Burgess (approved):
Looks good to me, I tested the following:

- `--with-sysdetect=no`, sysdetect component did not show in `papi_component_avail`
- `--with-sysdetect=yes`, sysdetect component showed in `papi_component_avail`
- No specification of `--with-sysdetect`, sysdetect component showed in `papi_component_avail`

Along with making sure that the output for configure matched the value to `--with-sysdetect`.

PR #269, opened 2024-10-23
sysdetect: accounting for missing core ids.

adanalis:
## Pull Request Description

Some systems have non-contiguous numbering of cores, which affects the way we count the number of cores each numa node has.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/sysdetect/cpu.c
+++ b/src/components/sysdetect/cpu.c
@@ -119,8 +119,10 @@ fill_cpu_info( _sysdetect_cpu_info_t *info )
     }
 
     for (a = 0; a < info->threads * info->cores * info->sockets; ++a) {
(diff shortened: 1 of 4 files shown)


src/components/sysdetect/cpu.c
```diff
@@ -119,8 +119,10 @@ fill_cpu_info( _sysdetect_cpu_info_t *info )
     }
 
     for (a = 0; a < info->threads * info->cores * info->sockets; ++a) {
-        CPU_CALL(cpu_get_attribute_at(CPU_ATTR__HWTHREAD_NUMA_AFFINITY, a, &info->numa_affinity[a]),
-                 info->numa_affinity[a] = 0);
```
gcongiu, line 123: Instead of setting numa_affinity[a]=0 in case of error we could have set it to -1 here?

gcongiu (approved):
I haven't tested it but it looks good

PR #278, opened 2024-11-11
Update all_native_events.c and get_event_component.c to allow users to optionally disable Cuda native events

Treece-Burgess:
## Pull Request Description
This PR looks to add an optional flag to `all_native_events.c` and `get_event_component.c` titled `--disable-cuda-events` with options of `yes` and `no` (the default is no). The reason for this is due to the runtime length of `all_native_events.c` and `get_event_component.c`when the Cuda component has been compiled in.

Along with the `--disable-cuda-events` option, I have also added `-h` and `--help`. Example of `--help` and `-h` is below (shown is `all_native_events.c`, but similarly is `get_event_component.c`):
```
This is the all_native_events program.
For all components compiled in, it attempts to: add the component events, start, and stop.
Usage: ./all_native_events [options]
General command options:
	-h, --help                           Print the help message.
	--disable-cuda-events=<yes,no>       Optionally disable processing the Cuda native events. Default is no.

```

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/ctests/all_native_events.c
+++ b/src/ctests/all_native_events.c
@@ -1,6 +1,7 @@
 /*
  * File:    all_native_events.c
  * Author:  Haihang You <you@cs.utk.edu>
+ * Author:  Treece Burgess tburgess@icl.utk.edu (updated in November 2024 to add a flag to enable or disable Cuda events.)
  */
 
 /* This test tries to add all native events from all components */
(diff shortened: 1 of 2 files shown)


src/ctests/all_native_events.c
```diff
@@ -22,169 +23,199 @@
...
+    int numcmp, cid;
+    int quiet;
+
+    /* Set quiet variable */
+    quiet=tests_quiet( argc, argv );
+
+    /* parse command line flags */
+    for (i = 0; i < argc; i++) {
+        if (strncmp(argv[i], "--disable-cuda-events=", 22) == 0) {
+            strcpy(disableCudaEvts, argv[i] + 22);
```
adanalis, line 107: It would be better to use: strcpy(disableCudaEvts, argv[i] + 22, PAPI_MIN_STR_LEN);

Treece-Burgess: I have updated `all_native_events.c` to use `strncpy`.

adanalis (approved):

PR #319, opened 2025-02-12
rocm_smi: pci_bandwidth_get return not supported

djwoun:
## Pull Request Description
Draft to test CI
rocm_smi: pci_bandwidth_get return not supported
Simply disabling the part of the code. 

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/components/rocm_smi/rocs.c
+++ b/src/components/rocm_smi/rocs.c
@@ -1377,7 +1377,7 @@ handle_derived_events_count(const char *v_name, int32_t dev, int64_t v_variant, 
             (*events_count) += ROCS_PCI_BW_VARIANT__CURRENT + 1;
         }
         int i;
-        for (i = 0; i < ROCS_PCI_BW_VARIANT__LANE_IDX - ROCS_PCI_BW_VARIANT__CURRENT + 1; ++i) {
+        for (i = 0; i < ROCS_PCI_BW_VARIANT__LANE_IDX - ROCS_PCI_BW_VARIANT__CURRENT; ++i) {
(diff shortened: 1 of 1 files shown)


src/components/rocm_smi/rocs.c
```diff
@@ -1022,10 +1022,10 @@ init_device_table(void)
 
     for (i = 0; i < device_count; ++i) {
         status = rsmi_dev_pci_bandwidth_get_p(i, &pcie_table[i]);
-        if (status != RSMI_STATUS_SUCCESS && status != RSMI_STATUS_NOT_YET_IMPLEMENTED) {
-            papi_errno = PAPI_EMISC;
-            goto fn_fail;
-        }
+        //if (status != RSMI_STATUS_SUCCESS && status != RSMI_STATUS_NOT_YET_IMPLEMENTED) {
```
adanalis, line 1025: If we don't expect this function to work, why do we have it in the code?

djwoun: This is the second bug of rocm_smi that fails during run time.
The function is expected to work with the MI210 on dopamine(returning RSMI_STATUS_SUCCESS). It returns "RSMI_STATUS_NOT_SUPPORTED" with the MI300(Odyssey) & MI250(Frontier).

I was not sure what the best solution would be, but I thought potentially we could simply check for RSMI_STATUS_SUCCESS && RSMI_STATUS_SUCCESS && RSMI_STATUS_NOT_SUPPORTED.

I wanted to check the CI, so I put a draft pr to run a test.

PR #328, opened 2025-03-12
Adjust NVIDIA library names

scaronni:
## Pull Request Description

The NVIDIA driver packages have historically been shipping symlinks for some unversioned libraries in various subpackages. This was a mistake, most of the libraries could not be used for linking and as a consequence they should have not been packaged at all.

Some of the these libraries, had even duplicate unversioned library symlink in the CUDA repository along with appropriate headers in a different package. For example `libnvidia-ml.so` is provided by the `cuda-nvml-devel` package along with the headers, while the driver provides only the runtime part `libnvidia-ml.so.1`.

In general unversioned shared libraries should not be used for runtime loding, only for linking, so change the code to always load the versioned library for `libcuda`, `libvndia-ml` and `libcudart`.

Sample output for the RPM (deb is similar):

```
$ rpm -qpl cuda-nvml-devel-12-6-12.6.77-1.x86_64.rpm | grep targets
/usr/local/cuda-12.6/targets
/usr/local/cuda-12.6/targets/x86_64-linux
/usr/local/cuda-12.6/targets/x86_64-linux/include
/usr/local/cuda-12.6/targets/x86_64-linux/include/nvml.h
/usr/local/cuda-12.6/targets/x86_64-linux/lib
/usr/local/cuda-12.6/targets/x86_64-linux/lib/stubs
/usr/local/cuda-12.6/targets/x86_64-linux/lib/stubs/libnvidia-ml.a
/usr/local/cuda-12.6/targets/x86_64-linux/lib/stubs/libnvidia-ml.so
```

```
$ rpm -qpvl cuda-cudart-devel-12-8-12.8.90-1.x86_64.rpm | grep libcudart.so
lrwxrwxrwx    1 root     root                       15 Jul  8  2015 /usr/local/cuda-12.8/targets/x86_64-linux/lib/libcudart.so -> libcudart.so.12
```

Some references:
- https://github.com/NVIDIA/yum-packaging-nvidia-driver/issues/9
- https://github.com/NVIDIA/go-nvml/blob/0e815c71ca6e8184387d8b502b2ef2d2722165b9/pkg/nvml/lib.go#L30
- https://github.com/rust-nvml/nvml-wrapper/pull/63
- https://pypi.org/project/nvidia-ml-py/ - official python library from NVIDIA (`nvmlLib = CDLL("libnvidia-ml.so.1")`

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/components/cuda/papi_cupti_common.c
+++ b/src/components/cuda/papi_cupti_common.c
@@ -54,13 +54,13 @@ cudaError_t ( *cudaRuntimeGetVersionPtr ) (int *);
 CUptiResult ( *cuptiGetVersionPtr ) (uint32_t* );
 
 /**@class load_cuda_sym
- * @brief Search for libcuda.so.
+ * @brief Search for libcuda.so.1.
  */
 static int load_cuda_sym(void)
 {
-    dl_drv = dlopen("libcuda.so", RTLD_NOW | RTLD_GLOBAL);
+    dl_drv = dlopen("libcuda.so.1", RTLD_NOW | RTLD_GLOBAL);
     if (!dl_drv) {
-        ERRDBG("Loading installed libcuda.so failed. Check that cuda drivers are installed.\n");
+        ERRDBG("Loading installed libcuda.so.1 failed. Check that cuda drivers are installed.\n");
         goto fn_fail;
     }
 
--- a/src/components/cuda/papi_cupti_common.c
+++ b/src/components/cuda/papi_cupti_common.c
@@ -141,24 +141,24 @@ void *cuptic_load_dynamic_syms(const char *parent_path, const char *dlname, cons
 }
 
 /**@class load_cudart_sym
- * @brief Search for libcudart.so. Order of search is outlined below.
+ * @brief Search for libcudart.so.12. Order of search is outlined below.
  *
  * 1. If a user sets PAPI_CUDA_RUNTIME, this will take precedent over
  *    the options listed below to be searched.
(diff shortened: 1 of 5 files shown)


Treece-Burgess:
Changing `libcuda.so` and `libnvidia-ml.so` to `libcuda.so.1` and `libnvidia-ml.so.1` I think is applicable. However, I think following slurms [implementation](https://github.com/SchedMD/slurm/commit/12cea079f4b5077e6e7d35b72e99b51a562bc8e2) is a better safety net. 

The issue I see with hard coding `libcudart.so` to `libcudart.so.12` is that the # changes between major toolkit versions and not all PAPI users may end up using a Cuda toolkit version >= 12.0. For example:

- Cuda toolkit 11.3.1
   - `libcudart.so.11.0`
- Cuda toolkit 12.0
   - `libcudart.so.12`

The best option I see for this, is to search for the versioned `libcudart.so`. @jagode @adanalis What do you two think?

Treece-Burgess:
Thank you for opening up the pull request. Hard coding the shared objects to use specific version numbers (e.g. `libcudart.so.12`) would not be applicable for PAPI as users most likely will be using different toolkit versions. I am working on a PR that will search the directory for the shared objects instead.

Treece-Burgess:
Thank you for opening up the pull request. Hard coding the shared objects to use specific version numbers (e.g. `libcudart.so.12`) would not be applicable for PAPI as users most likely will be using different toolkit versions. I am working on a PR that will search the directory for the shared objects instead and will be closing this PR.

PR #335, opened 2025-03-20
ROCP_SDK: Change tests/Makefile for spack builds.

adanalis:
## Pull Request Description

Instead of explicitly setting the path for amdclang++ in the tests/Makefile, assuming that the compiler will be in the rocprofiler directory structure, allow it to be picked up from the environment PATH, or explicitly set through the new variable AMDCXX

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/rocp_sdk/tests/Makefile
+++ b/src/components/rocp_sdk/tests/Makefile
@@ -1,6 +1,7 @@
 NAME=template
 include ../../Makefile_comp_tests.target
 
+AMDCXX   ?= amdclang++
 CFLAGS    = $(OPTFLAGS)
 CPPFLAGS += $(INCLUDE)
 LDFLAGS  += $(PAPILIB) $(TESTLIB) $(UTILOBJS)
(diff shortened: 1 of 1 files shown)


src/components/rocp_sdk/tests/Makefile
```diff
@@ -1,6 +1,7 @@
 NAME=template
 include ../../Makefile_comp_tests.target
 
+AMDCXX    = amdclang++
```
G-Ragghianti, line 4: To allow the user to set this variable to override the default, I think you need:
```suggestion
AMDCXX    := amdclang++
```

G-Ragghianti (changes requested):

src/components/rocp_sdk/tests/Makefile
adanalis: You are right. I fixed it.

G-Ragghianti (approved):

PR #358, opened 2025-05-05
Cuda component: Replace int typing with long long to avoid overflow with measured values

Treece-Burgess:
## Pull Request Description
This PR replaces the `int` datatype of `metricValues` in `cuptip_ctx_read` with `long long` such that we avoid overflow with measured values.

As stands with the current master branch if you run the `concurrent_profiling` test with cuda:::sm__cycles_active:stat=sum:device=0 you will end up with a negative value: 
```
Metrics for device #0:
Look at the sm__cycles_elapsed.max values for each test.
This value represents the time spent on device to run the kernels in each case, and should be longest for the serial range, and roughly equal for the single and multi device concurrent ranges.
PAPI event name							Measured value
cuda:::sm__cycles_active:stat=sum:device=0			-2147483648
PASSED
```

This PR resolves this issue:
```
Metrics for device #0:
Look at the sm__cycles_elapsed.max values for each test.
This value represents the time spent on device to run the kernels in each case, and should be longest for the serial range, and roughly equal for the single and multi device concurrent ranges.
PAPI event name							Measured value
cuda:::sm__cycles_active:stat=sum:device=0			2160446356
PASSED
```

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/components/cuda/cupti_profiler.c
+++ b/src/components/cuda/cupti_profiler.c
@@ -178,7 +178,7 @@ static int get_metric_eval_request(NVPW_MetricsEvaluator *metricEvaluator, const
 static int create_raw_metric_requests(NVPW_MetricsEvaluator *pMetricsEvaluator, NVPW_MetricEvalRequest *metricEvalRequest, NVPA_RawMetricRequest **rawMetricRequests, int *rawMetricRequestsCount);
 // Metric Evaluation
 static int get_number_of_passes_for_eventsets(const char *pChipName, const char *metricName, int *numOfPasses);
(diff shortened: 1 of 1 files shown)


dbarry9:
I am testing this pull request.

dbarry9:
I tested these changes on a system containing the NVIDIA Blackwell architecture (GeForce RTX 5080) with CUDA 12.8. The utilities are functional, and all component tests pass. The `concurrent_profiling` and `concurrent_profiling_noCuCtx` component tests no longer show negative counter values.

dbarry9 (approved):

PR #377, opened 2025-05-16
New search procedure for libhsa library

G-Ragghianti:
## Pull Request Description

In the rocp_sdk component, the search path for the library libhsa-runtime64.so should look under PAPI_ROCM_ROOT instead of PAPI_ROCP_SDK_ROOT because the library is provided by ROCM base package. To facility this for users, we will search under PAPI_ROCP_SDK_ROOT, then under PAPI_ROCM_ROOT, and then we will allow the OS to try to find the library for us.

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/components/rocp_sdk/rocp_sdk.c
+++ b/src/components/rocp_sdk/rocp_sdk.c
@@ -22,7 +22,7 @@
 #include "papi_memory.h"
 #include "extras.h"
 #include "sdk_class.h"
-#include "hsa.h"
+#include "rocprofiler-sdk/hsa.h"
 
 #define ROCPROF_SDK_MAX_COUNTERS (64)
 #define RPSDK_CTX_RUNNING (1)
@@ -488,19 +488,32 @@ hsa_is_enabled( void )
              (hsa_agent_get_infoPtr != NULL) );
 }
 
-static int
-load_hsa_sym( char *err_msg )
+void *dlopen_from_paths( char *libname, int pathCount, const char *paths[] )
 {
-    char pathname[PATH_MAX] = "libhsa-runtime64.so";
-    char *rocm_root = getenv("PAPI_ROCP_SDK_ROOT");
-    if (rocm_root != NULL) {
-        int count = snprintf(pathname, PATH_MAX, "%s/lib/libhsa-runtime64.so", rocm_root);
+    void *so = NULL;
+    for (int i = 0; i < pathCount; i++) {
+        if ( paths[i] == NULL ) {
+            continue;
+        }
+        char newpath[PATH_MAX];
+        int count = snprintf(newpath, PATH_MAX, "%s/lib/%s", paths[i], libname);
         if (count >= PATH_MAX) {
             SUBDBG("Status string truncated.");
         }
+        so = dlopen(newpath, RTLD_NOW | RTLD_GLOBAL);
+        if (so) {
+            return so;
+        }
     }
+    so = dlopen(libname, RTLD_NOW | RTLD_GLOBAL);
+    return so;
+}
 
-    rocm_dlp = dlopen(pathname, RTLD_NOW | RTLD_GLOBAL);
+static int
+load_hsa_sym( char *err_msg )
+{
+    const char *paths[] = {getenv("PAPI_ROCP_SDK_ROOT"), getenv("PAPI_ROCM_ROOT")};
+    rocm_dlp = dlopen_from_paths("libhsa-runtime64.so", 2, paths);
     if (rocm_dlp == NULL) {
         int count = snprintf(err_msg, PAPI_MAX_STR_LEN, "%s", dlerror());
         if (count >= PAPI_MAX_STR_LEN) {


src/components/rocp_sdk/rocp_sdk.c
```diff
@@ -492,7 +492,7 @@ static int
 load_hsa_sym( char *err_msg )
 {
     char pathname[PATH_MAX] = "libhsa-runtime64.so";
-    char *rocm_root = getenv("PAPI_ROCP_SDK_ROOT");
+    char *rocm_root = getenv("PAPI_ROCM_ROOT");
```
adanalis, line 495: I don't want to require all users to set both variables, since in most systems it will be the same path. How about we check for PAPI_ROCM_ROOT first, and if we don't find it then we check for PAPI_ROCP_SDK_ROOT and if both fail, we fall back to the dlopen() without a prefix.

G-Ragghianti:
Ok I'll do that.

G-Ragghianti:
The new search method is implemented in a new function `dlopen_from_paths` which will search any number of paths that you give it and fall back to allowing the OS to try to find the library.  This function should probably be included somewhere else so that other components can use it, but for now I think it would be better to only implement it here.  If it turns out to work well, we can put it somewhere else.

G-Ragghianti:
Tested on the ICL cluster:

```bash
./configure --with-components=rocp_sdk
export PAPI_ROCP_SDK_ROOT=/opt/rocm/ make
utils/papi_component_avail
Available components and hardware information.
PAPI version             : 7.2.0.0
Operating system         : Linux 6.1.129-1.el9.elrepo.x86_64
Vendor string and code   : AuthenticAMD (2, 0x2)
Model string and code    : AMD EPYC 7302 16-Core Processor (49, 0x31)
CPU revision             : 0.000000
CPUID                    : Family/Model/Stepping 23/49/0, 0x17/0x31/0x00
CPU Max MHz              : 3000
CPU Min MHz              : 1500
Total cores              : 32
SMT threads per core     : 2
Cores per socket         : 16
Sockets                  : 1
Cores per NUMA region    : 32
NUMA regions             : 1
Running in a VM          : no
Number Hardware Counters : 5
Max Multiplex Counters   : 384
Fast counter read (rdpmc): yes

Compiled-in components:
Name:   perf_event              Linux perf_event CPU counters
Name:   perf_event_uncore       Linux perf_event CPU uncore and northbridge
Name:   rocp_sdk                GPU events and metrics via AMD ROCprofiler-SDK
Name:   sysdetect               System info detection component

Active components:
Name:   perf_event              Linux perf_event CPU counters
                                Native: 141, Preset: 17, Counters: 5
                                PMUs supported: perf, perf_raw, amd64_fam17h_zen2

Name:   perf_event_uncore       Linux perf_event CPU uncore and northbridge
                                Native: 1, Preset: 0, Counters: 3
                                PMUs supported: amd64_rapl

Name:   rocp_sdk                GPU events and metrics via AMD ROCprofiler-SDK
                                Native: 144, Preset: 0, Counters: 144

Name:   sysdetect               System info detection component
                                Native: 0, Preset: 0, Counters: 0


```

G-Ragghianti:
With these changes, and updates to the papi package in spack, I have all the components compiled and active: https://github.com/icl-utk-edu/iSDK/actions/runs/15099000285

G-Ragghianti:
It looks like papi_internal.c would be an appropriate location for `dlopen_from_paths()`.  I can move it there if that is desired.

src/components/rocp_sdk/rocp_sdk.c
```diff
@@ -488,19 +488,32 @@ hsa_is_enabled( void )
...
     }
+    so = dlopen(libname, RTLD_NOW | RTLD_GLOBAL);
+    return so;
+}
 
-    rocm_dlp = dlopen(pathname, RTLD_NOW | RTLD_GLOBAL);
+static int
+load_hsa_sym( char *err_msg )
+{
+    const char *paths[] = {getenv("PAPI_ROCP_SDK_ROOT"), getenv("PAPI_ROCM_ROOT")};
```
adanalis, line 515: Add a call to dlerror() here to clear previous errors.

src/components/rocp_sdk/rocp_sdk.c
```diff
@@ -488,19 +488,32 @@ hsa_is_enabled( void )
...
+        }
+        char newpath[PATH_MAX];
+        int count = snprintf(newpath, PATH_MAX, "%s/lib/%s", paths[i], libname);
         if (count >= PATH_MAX) {
             SUBDBG("Status string truncated.");
         }
+        so = dlopen(newpath, RTLD_NOW | RTLD_GLOBAL);
+        if (so) {
+            return so;
+        }
```
adanalis, line 506: Since you check for the path being empty, if the execution reached this point, then the path was wrong. We should error, instead of silently ignoring the bad path, using something like this:
 
else{
    return NULL;
}

adanalis:
> It looks like papi_internal.c would be an appropriate location for `dlopen_from_paths()`. I can move it there if that is desired.

There is a tug-of-war between the design principles of factoring code across components and keeping components stand-alone and self-contained in their respective directories. So far, we've always taken the stand-alone/self-contained path, so let's not add component related code to papi_internal.c. Maybe we should have a new file, component_common.c or something like this for such code.

src/components/rocp_sdk/rocp_sdk.c
adanalis: That's true. I was going by Steve Kaufmann's request to fail when the user has provided a bad path, but this doesn't apply to the HSA library because we are not sure a priori where it will be. Long story short, your code is good the way it is.

adanalis (approved):

adanalis:
I tested this PR on a system with MI300A GPUs and a system with MI210 GPUs.

src/components/rocp_sdk/rocp_sdk.c
```diff
@@ -22,7 +22,7 @@
 #include "papi_memory.h"
 #include "extras.h"
 #include "sdk_class.h"
-#include "hsa.h"
+#include "rocprofiler-sdk/hsa.h"
```
adanalis, line 25: It turns out that this change is causing some conflicts when building both rocp_sdk and rocm_smi at the same time. Do not merge yet. I'm looking into it.

G-Ragghianti: I tested both together, and I got events.  https://github.com/icl-utk-edu/iSDK/actions/runs/15099000285/job/42438708390

What problem are you seeing?

PR #381, opened 2025-05-20
ROCM_SMI: Added -pthread flag in tests/Makefile.

adanalis:
## Pull Request Description

The Makefile in the tests directory of the rocm_smi component overwrites the LDFLAGS variable, so it won't inherit the "-pthread" flag even when necessary.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**


tokey-tahmid (approved):
Tested on MI200. The build output is correct and PAPI utilities are functional.

PR #398, opened 2025-06-04
RAPL Component: Add support in RAPL for Intel Emerald Rapids

Treece-Burgess:
## Pull Request Description
This PR adds RAPL support for Intel Emerald Rapids CPUs.  For the RAPL case, I verified the model number (207) via a [post](https://www.redhat.com/en/blog/red-hat-enterprise-linux-performance-results-5th-gen-intel-xeon-scalable-processors) by Red Hat that shows the hardware configuration.

**Note at this time the PAPI team does not have access to a machine with an Intel Emerald Rapids CPU to verify this addition**.

As a sanity check, I configured and built PAPI as follows `./configure --prefix=$PWD/test-install --with-components="rapl"` on an Intel Xeon Gold 6430 (Sapphire Rapids - Family/Model/Stepping 6/143/8). The build was successful, with the utilities `papi_avail`, `papi_component_avail`, `papi_native_avail`, and `papi_command_line` working as expected (`rapl` was disabled as I lacked the necessary permissions).

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/components/rapl/linux-rapl.c
+++ b/src/components/rapl/linux-rapl.c
@@ -479,6 +479,7 @@ _rapl_init_component( int cidx )
 			break;
 			
 		case 143:      /* Sapphire Rapids-SP */
+		case 207:      /* Emerald Rapids */
 		        package_avail=1;
 			pp0_avail=0;
 			pp1_avail=0;


src/papi_events.csv
```diff
@@ -1032,7 +1032,8 @@ PRESET,PAPI_VEC_SP,DERIVED_POSTFIX,N0|N1|N2|N3|+|+|+|,FP_ARITH:SCALAR_SINGLE,FP_
 PRESET,PAPI_VEC_INS,DERIVED_POSTFIX,N0|N1|N2|N3|N4|N5|+|+|+|+|+|,FP_ARITH_INST_RETIRED:128B_PACKED_SINGLE,FP_ARITH_INST_RETIRED:256B_PACKED_SINGLE,FP_ARITH_INST_RETIRED:512B_PACKED_SINGLE,FP_ARITH_INST_RETIRED:128B_PACKED_DOUBLE,FP_ARITH_INST_RETIRED:256B_PACKED_DOUBLE,FP_ARITH_INST_RETIRED:512B_PACKED_DOUBLE
 # End of icx, icl list
 
-# Intel Sapphire Rapids events
+# Intel Sapphire Rapids events and Emerald Rapids Events
+CPU,emr
```
jagode: Let's not include PAPI preset support in this commit, which should focus on RAPL support only. 

RAPL support is one thing, but PAPI CPU presets should be added later, once we get a chance to run the CAT benchmarks and verify the definitions. There are too many unknowns and potential issues with untested presets.  

Please remove the line "CPU,emr" from this PR.

jagode (changes requested):
Let's not include PAPI preset support in this commit, which should focus on RAPL support only.

RAPL support is one thing, but PAPI CPU presets should be added later, once we get a chance to run the CAT benchmarks and verify the definitions. There are too many unknowns and potential issues with untested presets.

Please remove the line "CPU,emr" from this PR.

src/papi_events.csv
Treece-Burgess: Got it, I have made the necessary changes in the PR to reflect this.

PR #403, opened 2025-06-06
ROCM: PAPI_strerror() cannot be used at shutdown.

adanalis:
## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/rocm/rocm.c
+++ b/src/components/rocm/rocm.c
@@ -261,7 +261,7 @@ rocm_shutdown_component(void)
     }
 
   fn_exit:
-    SUBDBG("EXIT: %s\n", PAPI_strerror(papi_errno));
+    SUBDBG("EXIT\n");
     return papi_errno;
   fn_fail:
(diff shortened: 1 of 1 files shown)


src/components/rocm/rocm.c
```diff
@@ -261,7 +261,7 @@ rocm_shutdown_component(void)
     }
 
   fn_exit:
-    SUBDBG("EXIT: %s\n", PAPI_strerror(papi_errno));
+    SUBDBG("EXIT\n");
```
Treece-Burgess, line 264: This is a suggestion and if you don't think it is necessary then go ahead and merge it, but you could add the error code with the SUBDBG e.g. `SUBDBG("EXIT: %d\n", papi_errno)` or `SUBDBG("EXIT: result: %d\n", papi_errno)`.

 `rocd_shutdown` from what I saw only returns `PAPI_OK`, but maybe in the future another error code is added.

Treece-Burgess (approved):

Issue #385, opened 2025-05-21
Incorrect Behavior Of PAPI_EDELAY_INIT Components


Treece-Burgess:
Before PR #284 was merged, if you went to retrieve a `PAPI_EDELAY_INIT` components info (e.g. `rocm_smi`) you would be met with the following output:
```
Name: rocm_smi
Short name: rocm_smi
Description: AMD GPU System Management Interface via rocm_smi_lib
Version: 2.0
Support Version: 
Kernal Version: 
Disabled reason: Not initialized. Access component events to initialize it.
Disabled: -26
Initialized: 0
CmpIdx: 3
Num Cntrs: -1
Num mpx cntrs: 0
Num preset events: 0
Num native events: -1
```
With the key takeaways being we would need to access a `rocm_smi` event to initialize it and the number of native events is currently set to -1 until we intialize it. 

With the addition of PR #284 you will now see:
```
Name: rocm_smi
Short name: rocm_smi
Description: AMD GPU System Management Interface via rocm_smi_lib
Version: 2.0
Support Version: 
Kernal Version: 
Disabled reason: Not initialized. Access component events to initialize it.
Disabled: 0
Initialized: 1
CmpIdx: 3
Num Cntrs: 261
Num mpx cntrs: 261
Num preset events: 0
Num native events: 261
```
Which shows we are initializing the `rocm_smi` component instead of the user initializing the component. This is due to line [2230](https://github.com/icl-utk-edu/papi/blob/6beafaf47a4c9187db54ca1cb3fea946847cf2fa/src/papi_internal.c#L2230) in `papi_internal.c` (Thanks to Daniel for the help on tracking this down).

PR #410, opened 2025-06-10
framework: force init per existing policy

dbarry9:
## Pull Request Description

PR #284 introduced code that always forced the initialization of all components. However, this defeats the purpose of having PAPI_EDELAY_INIT.

The changes in this pull request only force initialization of components when necessary.

These changes have been tested on systems containing:
- NVIDIA Hopper architecture
- AMD Zen3 CPU and AMD MI250X GPU architectures (Frontier).

This pull request addresses Issue #385.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/cuda/linux-cuda.c
+++ b/src/components/cuda/linux-cuda.c
@@ -118,7 +118,6 @@ papi_vector_t _cuda_vector = {
 
     .init_thread = cuda_init_thread,
     .shutdown_thread = cuda_shutdown_thread,
-    .init_comp_presets = cuda_init_comp_presets,
 
     .ntv_enum_events = cuda_ntv_enum_events,
     .ntv_code_to_name = cuda_ntv_code_to_name,
@@ -234,7 +233,18 @@ static int cuda_init_private(void)
 static int check_n_initialize(void)
 {
     if (!_cuda_vector.cmp_info.initialized) {
-        return cuda_init_private();
+        int papi_errno = cuda_init_private();
+        if( PAPI_OK != papi_errno ) {
+            return papi_errno;
+        }
+
+        // Setup the presets.
+        papi_errno = cuda_init_comp_presets();
+        if( PAPI_OK != papi_errno ) {
+            return papi_errno;
+        }
+
+        return papi_errno;
     }
     return _cuda_vector.cmp_info.disabled;
 }
--- a/src/papi.c
+++ b/src/papi.c
@@ -588,7 +588,7 @@ extern hwi_presets_t user_defined_events[PAPI_MAX_USER_EVENTS];
 extern int user_defined_events_count;
 extern int num_all_presets;
 extern int _papi_hwi_start_idx[PAPI_NUM_COMP];
-extern int first_comp_idx;
+extern int first_comp_with_presets;
 extern int first_comp_preset_idx;
 
 
@@ -1504,6 +1504,11 @@ PAPI_event_code_to_name( int EventCode, char *out )
         if( compIdx < 0 ) {
             return PAPI_ENOEVNT;
         }
+        if ( _papi_hwd[compIdx]->cmp_info.disabled == PAPI_EDELAY_INIT ) {
+            int junk;
+            _papi_hwd[compIdx]->ntv_enum_events(&junk, PAPI_ENUM_FIRST);
+        }
+
 
 		if (_papi_hwi_comp_presets[compIdx][preset_index].symbol == NULL )
 			papi_return( PAPI_ENOTPRESET );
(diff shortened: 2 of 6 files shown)


Treece-Burgess:
I am currently reviewing this PR.

src/components/cuda/linux-cuda.c
```diff
@@ -233,9 +232,19 @@ static int cuda_init_private(void)
 
 static int check_n_initialize(void)
 {
+    int status;
     if (!_cuda_vector.cmp_info.initialized) {
-        return cuda_init_private();
+        status = cuda_init_private();
```
Treece-Burgess: Would `cuda_init_comp_presets` not rely on `cuda_init_private` returning `PAPI_OK`? 

It appears to me that it would. Due to that I would check the return value of `cuda_init_private` immediately after instead of returning it at the end. 

E.g:
```
int papi_errno = cuda_init_private();
if (papi_errno  != PAPI_OK) {
    return papi_errno;
}
.
.
.

return papi_errno;
```

src/papi.c
```diff
@@ -1788,8 +1800,17 @@ PAPI_enum_event( int *EventCode, int modifier )
 
     if ( IS_PRESET(i) ) {
 
+        if ( _papi_hwd[cidx]->cmp_info.disabled == PAPI_EDELAY_INIT ) {
```
Treece-Burgess: You initialize the `PAPI_EDELAY_INIT` component here, but you do it again inside `if (modifier == PAPI_ENUM_FIRST)`. Are both times actually needed?

Also you use `cidx` in one case and `pe_idx` in another case in terms of indexing. Isn't `pe_idx` for `perf_event` which would not be `PAPI_EDELAY_INIT`?

src/papi.c
```diff
@@ -1833,6 +1854,12 @@ PAPI_enum_event( int *EventCode, int modifier )
             if( preset_index < 0 ) {
                 return ( PAPI_ENOEVNT );
             }
+
+            if ( _papi_hwd[first_comp_with_presets]->cmp_info.disabled == PAPI_EDELAY_INIT ) {
```
Treece-Burgess, line 1853: Same question as above.

src/papi_internal.c
```diff
@@ -2220,25 +2216,14 @@ _papi_hwi_init_global_presets( void )
         is_pe = 0;
         if (strcmp(_papi_hwd[i]->cmp_info.name, "perf_event") == 0) {
             is_pe = 1;
+            pe_idx = i;
         } else {
-            /* Only set the first non-perf_event component index once. */
-            if ( -1 == first_comp_idx ) {
-                first_comp_idx++;
+            /* Only set the first non-perf_event component with presets once. */
```
Treece-Burgess, line 2219: Just so I understand, why is this done?

src/components/cuda/linux-cuda.c
dbarry9: Yes, you are correct. It would rely on the return value being `PAPI_OK`. I will update the code per your suggestion.

src/papi_internal.c
dbarry9: Previously, this tracked the first non-perf_event component. However, it is more useful to track the first non-perf_event component _that contains presets_ because this information is needed when `PAPI_enum_event()` is called with the modifier `PAPI_PRESET_ENUM_FIRST_COMP`.

src/papi.c
dbarry9: > You initialize the PAPI_EDELAY_INIT component here, but you do it again inside if (modifier == PAPI_ENUM_FIRST). Are both times actually needed?

No, both times are not needed I will address this in my next push.

> Also you use cidx in one case and pe_idx in another case in terms of indexing. Isn't pe_idx for perf_event which would not be PAPI_EDELAY_INIT?

Yes, that is correct. I believe it is unnecessary to add this.

src/papi.c
dbarry9: To add on to my other comment, I think it would be best if I put this only in the necessary modifiers' cases.

src/papi_internal.c
Treece-Burgess: Thank you for clarifying as that makes sense to me now.

src/components/cuda/linux-cuda.c
Treece-Burgess: The updated code looks good to me!

Treece-Burgess (approved):
Did final tests and re-looked over the code. Everything looks good to me.

dbarry9:
Please do not merge yet. I need to perform other tests.

PR #418, opened 2025-06-13
rocm_smi: Expose energy and power consumption measurements

adam-mcdaniel:
## Pull Request Description

Currently, PAPI does not expose the `rsmi_dev_energy_count_get` and `rsmi_dev_current_socket_power_get` functions.
These functions are exceedingly important for HPC researchers to measure the energy costs of different workloads across AMD machines.

This PR works in conjunction with the [2 line PR to ROCm/rocm_smi_lib which adds these counters to the device function map](https://github.com/ROCm/rocm_smi_lib/pull/222).

## Author Checklist
- [x] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/rocm_smi/rocs.c
+++ b/src/components/rocm_smi/rocs.c
@@ -2498,9 +2498,9 @@ get_event_descr(const char *name, int64_t variant, int64_t subvariant)
                 return NULL;
         }
     } else if (strcmp(name, "rsmi_dev_energy_count_get") == 0) {
-        return papi_strdup("Accumulated GPU energy, in microjoules (µJ).");
+        return papi_strdup("Accumulated amount of GPU energy consumed in microjoules (uJ).");
     } else if (strcmp(name, "rsmi_dev_current_socket_power_get") == 0) {
-        return papi_strdup("Current power consumption in microwatts (µW)");
+        return papi_strdup("Current socket power (also known as instant power) in microwatts (uW).");
     } else if (strcmp(name, "rsmi_dev_temp_metric_get") == 0) {
         switch (variant) {
             case RSMI_TEMP_CURRENT:
@@ -4007,30 +4007,31 @@ access_rsmi_dev_energy_count(rocs_access_mode_e mode, void *arg)
     }
 
     rsmi_status_t status;
-    uint64_t raw_count = 0ULL;
+    uint64_t energy_counter = 0ULL;
     float    resolution = 0.0f;   /* in microjoules */
     uint64_t timestamp = 0ULL;    /* in ns, if you care */
 
     status = rsmi_dev_energy_count_get_p((uint32_t)event->device,
-                                         &raw_count,
+                                         &energy_counter,
                                          &resolution,
                                          &timestamp);
-    if (status != RSMI_STATUS_SUCCESS &&  status != RSMI_STATUS_INVALID_ARGS) {
-        /*
-         * Note: RSMI_STATUS_INVALID_ARGS can mean "power was null" if the device
-         * actually supports the call. If it’s truly unsupported, typically
-         * RSMI_STATUS_NOT_SUPPORTED is returned. Adjust logic if needed.
-         */
+    if (status != RSMI_STATUS_SUCCESS) {
+        if (status == RSMI_STATUS_INVALID_ARGS) {
+            SUBDBG("Energy counter is a nullptr. However, the function is supported with the provided arguments.\n");
+        }
+        else if (status == RSMI_STATUS_NOT_SUPPORTED) {
+            SUBDBG("Energy counter is a nullptr. However, the function is not supported with the installed software or hardware with the provided arguments.\n");
+        }
+
         return PAPI_EMISC;
     }
 
     /*
-     * According to the doc, 'raw_count' * 'resolution' (in µJ) is the total
-     * energy consumed since the counter began accumulating. We store as
-     * an int64_t in microjoules. Watch for overflow if the accumulated
-     * energy is very large.
+     * According to the doc (e.g. python_smi_tools/rocm_smi.py), 'energy_counter' * 'resolution' (in uJ)
+     * is the total energy consumed since the counter began accumulating. We store as an int64_t in
+     * microjoules. Watch for overflow if the accumulated energy is very large.
      */
-    double total_uJ = (double)raw_count * (double)resolution;
+    double total_uJ = (double)energy_counter * (double)resolution;
     if (total_uJ > INT64_MAX) {
         /* clamp to avoid overflow */
         total_uJ = INT64_MAX;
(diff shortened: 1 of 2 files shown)


adam-mcdaniel:
@Treece-Burgess it's now been merged into the `amd-mainline` branch! Let me know what else I can do to get the changes merged here as well!

Treece-Burgess:
@adam-mcdaniel Thank you for having that taken care of, I will try to get around to reviewing this PR next week!

src/components/rocm_smi/rocs.c
```diff
@@ -3975,3 +3995,74 @@ access_rsmi_dev_vendor_name(rocs_access_mode_e mode, void *arg)
...
+        return PAPI_ENOSUPP; /* not supported */
+    }
+
+    rsmi_status_t status;
+    uint64_t power = 0ULL; /* in µW */
+
+    status = rsmi_dev_current_socket_power_get_p((uint32_t)event->device, &power);
+    if (status != RSMI_STATUS_SUCCESS && status != RSMI_STATUS_INVALID_ARGS) {
+        /*
+         * Note: RSMI_STATUS_INVALID_ARGS can mean "power was null" if the device
```
Treece-Burgess: What exactly does "power was null" mean (likewise with energy counter being null)?

- Does this mean the value is technically 0? I see in the [documentation](https://rocm.docs.amd.com/projects/rocm_smi_lib/en/docs-6.4.0/doxygen/html/group__PowerQuer.html#gafb3d36d17698b411c320982cc8f8da82) your comment is correct. But is there more explanation on this?

src/components/rocm_smi/rocs.c
```diff
@@ -3975,3 +3995,74 @@ access_rsmi_dev_vendor_name(rocs_access_mode_e mode, void *arg)
...
+
+    /* This event is read-only. If a caller tries to WRITE, deny it. */
+    if (mode != ROCS_ACCESS_MODE__READ || mode != event->mode) {
+        return PAPI_ENOSUPP; /* not supported */
+    }
+
+    rsmi_status_t status;
+    uint64_t power = 0ULL; /* in µW */
+
+    status = rsmi_dev_current_socket_power_get_p((uint32_t)event->device, &power);
```
Treece-Burgess: On a machine with 2 * AMD MI210s, I run into the error `RSMI_STATUS_NOT_SUPPORTED`, I saw that in documentation it is noted that this function is supported by newer devices. I was able to test on an MI300A and there I am able to monitor the `current_socket_power`. 

What devices have you tested this on? As I do not see a list from AMD that explicitly states which devices fall under this "newer" category.

src/components/rocm_smi/rocs.c
```diff
@@ -2481,6 +2497,10 @@ get_event_descr(const char *name, int64_t variant, int64_t subvariant)
             default:
                 return NULL;
         }
+    } else if (strcmp(name, "rsmi_dev_energy_count_get") == 0) {
+        return papi_strdup("Accumulated GPU energy, in microjoules (µJ).");
+    } else if (strcmp(name, "rsmi_dev_current_socket_power_get") == 0) {
+        return papi_strdup("Current power consumption in microwatts (µW)");
```
Treece-Burgess: Can you change the definition to `Current socket power (also known as instant power) in microwatts (uW).`?

src/components/rocm_smi/rocs.c
```diff
@@ -2481,6 +2497,10 @@ get_event_descr(const char *name, int64_t variant, int64_t subvariant)
             default:
                 return NULL;
         }
+    } else if (strcmp(name, "rsmi_dev_energy_count_get") == 0) {
+        return papi_strdup("Accumulated GPU energy, in microjoules (µJ).");
```
Treece-Burgess: Can you change the definition to `Accumulated amount of GPU energy consumed in microjoules (uJ).`?

src/components/rocm_smi/rocs.c
```diff
@@ -3975,3 +3995,74 @@ access_rsmi_dev_vendor_name(rocs_access_mode_e mode, void *arg)
...
+    float    resolution = 0.0f;   /* in microjoules */
+    uint64_t timestamp = 0ULL;    /* in ns, if you care */
+
+    status = rsmi_dev_energy_count_get_p((uint32_t)event->device,
+                                         &raw_count,
+                                         &resolution,
+                                         &timestamp);
+    if (status != RSMI_STATUS_SUCCESS &&  status != RSMI_STATUS_INVALID_ARGS) {
+        /*
+         * Note: RSMI_STATUS_INVALID_ARGS can mean "power was null" if the device
```
Treece-Burgess: Comment states "power was null", but this is for the `energy_count` event. Please change to `energy counter is null`.

src/components/rocm_smi/rocs.c
```diff
@@ -3975,3 +3995,74 @@ access_rsmi_dev_vendor_name(rocs_access_mode_e mode, void *arg)
...
+        return PAPI_EMISC;
+    }
+
+    /*
+     * According to the doc, 'raw_count' * 'resolution' (in µJ) is the total
+     * energy consumed since the counter began accumulating. We store as
+     * an int64_t in microjoules. Watch for overflow if the accumulated
+     * energy is very large.
+     */
+    double total_uJ = (double)raw_count * (double)resolution;
```
Treece-Burgess: I see in their example function `TestMetricsCounterRead::Run` in the [source file](https://github.com/ROCm/rocm_smi_lib/blob/7dba992ebd1beef78ec78d0d2974d6aa531786ff/tests/rocm_smi_test/functional/metrics_counter_read.cc#L124)`metrics_counter_read.cc` that they do multiple `power` (your `raw_count`) * `counter_resolution` (your `resolution`). 

Can you link the documentation in reply to this comment, where it states `raw_count` * `resolution` is the total energy consumed...?

src/components/rocm_smi/rocs.c
Treece-Burgess: From reading through the documentation again, if `socket_power`  is set to a `nullptr` then the function call failed. Either by the return value of `RSMI_STATUS_NOT_SUPPORTED` or `RSMI_STATUS_NOT_SUPPORTED`.  This holds true as well for if `power` is set to a `nullptr`.

See [documentation](https://rocm.docs.amd.com/projects/rocm_smi_lib/en/docs-6.4.0/doxygen/html/group__PowerQuer.html#gafb3d36d17698b411c320982cc8f8da82).

src/components/rocm_smi/rocs.c
Treece-Burgess: Upon further looking into this I did not find specifically documentation that confirms that `power` * `counter_resolution` is accumulated energy in micro-joules. However, in the function titled `showEnergy` located in the file `python_smi_tools/rocm_smi.py` it does state that "Accumulated Energy" is equal to `power` * `counter_resolution`:
```
printLog(device, "Accumulated Energy (uJ)", round(power.value * counter_resolution.value, 2))
```

src/components/rocm_smi/rocs.c
Treece-Burgess: Below are the AMD GPUs that the two new events successfully run on (this is from what I have available to me to test on currently):

`energy_count`:
- AMD MI300A
- AMD MI210



`current_socket_power`:
- MI300A

src/components/rocm_smi/rocs.c
adam-mcdaniel: Ahh I see -- I'll change the comment to be correct.

src/components/rocm_smi/rocs.c
adam-mcdaniel: I've successfully tested this on the MI300A and the MI250X on Frontier here at ORNL. The `current_socket_power` is only supported by newer GPUs -- I've only been able to test that aspect on the MI300A myself as well.

src/components/rocm_smi/rocs.c
Treece-Burgess: @adam-mcdaniel No need to add the comments. I have made the change to have `SUBDBG` messages for each case. It serves as documentation, but also can allow users to know what is occurring. 

I am going to make one more push to have `#if defined` wrapped around the cases for `rsmi_dev_current_socket_power_get`. As this was only introduced in ROCm 6.0.0.

src/components/rocm_smi/rocs.c
Treece-Burgess: Got it! Thank you for the reply!

adanalis (approved):

Treece-Burgess (approved):

Issue #421, opened 2025-06-16
rocm component test segfault


dbarry9:
When running the `rocm` component test `intercept_multi_thread_monitoring` on Frontier (AMD MI250X; ROCm 6.2.0, 6.2.4, 6.3.1, 6.4.0), I encounter the following:

> ./intercept_multi_thread_monitoring : multi GPU activity monitoring program.
> [tid:140701867206336] rocm:::SQ_INSTS_VALU:device=0 : 0
> [tid:140701867206336] rocm:::SQ_INSTS_SALU:device=0 : 0
> [tid:140701867206336] rocm:::SQ_WAVES:device=0 : 0
> [tid:140701867206336] rocm:::SQ_WAVES_RESTORED:device=0 : 0
> FAILED!!!
> Line # 125 Error in match_expected_counter: Invalid argument
> Some tests require special hardware, permissions, OS, compilers
> or library versions. PAPI may still function perfectly on your
> system without the particular feature being tested here.
> Segmentation fault (core dumped)

PR #422, opened 2025-06-17
rocm: fix segmentation fault in component test

dbarry9:
## Pull Request Description

On Frontier, the invocation of the `exit()` call before `pthread_merge()` causes a segmentation fault. I remedy this issue by only calling `test_warn()`, `test_fail()`, and `hip_test_fail()` after the threads have been merged.

These changes were tested using ROCm versions 6.1.3, 6.2.0, 6.2.4, 6.3.1, and 6.4.0 with the AMD MI250X architecture on the Frontier supercomputer.

This pull request addresses Issue #421.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/rocm/tests/multi_thread_monitoring.cpp
+++ b/src/components/rocm/tests/multi_thread_monitoring.cpp
@@ -15,13 +15,34 @@
 #include <pthread.h>
 #include "common.h"
 
+#define PASS     0x0
+#define PASSWARN 0x1
+#define FAIL     0x2
+#define HIPFAIL  0x4
+
+int *testLINE = NULL;
+int *status = NULL;
+char **errMSG = NULL;
+int *papi_errno = NULL;
+hipError_t *hip_errno = NULL;
+
 int quiet;
+
+static void log_error(int testLINE_arg, int status_arg, const char *errMSG_arg, int tid) {
+
+    testLINE[tid] = testLINE_arg;
+    status[tid] = status_arg;
+
+    int ret = snprintf(errMSG[tid], PAPI_MAX_STR_LEN, "%s", errMSG_arg);
+    if ( ret < 0 || ret >= PAPI_MAX_STR_LEN ) {
+        fprintf(stdout, "[%s, %d] WARNING: Could not copy string %s into buffer.\n", __FILE__, __LINE__, errMSG_arg);
+    }
+
+    return;
+}
 
 static void *run(void *thread_num_arg)
 {
-    int papi_errno;
-    int pass_with_warning = 0;
-    hipError_t hip_errno;
     int j;
 
 #define NUM_EVENTS 4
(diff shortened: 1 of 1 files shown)


Treece-Burgess:
I am reviewing this PR.

src/components/rocm/tests/multi_thread_monitoring.cpp
```diff
@@ -35,59 +57,81 @@ static void *run(void *thread_num_arg)
...
+            status[thread_num] = PASSWARN;
         }
     }
 
-    papi_errno = PAPI_start(eventset);
-    if (papi_errno != PAPI_OK) {
-        test_fail(__FILE__, __LINE__, "PAPI_start", papi_errno);
+    papi_errno[thread_num] = PAPI_start(eventset);
+    if (papi_errno[thread_num] != PAPI_OK) {
+        localErr = strdup("PAPI_start");
```
Treece-Burgess: Instead of having the `localErr` variable and having to call `strdup` which you should `free`. I would just pass the local error string directly to `log_error`. From my testing that worked.

src/components/rocm/tests/multi_thread_monitoring.cpp
```diff
@@ -35,59 +57,81 @@ static void *run(void *thread_num_arg)
...
 
     for (int j = 0; j < NUM_EVENTS; ++j) {
         char named_event[PAPI_MAX_STR_LEN];
         sprintf(named_event, "%s:device=%d", events[j], thread_num);
-        papi_errno = PAPI_add_named_event(eventset, (const char*) named_event);
-        if (papi_errno != PAPI_OK && papi_errno != PAPI_ENOEVNT) {
-            test_fail(__FILE__, __LINE__, "PAPI_add_named_event", papi_errno);
-        } else if (papi_errno == PAPI_ENOEVNT) {
-            pass_with_warning = 1;
+        papi_errno[thread_num] = PAPI_add_named_event(eventset, (const char*) named_event);
```
Treece-Burgess, line 74: Indexing wise `NUM_EVENTS` and `thread_num` may not match. Due to that are you just concerned with the last error message on that thread for `PAPI_add_named_event`?

src/components/rocm/tests/multi_thread_monitoring.cpp
```diff
@@ -35,59 +57,81 @@ static void *run(void *thread_num_arg)
...
+        if (papi_errno[thread_num] != PAPI_OK && papi_errno[thread_num] != PAPI_ENOEVNT) {
+            localErr = strdup("PAPI_add_named_event");
+            log_error(__LINE__, FAIL, localErr, thread_num);
+            pthread_exit(NULL);
+        } else if (papi_errno[thread_num] == PAPI_ENOEVNT) {
```
Treece-Burgess, line 78: Just a side comment, this is an interesting check from the current master branch code as depending on the device we could call `PAPI_start` on an empty eventset.

src/components/rocm/tests/multi_thread_monitoring.cpp
```diff
@@ -184,6 +271,39 @@ int multi_thread(int argc, char *argv[])
         pthread_join(thread[i], NULL);
     }
 
+    int status_main = PASS;
+    int tid;
```
Treece-Burgess, line 262: This comment is sort of nit-picky so if you want to, you can leave what you have.

`tid` normally stands for thread id or atleast that is what I normally see. From what I understand in your code, `tid` really means the device index (correct me if I am wrong). You may want to change `tid` to `devId` to be more clear.

src/components/rocm/tests/multi_thread_monitoring.cpp
```diff
@@ -184,6 +271,39 @@ int multi_thread(int argc, char *argv[])
...
+    for ( tid = 0; tid < dev_count; ++tid) {
+        if (PASS != status[tid]) {
+            status_main = status[tid];
+            break;
+        }
+    }
+
+    switch ( status_main ) {
```
Treece-Burgess, line 270: Just wanted to say that this `switch` statement and the `log_error` functions were great ideas.

src/components/rocm/tests/multi_thread_monitoring.cpp
dbarry9: I agree and have updated the pull request accordingly.

src/components/rocm/tests/multi_thread_monitoring.cpp
dbarry9: It does mean the device index here; however, in these tests, the number of threads spawned is equal to the number of devices.

src/components/rocm/tests/multi_thread_monitoring.cpp
dbarry9: I think the concern here should be if an error occurs, so in that case the last error (any case where `papi_errno[thread_num] != PAPI_OK`) should suffice.

src/components/rocm/tests/multi_thread_monitoring.cpp
Treece-Burgess: That is fair!

PR #449, opened 2025-07-16
counter_analysis_toolkit: add CMake support

dbarry9:
## Pull Request Description

To accommodate different instruction sets and precisions available in modern architectures, these changes add CMake as a more flexible build system for the Counter Analysis Toolkit.

These changes have been tested on the Intel Sapphire Rapids architecture.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**


src/counter_analysis_toolkit/README
Treece-Burgess, line 1: I would make this a markdown file (extension `.md`). You can then break sections like `CMake Compilation (Recommended)` by a header (#, ##, ###, etc.) and then add a code block so it is easier for a user to copy and paste code to a script.

## CMake Compilation (Recommended)
```
mkdir build && cd build
cmake -Dpapi_dir=/path/to/your/papi/installation ..
```

## Legacy Compilation:
```
make PAPIDIR=/path/to/your/papi/installation
```

src/counter_analysis_toolkit/README
```diff
@@ -2,9 +2,12 @@ Description:
 Benchmarks for helping in the understanding of native events
 by stressing different aspects of the architecture selectively.
 
-Compilation:
-make PAPIDIR=/path/to/your/papi/installation
+CMake Compilation (Recommended):
+mkdir build && cd build
+cmake -Dpapi_dir=/path/to/your/papi/installation ..
```
Treece-Burgess: For `cmake` you use `papi_dir` and for `make` you use `PAPIDIR`, a couple questions from that:

- Why not make them the exact same var name?
- Why not just use `PAPI_DIR` as shown in Step 4 of [Downloading and Installing PAPI](https://github.com/icl-utk-edu/papi/wiki/Downloading-and-Installing-PAPI#downloading-and-installing-papi)?
    - I think more users are more likely to set that by default and you could just check that in the CMakeLists.txt file.

src/counter_analysis_toolkit/README
```diff
@@ -2,9 +2,12 @@ Description:
...
 
+Legacy Compilation:
```
Treece-Burgess: Per our Slack discussion, for the Legacy Compilation the directions need to add `cd src`. You may want to distinguish it from the PAPI framework `src` as well. So `counter_analysis_toolkit/src`.

src/counter_analysis_toolkit/src/replicate.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
Treece-Burgess, line 1: This file is nearly identical to the removed `src/counter_analysis_toolkit/replicate.sh`. Essentially it was renamed to the current path location and added the additional argument check. 

For tracking purposes can you make it such that it shows that `src/counter_analysis_toolkit/replicate.sh` was renamed (->) to `src/counter_analysis_toolkit/src/replicate.sh`?

src/counter_analysis_toolkit/src/replicate.sh
```diff
@@ -0,0 +1,10 @@
...
+
+if [ $# -lt 1 ]
+then
+    echo -e "replicate.sh: Must supply output dir for generated source!"
+    exit
+fi
+for ((i=1; i<12; i++)); do
+    cp $1/icache_seq_kernel_0.so $1/icache_seq_kernel_${i}.so;
```
Treece-Burgess: The semicolon is not needed here. Primarily they are used for command separators in bash.

src/counter_analysis_toolkit/src/replicate.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+if [ $# -lt 1 ]
+then
+    echo -e "replicate.sh: Must supply output dir for generated source!"
```
Treece-Burgess: Why do you need `echo -e`? From my understanding that is used to interpret backslash characters. Also from what I could find it is not POSIX compliant.

src/counter_analysis_toolkit/src/replicate.sh
```diff
@@ -0,0 +1,10 @@
...
+    exit
```
Treece-Burgess: Just using `exit` will return the value of 0 as the previous `echo` command will run successfully. 

`exit 3` seems to indicate that there is an invalid or missing argument. However, this is not standardized. So you could use `exit 1` just as a generic error.

src/counter_analysis_toolkit/src/gen_seq_dlopen.sh
```diff
@@ -1,8 +1,13 @@
 #!/bin/bash
 
-DRV_F=icache_seq.c
-KRN_F=icache_seq_kernel.c
-HEAD_F=icache_seq.h
+if [ $# -lt 1 ]
+then
+    echo -e "gen_seq_dlopen.sh: Must supply output dir for generated source!"
```
Treece-Burgess: Same as below.

Treece-Burgess: Why is an output dir needed now, but originally it was not needed?

src/counter_analysis_toolkit/CMakeLists.txt
```diff
@@ -0,0 +1,494 @@
...
+option( color "Use ANSI color output" true )
+
+#-------------------------------------------------------------------------------
+# Default compiler flags.
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -Wall -Wextra")
+
+#-------------------------------------------------------------------------------
+# PAPI -- required.
+message( "" )
+message( "---------------------------------------- PAPI: papi_dir = ${papi_dir}" )
```
Treece-Burgess: This is personal preference, but instead of having `---------------------------------------- PAPI: papi_dir = ${papi_dir}"`, I would do sections and put the section title in the middle, so:
```
----------- PAPI -----------
papi_dir = ${papi_dir}
```

ROCm SMI does the following: 
```
&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&
                    CMake ROCm SMI (Library) [root]                
&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&
```

I think it overall reads and looks better, but if you prefer the other way definitely fine too.

src/counter_analysis_toolkit/CMakeLists.txt
```diff
@@ -0,0 +1,494 @@
...
+if (NOT papi_dir)
```
Treece-Burgess: From my testing you have to run `cmake -Dpapi_dir=$PWD/test-install`. Why not also provide the option for a user to `export papi_dir`? This goes with my comment about using `PAPI_DIR` as well.

So:
```
if (NOT DEFINED papi_dir AND NOT DEFINED ENV{papi_dir})
    message(FATAL_ERROR "PAPI is required" )
else()
    target_link_directories(cat_collect PRIVATE ${papi_dir}/lib)
    target_link_libraries(cat_collect PRIVATE papi)
    include_directories(${papi_dir}/include)
endif()
```

src/counter_analysis_toolkit/CMakeLists.txt
```diff
@@ -0,0 +1,494 @@
...
+    CounterAnalysisToolkit
+    VERSION 2025.07.14
+    LANGUAGES C
+)
+
+include(CheckCCompilerFlag)
+
+#-------------------------------------------------------------------------------
+# Detect CPU architecture.
+if(NOT DEFINED ARCH)
```
Treece-Burgess: I tested setting ARCH as follows: `cmake -Dpapi_dir=$PWD/test-install -DARCH=test ..`

Doing this will see me still make it to the end of the entire cmake process:
```
[ICL:methane build]$ cmake -Dpapi_dir=/home/tburgess/daniel_fork/papi/src/test-install -DARCH=test ..

---------------------------------------- PAPI: papi_dir = /home/tburgess/daniel_fork/papi/src/test-install

---------------------------------------- OpenMP

---------------------------------------- MPI: use_mpi = OFF
-- User has requested to NOT use MPI
-- Configuring done (0.0s)
-- Generating done (0.0s)
-- Build files have been written to: /home/tburgess/daniel_fork/papi/src/counter_analysis_toolkit/build
```

I do not think this should be the case, as current support is for POWER, ARM, and x86.

Something else that could be done as well is have this as the `----- Architecture -----` section. I think the workflow could be as follows:
1. Check to see if `ARCH` has been set, if it has not been set then we look to detect it for them.
2. Check to see if `ARCH` has been set, if it has been we make sure it is an architecture we support. If not we error.
    - If we do support it, we then detect the CPU architecture ourselves internally to make sure the ARCH provided matches what we detect internally. If they do not match, then I would show a warning and not error.

src/counter_analysis_toolkit/CMakeLists.txt
```diff
@@ -0,0 +1,494 @@
...
+
+        add_library(vec_nonfma_hp OBJECT src/vec_nonfma_hp.c)
+        target_compile_options(vec_nonfma_hp PRIVATE ${VEC})
+
+        add_library(vec_nonfma_sp OBJECT src/vec_nonfma_sp.c)
+        target_compile_options(vec_nonfma_sp PRIVATE ${VEC})
+
+        add_library(vec_nonfma_dp OBJECT src/vec_nonfma_dp.c)
+        target_compile_options(vec_nonfma_dp PRIVATE ${VEC})
+    else()
```
Treece-Burgess: If the arch is neither `aarch64` or `x86` it seems that you default to `POWER`. Would it not be better to also use `elseif` for power e.g. `elseif(UNAME_M STREQUAL "ppc64le")` and then leave the `else` statement to handle an error if the arch is not supported?

src/counter_analysis_toolkit/README
```diff
@@ -2,9 +2,12 @@ Description:
...
+make PAPIDIR=/path/to/your/papi/installation
 
 Usage:
 ./cat_collect -in event_list.txt -out OUTPUT_DIRECTORY -branch -dcr
```
Treece-Burgess: Not sure where to put this, but there is no default `event_list.txt` after you run `make` for the CMake workflow. Currently with CAT, you do have a default `event_list.txt`. I am not sure if you wanted to keep that, but wanted to point it out.

Treece-Burgess: As a future suggestion, if a user provides a `-out` and it does not already exist. Could you not create it on the fly for the user?

src/counter_analysis_toolkit/src/Makefile
```diff
@@ -66,7 +66,7 @@ all: branch.o d_cache eventstock.o flops i_cache instr vector
 
 d_cache: timing_kernels.o prepareArray.o compar.o dcache.o
 
-i_cache: icache.o icache_seq_kernel_0.o
+i_cache: icache.o icache_seq_kernel.o
```
Treece-Burgess, line 69: What was the significance of the zero originally in the name? As it is now discarded.

src/counter_analysis_toolkit/src/Makefile
```diff
@@ -92,15 +92,15 @@ flops: flops.c flops.h cat_arch.h
...
+icache_seq_kernel.o: icache_seq.c icache_seq.h icache_seq_kernel.c
 	$(CC) $(CFLAGS) $(OPT0) $(INCFLAGS) -c icache_seq.c -o icache_seq.o
-	$(CC) $(CFLAGS) $(OPT0) $(INCFLAGS) -fPIC -c icache_seq_kernel.c -o icache_seq_kernel_0.o
-	$(CC) $(CFLAGS) $(OPT0) -shared -o icache_seq_kernel_0.so icache_seq_kernel_0.o
-	bash replicate.sh
-	rm icache_seq_kernel_0.o
+	$(CC) $(CFLAGS) $(OPT0) $(INCFLAGS) -fPIC -c icache_seq_kernel.c -o icache_seq_kernel.o
+	$(CC) $(CFLAGS) $(OPT0) -shared -o icache_seq_kernel_0.so icache_seq_kernel.o
+	bash replicate.sh ${PWD}
+	rm ${PWD}/icache_seq_kernel.o
```
Treece-Burgess, line 103: Is `${PWD}` really needed? As `make` will be ran in `counter_analysis_toolkit/src`which is where `icache_seq_kernel.o` will be created.

src/counter_analysis_toolkit/src/Makefile
```diff
@@ -92,15 +92,15 @@ flops: flops.c flops.h cat_arch.h
 	$(CC) $(CFLAGS) $(FLOP) $(OPT1) $(INCFLAGS) -c flops.c -o flops.o
 
 icache.o: icache.c icache.h
-	bash gen_seq_dlopen.sh
+	bash gen_seq_dlopen.sh ${PWD}
```
Treece-Burgess, line 95: The output dir for `gen_seq_dlopen.sh` will just be the current PATH everytime?

src/counter_analysis_toolkit/src/Makefile
```diff
@@ -92,15 +92,15 @@ flops: flops.c flops.h cat_arch.h
...
 	$(CC) $(CFLAGS) $(OPT0) $(INCFLAGS) -c icache.c -o icache.o
 
-icache_seq_kernel_0.o: icache_seq.c icache_seq.h
+icache_seq_kernel.o: icache_seq.c icache_seq.h icache_seq_kernel.c
```
Treece-Burgess, line 98: On Methane at ICL, running `make PAPIDIR=/home/tburgess/daniel_fork/papi/src/test-install`, I end up with the following during compilation:
```
icache_seq_kernel.c: In function ‘seq_kernel_TRUE_IF_10’:
icache_seq_kernel.c:69:5: error: expected declaration or statement at end of input
   69 |     is_zero *= result;
      |     ^~~~~~~
icache_seq_kernel.c:16:15: warning: variable ‘cntr_value’ set but not used [-Wunused-but-set-variable]
   16 |     long long cntr_value;
      |               ^~~~~~~~~~
icache_seq_kernel.c:12:9: warning: unused variable ‘jj’ [-Wunused-variable]
   12 |     int jj, is_zero;
      |         ^~
icache_seq_kernel.c:11:37: warning: unused parameter ‘epilogue’ [-Wunused-parameter]
   11 | long long seq_kernel_TRUE_IF_10(int epilogue){
      |                                 ~~~~^~~~~~~~
icache_seq_kernel.c:70: warning: control reaches end of non-void function [-Wreturn-type]
   70 | 
      | 
make: *** [Makefile:100: icache_seq_kernel.o] Error 1
make: *** Waiting for unfinished jobs....

```

src/counter_analysis_toolkit/src/Makefile
```diff
@@ -92,15 +92,15 @@ flops: flops.c flops.h cat_arch.h
...
 	$(CC) $(CFLAGS) $(OPT0) $(INCFLAGS) -c icache_seq.c -o icache_seq.o
-	$(CC) $(CFLAGS) $(OPT0) $(INCFLAGS) -fPIC -c icache_seq_kernel.c -o icache_seq_kernel_0.o
-	$(CC) $(CFLAGS) $(OPT0) -shared -o icache_seq_kernel_0.so icache_seq_kernel_0.o
-	bash replicate.sh
-	rm icache_seq_kernel_0.o
+	$(CC) $(CFLAGS) $(OPT0) $(INCFLAGS) -fPIC -c icache_seq_kernel.c -o icache_seq_kernel.o
+	$(CC) $(CFLAGS) $(OPT0) -shared -o icache_seq_kernel_0.so icache_seq_kernel.o
```
Treece-Burgess, line 101: You removed 0 from the `icache_seq_kernel.o` filenames, do you need to remove it from `icache_seq_kernel_0.so`?

src/counter_analysis_toolkit/src/Makefile
Treece-Burgess, line 98: Why have you now added `icache_seq_kernel.c`?

Treece-Burgess (changes requested):

src/counter_analysis_toolkit/src/Makefile
dbarry9: If the legacy Makefile is being used to build CAT, then yes. This is okay because I only pass `${PWD}` explicitly in the legacy Makefile.

src/counter_analysis_toolkit/README
dbarry9: I have updated the README to use markdown, per your suggestion.

Treece-Burgess:
> @Treece-Burgess I am writing this message just to document that we will need to update the CI tests to account for the new directory structure introduced in this PR.

Yes, once you are done making changes. I think I will update the CI through this PR as well. I will default to use the CMake option. Is that fine with you?

src/counter_analysis_toolkit/src/Makefile
dbarry9: `${PWD}` is not needed, per se; however, `replicate.sh` expects an argument to be given (which was necessary for the CMake build process).

src/counter_analysis_toolkit/src/Makefile
dbarry9: No, we do not need to remove it from `cache_seq_kernel_0.so` (and we do not want to remove it because there are other `icache_seq_kernel_X.so` files that `replicate.sh` produces); it was just extraneous for the object `icache_seq_kernel.o`.

src/counter_analysis_toolkit/src/Makefile
dbarry9: I have added it here to be consistent with the CMake build process, which lists it as a source file for the object `icache_seq_kernel`.

dbarry9:
> > @Treece-Burgess I am writing this message just to document that we will need to update the CI tests to account for the new directory structure introduced in this PR.
> 
> Yes, once you are done making changes. I think I will update the CI through this PR as well. I will default to use the CMake option. Is that fine with you?

Yes, that is good.

src/counter_analysis_toolkit/src/gen_seq_dlopen.sh
dbarry9: This is due CMake not placing object code (generated at compile time) into the source code directory, so the scripts (which live in the source directory) need to know where the generated output will live.

src/counter_analysis_toolkit/src/replicate.sh
dbarry9: I will look into this.

src/counter_analysis_toolkit/src/Makefile
dbarry9: It was a design decision, but it was not necessary.

Treece-Burgess: Got it. Thank you for the clarification!

src/counter_analysis_toolkit/README
dbarry9: Done.

src/counter_analysis_toolkit/README
dbarry9: I have changed them all to `PAPI_DIR` because as you stated, that is what the documentation uses by default.

src/counter_analysis_toolkit/src/replicate.sh
dbarry9: The issue seems to be that git does not know how to find the similarity between the old file and what appears to it as a new file (since such a large percentage of the lines have changed). I am not sure how to remedy this.

src/counter_analysis_toolkit/CMakeLists.txt
dbarry9: I agree that it would look nicer to change it. I decided to go with your first suggested option for now:
```
----------- PAPI -----------
papi_dir = ${papi_dir}
```

src/counter_analysis_toolkit/README
dbarry9: This is a good idea. I will create an issue for this soon.

dbarry9: Thank you for catching this! I think it would be good to include in the binary directory, so I have added it.

src/counter_analysis_toolkit/src/replicate.sh
dbarry9: I have updated it to use printf.

src/counter_analysis_toolkit/CMakeLists.txt
dbarry9: This is more robust. I have updated CMakeLists.txt accordingly.

src/counter_analysis_toolkit/CMakeLists.txt
dbarry9: I agree that the ARCH detection should be more robust. Since we detect the architecture in both of the cases you have described, maybe we should not even allow the user to specify the architecture (as that adds extra unnecessary burden to specify an architecture that has to match what we can and will automatically detect).

src/counter_analysis_toolkit/src/Makefile
dbarry9: I was not able to reproduce this with GCC 12.4.0, which is what I use on this machine. A couple questions so that I can reproduce this:
- Are you using the default compiler?
- Are you using CMake?

src/counter_analysis_toolkit/src/replicate.sh
Treece-Burgess: Sounds good! It is not a huge deal just wanted to see if it was possible. Thank you for looking into it!

src/counter_analysis_toolkit/README.md
```diff
@@ -0,0 +1,44 @@
...
+ICACHE:MISSES 0
+ICACHE:HIT 0
+OFFCORE_RESPONSE_0:DMND_DATA_RD:L3_HIT:SNP_ANY 0
+```
+
+# Build Instructions
+## CMake Compilation (Recommended):
+```
+mkdir build && cd build
+cmake -DPAPI_DIR=/path/to/your/papi/installation ..
```
Treece-Burgess, line 32: I think I forgot to add this on my last review, but I would add `make` as the last command in this list of commands to run.

```
mkdir build && cd build
cmake -DPAPI_DIR=/path/to/your/papi/installation ..
make
```

src/counter_analysis_toolkit/src/Makefile
dbarry9: I have created Issue #463 to track this, since it is a separate issue from the CMake additions.

src/counter_analysis_toolkit/CMakeLists.txt
dbarry9: I have added a warning that will get printed to the user if they try to provide a `-DARCH` argument to CMake.

src/counter_analysis_toolkit/CMakeLists.txt
```diff
@@ -0,0 +1,527 @@
+cmake_minimum_required(VERSION 3.18)
+
+#-------------------------------------------------------------------------------
+# Compiler info.
```
Treece-Burgess, line 4: If you run a `cmake` command once and then try to run another `cmake` command again some output will not show. Specifically the `Compiler` section will only output "Using GCC" and the `OpenMP` section will not output anything.  For example, running the below two commands in succession:
```
cmake -DPAPI_DIR=/home/tburgess/daniel_fork/papi/src/test-install ..
cmake -DPAPI_DIR=/home/tburgess/daniel_fork/papi/src/test-install ..
```

Output:
```
------------------ Compiler ------------------
-- Using GCC.

-------------------- Arch --------------------
-- Detecting CPU architecture...
-- uname -m : x86_64
-- Setting ARCH=X86
-- Compiler supports AVX128
-- Hardware supports AVX128
-- Compiler supports AVX256
-- Hardware supports AVX256
-- Compiler supports AVX512
-- Hardware supports AVX512
-- Compiler supports AVX512VL
-- Hardware supports AVX512VL
-- Compiler supports FMA
-- Hardware supports FMA

-------------------- PAPI --------------------
PAPI_DIR = /home/tburgess/daniel_fork/papi/src/test-install

------------------- OpenMP -------------------

-------------------- MPI ---------------------
USE_MPI = OFF
-- User has not requested to use MPI
-- Configuring done (0.1s)
-- Generating done (0.3s)
-- Build files have been written to: /home/tburgess/daniel_fork/papi/src/counter_analysis_toolkit/build
```

Is this expected behavior?

src/counter_analysis_toolkit/README.md
dbarry9: This is done.

src/counter_analysis_toolkit/CMakeLists.txt
dbarry9: I have updated the PR to address this.

Treece-Burgess:
During a previous PICL meeting is was discuss to use  [meson-python](https://mesonbuild.com/meson-python/) instead of autotools. Due to this, I discussed with @dbarry9  if continuing to keep this PR open was applicable. We both agreed that the CAT build system should be inline with the PAPI build system. Therefore, I will be closing this PR.

PR #458, opened 2025-07-23
Cuda: For the CTC metrics, add a note to each metrics description that the Pefworks API does not support profiling them

Treece-Burgess:
## Pull Request Description
CTC metrics such as:
```
cuda:::CTC.TriageCompute.ctc__rx_bytes
cuda:::CTC.TriageCompute.ctc__rx_bytes.peak_sustained
cuda:::CTC.TriageCompute.ctc__rx_bytes.peak_sustained_active
```

are found on NVIDIA GH200's. These metrics can be successfully added, but are unable to be profiled:
```
[tburgess@hopper1 bin]$ ./papi_command_line cuda:::CTC.TriageCompute.ctc__rx_bytes:stat=max:device=0

Successfully added: cuda:::CTC.TriageCompute.ctc__rx_bytes:stat=max:device=0

Error! PAPI_start
```

CTC metrics [require](https://forums.developer.nvidia.com/t/unable-to-profile-the-metric-ctc-triagecompute-ctc-rx-bytes-avg-on-a-machine-with-1-gh200/338432/6?u=tburgess) the CUPTI Range Profling API to be successfully profiled. Currently the `cuda` component uses the Perfworks API.

This PR adds the following note to the CTC metrics description:
```
NOTE: The NVIDIA Perfworks API that the cuda component utilizes does not support profiling CTC metrics.
```

# Testing
Testing was done on a machine with 1 * GH200 using Cuda Toolkit 12.6.3.

For `./papi_native_avail`, CTC metrics now show the note:
```
| cuda:::CTC.TriageCompute.ctc__rx_bytes                                       |      
|            # of bytes of CTC RX data, at 2048B granularity. Units=(bytes). Nu|
|            mpass=1. NOTE: The NVIDIA Perfworks API that the cuda component ut|
|            ilizes does not support profiling CTC metrics. 
```

And non-CTC metrics do not show this note.

For `PAPI_get_event_info( ... )`, using a CTC metric will correctly show the note for the `long_descr` member variable in the structure `PAPI_event_info_t`:
```
Long Descr: # of bytes of CTC RX data, at 2048B granularity. Units=(bytes). Numpass=1. NOTE: The NVIDIA Perfworks API that the cuda component utilizes does not support profiling CTC metrics.
```

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**


src/components/cuda/cupti_profiler.c
```diff
@@ -2648,6 +2648,17 @@ static int get_metric_properties(const char *pChipName, const char *metricName,
         return PAPI_EBUF;
     }
 
+    // CTC metrics are not supported by the Perfworks API; therefore, append a note to the CTC metrics description
```
dbarry9, line 2651: Just to clarify, the reason you do not need to encase the following code in an `#if defined(API_PERFWORKS)` is because the routines in `cupti_profiler.c` are only called when the check `#if defined(API_PERFWORKS)` is successful in `cupti_dispatch.c`?

Treece-Burgess: Yes, the Cuda component currently only has support for Perfworks API. So `cupti_profiler.c` only has Perfworks API calls. Code for the Events API will be placed in the file `cupti_events.c` and should never collide. As the check here:
```
int cupti_api = cuptic_determine_runtime_api();
if (cupti_api == API_PERFWORKS) {
```

in `cupti_dispatch.c` will always check to see if we use Perfworks or Events.

dbarry9 (approved):

dbarry9:
I am doing a final test of the most recent commit.

src/components/cuda/cupti_profiler.c
```diff
@@ -2648,6 +2648,17 @@ static int get_metric_properties(const char *pChipName, const char *metricName,
...
+    char *ctcMetric = strstr(metricName, "CTC.");
```
dbarry9: When I test these changes, there is no note appended to the event descriptions of the events named cuda:::ctc__*; however, these suffer from the same issue as the events that contain the pattern "CTC." in them:
> $ papi_command_line cuda:::ctc__rx_bytes
> This utility lets you add events from the command line interface to see if they work.
> Successfully added: cuda:::ctc__rx_bytes
> Error! PAPI_start
> 

I think we need to update this substring search to something like `char *ctcMetric = strstr(metricName, "ctc__");` to encompass all cases.

dbarry9 (changes requested):

src/components/cuda/cupti_profiler.c
Treece-Burgess: Thanks for catching that and I think that is a good suggestion to use `ctc__` as the substring check.

I have made the change with the most recent push and the note now shows for `cuda:::ctc__rx_bytes`:
```
| cuda:::ctc__rx_bytes                                                         |      
|            # of bytes of CTC RX data, at 2048B granularity. Units=(bytes). Nu|
|            mpass=1. NOTE: The NVIDIA Perfworks API that the cuda component ut|
|            ilizes does not support profiling CTC metrics.                    |      
|     :stat=avg                                                                |      
|            Mandatory stat qualifier [avg, max, min, sum]                     |      
|            Mandatory device qualifier [0]                                    |      
```

dbarry9:
I have re-tested these changes on an NVIDIA GH200 480GB device with CUDA 12.8.1. The output from the utilities and the CUDA component tests are correct. The only notable difference is the output of papi_native_avail contains the string "NOTE: The NVIDIA Perfworks API that the cuda component utilizes does not support profiling CTC metrics." as intended.

dbarry9 (approved):

PR #460, opened 2025-07-24
ROCP_SDK: Allow different ROCprofiler-SDK installation paths.

adanalis:
## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/rocp_sdk/README.md
+++ b/src/components/rocp_sdk/README.md
@@ -53,6 +53,12 @@ Example:
 
 Note that this variable takes precedence over PAPI\_ROCP\_SDK\_ROOT.
 
+Additionally, the component supports installations where ROCprofiler-SDK does not reside inside the ROCm installation. For example, if installed from source (https://github.com/ROCm/rocprofiler-sdk). In such a case, the PAPI user must set PAPI_ROCP_SDK_ROOT to point to the ROCprofiler-SDK installation and PAPI_ROCM_ROOT to point to the ROCm installation.
+
+Example:
+    export PAPI_ROCP_SDK_ROOT=${HOME}/my_packages
+    export PAPI_ROCM_ROOT=/opt/rocm
+
 ## Known Limitations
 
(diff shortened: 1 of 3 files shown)


Treece-Burgess:
I am reviewing this PR.

src/components/rocp_sdk/README.md
```diff
@@ -53,6 +53,12 @@ Example:
 
 Note that this variable takes precedence over PAPI\_ROCP\_SDK\_ROOT.
 
+Additionally, the component supports installations where ROCprofiler-SDK does not reside inside the ROCm installation. In such a case, the PAPI user must set PAPI_ROCP_SDK_ROOT to point to the ROCprofiler-SDK installation and PAPI_ROCM_ROOT to point to the ROCm installation.
```
Treece-Burgess: Question: For the case of setting `PAPI_ROCP_SDK_ROOT` to an installation of `rocprofiler-sdk` outside of the ROCm installation, is it required for a user to have to build from [source](https://github.com/ROCm/rocprofiler-sdk)?

- If it is, I would explicitly state in the README that this option must follow AMD's directions for [Building ROCprofiler-SDK from source](https://github.com/ROCm/rocprofiler-sdk/blob/amd-staging/source/docs/install/installation.rst#building-rocprofiler-sdk-from-source).

- If it is not, then I think we would need to be more clear about what needs to be included in this `rocprofiler-sdk` installation: E.g.:
    - Do you need `librocprofiler-sdk.so` to be found and if so what path structure does it need?
        - From my testing yes this need to be found and needs to have a path structure of `lib/librocprofiler-sdk.so`. Otherwise you will error by not finding `librocprofiler-sdk.so`.
    - Do you need `share/rocprofiler-sdk`to be found?
        - From my testing yes this needs to be found.  Otherwise you will error by not finding `counter_defs.yaml`.

Treece-Burgess (changes requested):

Treece-Burgess (approved):

PR #482, opened 2025-09-17
PAPI CI: Explicitly set path to `libcupti.so` and `libcudart.so` to avoid error while loading shared libraries

Treece-Burgess:
## Pull Request Description
Some of the ICL systems, namely Guyot and Hexane have been updated to support Cuda Toolkit 13. Due to this change, these systems see their system-installed Cuda Toolkit versions' be 13. Therefore, we must tell the system at runtime where to find the library; otherwise, we run into the error `error while loading shared libraries: libcudart.so.12`.

This PR updates the Cuda component tests to now:

- Explicitly set the path to `libcudart.so` using `LD_LIBRARY_PATH`.
- Explicitly set the path to `libcupti.so` using `LD_LIBRARY_PATH` as a precautionary safety net. I do not encounter the error `error while loading shared libraries: libcupti.so.12` for the Cuda component tests, but I did encounter it with the sample codes in CUPTI.
- Use `module unload glibc`

Thank you to Geri for discussing the topic with me.

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/.github/workflows/ci_individual_component.sh
+++ b/.github/workflows/ci_individual_component.sh
@@ -49,9 +49,11 @@ fi
 
 ## Set the cuda component or the nvml component environment variable
 if [ "$COMPONENT" = "cuda" ] || [ "$COMPONENT" = "nvml" ]; then
+    module unload glibc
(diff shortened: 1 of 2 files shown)


djwoun (approved):
Reviewed together on Zoom

PR #493, opened 2025-10-07
sde: Create a run script for the component tests and avoid recompilation on each make

Treece-Burgess:
## Pull Request Description
Currently in the master branch, you must manually run each `sde` component test i.e. `./Counting_Set_MemLeak_Test`. This can become tedious if you are testing multiple different PAPI configurations causing you to have to manually re-run each test.

This PR introduces the script `run_sde_tests.sh` which will make and run each test. The script also takes one optional argument `-verbose` which will enable verbose output for a test. 

Along with introducing the run script. This PR restructures the targets/target dependencies such that they are not always recompiled.

## Testing
Testing was done on Methane at ICL which has the system setup:
- CPU: Intel Xeon Gold 6140
- GPU: 1 * A100
- OS: Rocky Linux 9.5

1. Running `sh run_sde_tests.sh`, resulted in:
    - All `sde` component tests passing
2. Running `sh run_sde_tests.sh -verbose`, resulted in:
    - All `sde` component tests passing and showing verbose output    

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/components/sde/tests/run_sde_tests.sh
+++ b/src/components/sde/tests/run_sde_tests.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+# Set path to components/sde/tests/lib as
+# this is needed for nearly all the tests
+# to run successfully.
+export LD_LIBRARY_PATH=$PWD/lib:$LD_LIBRARY_PATH
+
+# Add -verbose to show output, i.e. sh run_sde_tests.sh -verbose.
+if [ "$1" = "-verbose" ]; then
+    run_verbose="$1"
+fi
+
+make_sde_test_targets=(
+    "Minimal_Test"
+    "Minimal_Test++"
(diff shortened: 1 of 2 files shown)


src/components/sde/tests/run_sde_tests.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
```
dbarry9, line 2: I get the error:

> ./Counting_Set_MemLeak_Test: error while loading shared libraries: libCounting_Set.so: cannot open shared object file: No such file or directory

and similar errors for the other tests, as well.
Would it be a good idea to include a line, such as `export LD_LIBRARY_PATH=${PWD}/lib:${LD_LIBRARY_PATH}`?

PR #497, opened 2025-10-23
2025.10.23 updates to the intel_gpu component

dbarry9:
## Pull Request Description

- Some `intel_gpu` component tests were using deprecated environment variables.
- Introduce `PAPI_INTEL_GPU_ROOT` environment variable to be consistent with other components.

These changes have been tested using the Intel Arc B580 (Battlemage) and Iris Xe architectures.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/intel_gpu/linux_intel_gpu_metrics.c
+++ b/src/components/intel_gpu/linux_intel_gpu_metrics.c
@@ -205,8 +205,10 @@ intel_gpu_init_component(int cidx)
 		goto fn_fail;
 	}
 	if (GPUDetectDevice(&avail_devices, &num_avail_devices) || (num_avail_devices==0)) {
-		errStr = "The intel_gpu component does not detect metrics device.";
-		strncpy_se(_intel_gpu_vector.cmp_info.disabled_reason, PAPI_MAX_STR_LEN,
+		errStr = "The intel_gpu component does not detect metrics device. Ensure\n" \
+                 "                 that LD_LIBRARY_PATH contains the dependencies libze_loader.so,\n" \
+                 "                 libigdmd.so, and libigdml.so";
+		strncpy_se(_intel_gpu_vector.cmp_info.disabled_reason, PAPI_2MAX_STR_LEN,
 				   errStr, strlen(errStr));
 		retval = PAPI_ENOSUPP;
 		goto fn_fail;


src/components/intel_gpu/tests/Makefile
```diff
@@ -2,8 +2,16 @@ NAME=intel_gpu
 include ../../Makefile_comp_tests.target
 CFLAGS += -DDEBUG  -I../../../testlib
 
-INTEL_L0_HEADERS ?=/usr/include
-INTEL_L0_LIB64 ?=/usr/lib
+## the following set are used to define oneAPI Level Zero Metric API based implementation
+INTEL_L0_HEADERS = $(PAPI_INTEL_GPU_ROOT)/include
```
Treece-Burgess, line 9: Same question as `Rules.intel_gpu`.

src/components/intel_gpu/README.md
```diff
@@ -1,48 +1,56 @@
...
-* [Intel(R) Metrics Discovery Application Programming Interface](https://github.com/intel/metrics-discovery)
+* [Intel(R) Metrics Discovery Application Programming Interface (libigdmd.so)](https://github.com/intel/metrics-discovery)
+* [Intel(R) Metrics Library Application Programming Interface (libigdml.so)](https://github.com/intel/metrics-library)
 * /proc/sys/dev/i915/perf_stream_paranoid is set to "0"
 * User needs to be added into Linux render/video groups
 
-## Runtime environment:
+## Environment Variables
+
+PAPI requires the location of the Level Zero install directory. This can be
```
Treece-Burgess, line 23: Q: The only environment variable that is mentioned needs to be pointed to the location of the Level Zero install directory. However, you mention a few other prerequisites above, will they be under the same directory structure as Level Zero?

src/components/intel_gpu/README.md
```diff
@@ -1,48 +1,56 @@
...
-   ```
+*   ```sh
+    export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:<path-to-libze_loader>:<path-to-libigdmd>:<path-to-libigdml>
+    ```
 * To enable metrics
-	```sh
-	ZET_ENABLE_METRICS=1
-	```
+    ```sh
+    ZET_ENABLE_METRICS=1
```
Treece-Burgess, line 42: Without setting `ZET_ENABLE_METRICS=1`, `papi_component_avail` will show:
```
Name:   intel_gpu               Intel GPU performance metrics
   \-> Disabled: Please set ZET_ENABLE_METRICS=1 to access GPU metrics.
```

Is there no way to set this internally as it seems to be needed for the component even to be active.

src/components/intel_gpu/README.md
```diff
@@ -1,48 +1,56 @@
...
+Access to this directory is required at both compile (for include files) and
+at runtime (for libraries).
+
+## Runtime Environment:
 
-*  ```sh
-   export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:<path-to-libze_loader>:<path-to-libmd>
-   ```
+*   ```sh
+    export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:<path-to-libze_loader>:<path-to-libigdmd>:<path-to-libigdml>
```
Treece-Burgess, line 38: Q: On Headroom at Oregon (Intel Data Center Max 1100), I did not set a path to `libigdmd.so` for the intel_gpu component to be active. What cases is this needed to be explicitly set?

You may want to preface that before adding the `export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:<path-to-libze_loader>:<path-to-libigdmd>:<path-to-libigdml>`

src/components/intel_gpu/Rules.intel_gpu
```diff
@@ -1,7 +1,14 @@
 # $Id$
 
 ## the following set are used to define oneAPI Level Zero Metric API based implementation
-INTEL_L0_HEADERS ?= /usr/include
+INTEL_L0_HEADERS = $(PAPI_INTEL_GPU_ROOT)/include
```
Treece-Burgess, line 7: Q: Why not have `PAPI_INTEL_GPU_ROOT ?= /usr`, which you then could have:
```
INTEL_L0_HEADERS = $(PAPI_INTEL_GPU_ROOT)/include
INTEL_L0_LIB64 = $(PAPI_INTEL_GPU_ROOT)/lib
```

This would be inline more with how other components set their environment variables, e.g. cuda or rocm_smi.

src/components/intel_gpu/tests/readme.txt
```diff
@@ -1,6 +1,6 @@
 Test examples:
 To collect metrics, set ZET_ENABLE_METRICS=1
-To collect metrics in query mode, set ZET_ENABLE_API_TRACING_EXP=1
+To collect metrics in query mode, set ZE_ENABLE_TRACING_LAYER=1
```
Treece-Burgess, line 3: Q: I do remember the following output when you used the `intel_gpu` component: 
```
ZET_ENABLE_API_TRACING_EXP is deprecated. Use ZE_ENABLE_TRACING_LAYER instead
```
However, do you know if this is version or GPU based?

src/components/intel_gpu/tests/gpu_query_gemm.cc
```diff
@@ -45,7 +45,7 @@
 
 #if defined(ENABLE_PAPI)
 #include "papi.h" 
-const char *env_str = "ZET_ENABLE_API_TRACING_EXP=1";
+const char *env_str = "ZE_ENABLE_TRACING_LAYER=1";
```
Treece-Burgess, line 48: Using `ZE_ENABLE_TRACING_LAYER=1`, will result in 0 values being output and the error output of `EnableEventBasedQuery: failed with status 0x78000001, abort.` showing. This does not occur with the old env.
```
tburgess@headroom:~/daniel_fork/papi/src/components/intel_gpu/tests$ ./gpu_query_gemm 
Level Zero Matrix Multiplication (matrix size: 1024 x 1024, repeats 4 times)
Target device: Intel(R) Data Center GPU Max 1100
EnableEventBasedQuery: failed with status 0x78000001, abort.
Matrix multiplication time: 0.00351712 sec
Matrix multiplication time: 0.00351616 sec
Matrix multiplication time: 0.00351472 sec
Matrix multiplication time: 0.00351808 sec
Total execution time: 5.40232 sec
ComputeBasic.GpuTime                               ......  0
ComputeBasic.GpuCoreClocks                         ......  0
ComputeBasic.AvgGpuCoreFrequencyMHz                ......  0
Total time: 5.54342 sec
```

I encountered this on Headroom at Oregon with an Intel Data Center Max 1100.

src/components/intel_gpu/README.md
```diff
@@ -1,48 +1,56 @@
...
-	```sh
-	ZET_ENABLE_METRICS=1
-	```
+    ```sh
+    ZET_ENABLE_METRICS=1
+    ```
 * To change the sampling period from default 400000, 
-	```sh
+    ```sh
     METRICS_SAMPLING_PERIOD=value
```
Treece-Burgess, line 47: Q: I have tried changing this to multiple different values, but do not see a different in final values that output. Do you know how this exactly works?

src/components/intel_gpu/README.md
```diff
@@ -1,48 +1,56 @@
...
+    export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:<path-to-libze_loader>:<path-to-libigdmd>:<path-to-libigdml>
+    ```
 * To enable metrics
-	```sh
-	ZET_ENABLE_METRICS=1
-	```
+    ```sh
+    ZET_ENABLE_METRICS=1
+    ```
 * To change the sampling period from default 400000,
```
Treece-Burgess: There is an extraneous `,` here if you want to remove it in this PR as well.

src/components/intel_gpu/README.md
```diff
@@ -1,48 +1,56 @@
...
-## Enable intel_gpu component
+## Enable intel_gpu Component
 
 To enable intel_gpu component, the PAPI library is built with configure option as 
 
-	./configure --with-components="intel_gpu"
-
-It is requred to build with Intel oneAPI Level Zero header files. The directory of Level0 header files can be defined using INTEL_L0_HEADERS. Default installation location is /usr/include/level_zero/.
-
+    ./configure --with-components="intel_gpu"
```
Treece-Burgess: nit: You could follow suit with below on how you use ```  for the code blocks.

src/components/intel_gpu/README.md
Treece-Burgess: To add to this, when I run the `gpu_metric_list` test without setting `ZET_ENABLE_METRICS=1` I still see a single metric output instead of an error code:
```
tburgess@headroom:~/daniel_fork/papi/src/components/intel_gpu/tests$ ./gpu_metric_list
Name:  intel_gpu
Description: Intel GPU performance metrics
CPU_CLK_UNHALTED:THREAD_P
	Cycles in unhalted state., masks:Thread cycles when thread is not in halt state
Total 1 metrics are supported
```

src/components/intel_gpu/README.md
dbarry9: > What cases is this needed to be explicitly set?

This needs to be explicitly set when the shared object files are not located in one of the Linux default directories listed under /etc/ld.so.conf.d/ (usually /usr/lib64, /lib64, /usr/lib and /lib). I will add a preface.

src/components/intel_gpu/README.md
dbarry9: There is no guarantee that they will be under the same directory structure. The Level Zero Loader library has both header and shared object files needed by the `intel_gpu` component, but the other two libraries only contain shared object files needed by the component. As such, there is no need for a dedicated environment variable for the other two prerequisites because they can be resolved via the `LD_LIBRARY_PATH` variable if need be.

src/components/intel_gpu/README.md
dbarry9: I am not familiar with this. We may need to raise this as a separate issue.

src/components/intel_gpu/Rules.intel_gpu
dbarry9: Excellent idea! Thank you.

src/components/intel_gpu/tests/gpu_query_gemm.cc
dbarry9: Thank you for uncovering this error. I am looking into it.

src/components/intel_gpu/tests/readme.txt
dbarry9: According to the [Level Zero Specification](https://oneapi-src.github.io/level-zero-spec/level-zero/latest/tools/PROG.html), `ZET_ENABLE_API_TRACING_EXP` is deprecated, but there is no mention of specific devices this applies to.

Furthermore, this [page](https://github.com/oneapi-src/level-zero/blob/master/source/layers/tracing/README.md) states that `ZE_ENABLE_TRACING_LAYER` must be set to 1, also with no mention of device compatibility.

These indicate that that `ZE_ENABLE_TRACING_LAYER` is the proper environment variable to set, unless someone is using an old version of the Level Zero API.

Edit: To answer your question, this is API version-based.

src/components/intel_gpu/tests/gpu_query_gemm.cc
dbarry9: I will create an issue and a separate PR for this. Update: I have created Issue #523 to track this bug.

Treece-Burgess: Got it, thanks for creating the Issue!

PR #505, opened 2025-11-06
amd_smi: fix linker error

dbarry9:
## Pull Request Description
I encountered the following error when I configure PAPI with both the amd_smi and rocp_sdk components:
>Error:  Unable to find cray-libsci/24.11.0 libraries compatible with gcc/12.2.0 targeting 'x86-trento'.

Unloading the `cray-libsci` module resolves the above error, but then yields the following:
> cc -DPAPI_NUM_COMP=3 -std=c11 -O2 -I. -I../../.. -I../../../testlib -I../../../validation_tests -I../../../include -I/opt/rocm-7.0.2/include -I/opt/rocm-7.0.2/include/rocm_smi -o amdsmi_hello amdsmi_hello.o ../../../testlib/libtestlib.a ../../../libpapi.a -L/opt/rocm-7.0.2/lib -lamd_smi -Wl,-rpath,/opt/rocm-7.0.2/lib64 -Wl,-rpath,/opt/rocm-7.0.2/lib -ldl -g -lpthread
> /usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: ../../../libpapi.a(sdk_class.o): undefined reference to symbol '_ZNKSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE7compareEPKc@@GLIBCXX_3.4.21'
> /usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: /usr/lib64/libstdc++.so.6: error adding symbols: DSO missing from command line

This can be resolved by appending -lstdc++ to the linker flags, and the most consistent manner to achieve this is to use the existing LDFLAGS variable by including Makefile_comp_tests.target. However, this also overwrites the CC variable from cc (which in my case aliases a different compiler). In my testing, this results in the error:
> /usr/bin/ld: /opt/rocm-7.0.2/lib/libamd_smi.so: undefined reference to `std::ios_base_library_init()@GLIBCXX_3.4.32'

This error is resolved by using HIPCC instead of CC for the targets.

These changes have been tested on Frontier, which has the AMD MI250X architecture and ROCm 7.0.2.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/amd_smi/tests/Makefile
+++ b/src/components/amd_smi/tests/Makefile
@@ -1,3 +1,6 @@
+NAME=amd_smi
+include ../../Makefile_comp_tests.target
+
 # ROCm + PAPI build settings
 # Set PAPI_AMDSMI_ROOT to your ROCm install prefix (e.g., /opt/rocm or /opt/rocm-6.4.0).
 # Used to find AMD SMI / HIP / ROCm headers and libraries. You can override on the command line.
@@ -31,16 +34,16 @@ EFFECTIVE_INCLUDE_C := $(INCLUDE_PAPI) $(INCLUDE_AMDSMI)
 EFFECTIVE_INCLUDE_HIP := $(EFFECTIVE_INCLUDE_C) $(INCLUDE_HIP)
 
 # Linker flags
-RPATH_FLAGS     ?= -Wl,-rpath,$(PAPI_AMDSMI_ROOT)/lib64 -Wl,-rpath,$(PAPI_AMDSMI_ROOT)/lib
-COMMON_LDFLAGS  = $(RPATH_FLAGS) -ldl -g -lpthread
-LDFLAGS_AMDSMI  = $(UTILOBJS) $(PAPILIB) -L$(PAPI_AMDSMI_ROOT)/lib -lamd_smi $(COMMON_LDFLAGS)
-LDFLAGS_GEMM    = $(UTILOBJS) $(PAPILIB) -L$(PAPI_AMDSMI_ROOT)/lib -lrocblas $(COMMON_LDFLAGS)
+COMMON_LDFLAGS  = $(UTILOBJS) $(PAPILIB) -L$(PAPI_AMDSMI_ROOT)/lib -L$(PAPI_AMDSMI_ROOT)/lib64 $(LDFLAGS) -pthread
+LDFLAGS_AMDSMI  = $(COMMON_LDFLAGS) -lamd_smi
+LDFLAGS_GEMM    = $(COMMON_LDFLAGS) -lrocblas
 
 # AMD SMI tests to build with 'make tests'
 TESTS_AMDSMI := amdsmi_energy_monotonic amdsmi_ctx_conflict amdsmi_set_test
 
 # Default target
-all: amdsmi_hello amdsmi_basics amdsmi_gemm $(TESTS_AMDSMI)
+TESTS = amdsmi_hello amdsmi_basics amdsmi_gemm $(TESTS_AMDSMI)
+amd_smi_tests: $(TESTS)
 
 # ------------------------
 # Compile rules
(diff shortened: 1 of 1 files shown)


src/components/amd_smi/tests/Makefile
```diff
@@ -31,16 +34,15 @@ EFFECTIVE_INCLUDE_C := $(INCLUDE_PAPI) $(INCLUDE_AMDSMI)
 EFFECTIVE_INCLUDE_HIP := $(EFFECTIVE_INCLUDE_C) $(INCLUDE_HIP)
 
 # Linker flags
-RPATH_FLAGS     ?= -Wl,-rpath,$(PAPI_AMDSMI_ROOT)/lib64 -Wl,-rpath,$(PAPI_AMDSMI_ROOT)/lib
-COMMON_LDFLAGS  = $(RPATH_FLAGS) -ldl -g -lpthread
-LDFLAGS_AMDSMI  = $(UTILOBJS) $(PAPILIB) -L$(PAPI_AMDSMI_ROOT)/lib -lamd_smi $(COMMON_LDFLAGS)
-LDFLAGS_GEMM    = $(UTILOBJS) $(PAPILIB) -L$(PAPI_AMDSMI_ROOT)/lib -lrocblas $(COMMON_LDFLAGS)
+LDFLAGS_AMDSMI  = $(UTILOBJS) $(PAPILIB) -L$(PAPI_AMDSMI_ROOT)/lib -L$(PAPI_AMDSMI_ROOT)/lib64 -lamd_smi $(LDFLAGS)
+LDFLAGS_GEMM    = $(UTILOBJS) $(PAPILIB) -L$(PAPI_AMDSMI_ROOT)/lib -L$(PAPI_AMDSMI_ROOT)/lib64 -lrocblas $(LDFLAGS)
```
Treece-Burgess: Suggestion: The only main difference between `LDFLAGS_AMDSMI` and `LDFLAGS_GEMM` is `-lrocblas`. Instead of having a completely new variable for the GEMM case. You could simply just append `-lrocblas` to the target `amdsmi_gemm`.

djwoun:
Leaving Comment For Documentation.

cc is a C driver; wrappers like CrayPE may inject -lstdc++, plain cc does not. libamd_smi.so is C++, and its dependency doesn’t resolve C++ symbols in our objects. So we must link with hipcc (or a C++ driver like CrayPE’s CC), or add -lstdc++.
LDFLAGS applies the same linker options to every target, keeping libraries and paths consistent.

src/components/amd_smi/tests/Makefile
```diff
@@ -31,16 +34,15 @@ EFFECTIVE_INCLUDE_C := $(INCLUDE_PAPI) $(INCLUDE_AMDSMI)
 EFFECTIVE_INCLUDE_HIP := $(EFFECTIVE_INCLUDE_C) $(INCLUDE_HIP)
 
 # Linker flags
-RPATH_FLAGS     ?= -Wl,-rpath,$(PAPI_AMDSMI_ROOT)/lib64 -Wl,-rpath,$(PAPI_AMDSMI_ROOT)/lib
-COMMON_LDFLAGS  = $(RPATH_FLAGS) -ldl -g -lpthread
-LDFLAGS_AMDSMI  = $(UTILOBJS) $(PAPILIB) -L$(PAPI_AMDSMI_ROOT)/lib -lamd_smi $(COMMON_LDFLAGS)
-LDFLAGS_GEMM    = $(UTILOBJS) $(PAPILIB) -L$(PAPI_AMDSMI_ROOT)/lib -lrocblas $(COMMON_LDFLAGS)
+LDFLAGS_AMDSMI  = $(UTILOBJS) $(PAPILIB) -L$(PAPI_AMDSMI_ROOT)/lib -L$(PAPI_AMDSMI_ROOT)/lib64 -lamd_smi $(LDFLAGS)
```
djwoun: Add back -lpthread to LDFLAGS_AMDSMI and LDFLAGS_GEMM

djwoun (approved):

PR #504, opened 2025-11-04
Fix read/reset for topdown component

daboehme:
## Pull Request Description

I found that the topdown component does not handle `PAPI_reset` and `PAPI_read` correctly. This PR fixes that.

I'm developing Caliper, an instrumentation-based profiler. We call `PAPI_read` and `PAPI_reset` at each instrumentation point so that we can properly associate events with the instrumented code region. I noticed that the reset did not appear to work for the topdown events, and looking at the code I found that it was indeed a no-op. I managed to restructure the code so that reset works now, but I'm no expert so please check if I missed anything. Here's what I did:

* Before, `_topdown_start` resetted the counters and `_topdown_reset` was a no-op. Also, `_topdown_read` would just call `_topdown_stop`, which in turn would close the file descriptors necessary to reset the counters.
* Now, the counter reading code is in `_topdown_read`, and `_topdown_stop` calls `_topdown_read` (and closes the FD's).
* The reset code is now in `_topdown_reset`, and `_topdown_start` calls `_topdown_reset`.

That does seem to do the trick for me. Here's example output with a code that alternates STREAM triad and matrix multiply in a loop, which have wildly different topdown stats. Before, the events would not reset and the topdown stats just converge to an overall program average:

```
iter kernel     retiring  bad_spec fe_bound  be_bound
   0 triad      21.176500 3.137260 22.352900 53.725500
   0 matmul_f64 76.862700 5.882350 12.156900  5.490200
   1 triad      70.588200 5.490200 11.372500 12.941200
   1 matmul_f64 75.686300 5.490200 11.372500  7.450980
   2 triad      72.549000 5.098040 11.372500 10.980400
   2 matmul_f64 75.294100 5.490200 11.372500  8.235290
   3 triad      73.333300 5.098040 11.372500 10.588200
   3 matmul_f64 75.294100 5.098040 11.372500  8.627450
   4 triad      73.725500 5.098040 11.372500 10.196100
   4 matmul_f64 75.294100 5.098040 11.372500  8.627450
   5 triad      73.725500 5.098040 11.372500 10.196100
   5 matmul_f64 74.902000 5.098040 11.372500  8.627450
   6 triad      74.117600 5.098040 11.372500  9.803920
   6 matmul_f64 74.902000 5.098040 11.372500  9.019610
   7 triad      74.117600 5.098040 11.372500  9.803920
   7 matmul_f64 74.902000 5.098040 11.372500  9.019610
```

[...]

PR #509, opened 2025-11-11
topdown: Restructure code for error checking with sys call, make _topdown_reset actually reset, and avoid closing file descriptors when calling PAPI_read

Treece-Burgess:
## Pull Request Description
Background:
Currently in the master branch when using the `topdown` component the following issues can arise:
1.  Calling `PAPI_read` will result in the file descriptors being closed. As internally `_topdown_read` calls `_topdown_stop` which closes the file descriptors.
2.  `PAPI_reset` does not work properly as `_topdown_reset` does not perform any operations. 

Note: Thanks to David Boehme for opening PR #504 which brings this issue to light. This PR continues to iterate on his work.

Solution:
I followed `perf_event`'s implementation closely to restructure the internal `_topdown_*` calls. Specifically:

1. `_topdown_stop` - Now disables the event group rather than closing the file descriptors. 
    - Closing the file descriptors in stop would result in an error IF directly after a user called start.
2. `_topdown_reset` - Calls `_wrapper_perf_event_ioc_reset`, which resets the event count specified by `slots_fd` and `metrics_fd`to zero.
3. `_topdown_read` - Takes the implemented code from `topdown_stop` (minus the code that closed the file descriptors).
4. `_topdown_start` -  Calls `_wrapper_perf_event_ioc_reset`, which resets the event count specified by `slots_fd` and `metrics_fd`to zero. Along with, enabling the events specified by the file descriptors `slots_fd` and `metrics_fd`.

## Testing
Testing was done on Picard at Oregon, system info is as follows:
- CPU: Intel(R) Xeon(R) Gold 6430
- OS: RHEL 8.10

The PAPI utilities (`papi_component_avail`, `papi_native_avail`, and `papi_command_line`) all ran successfully. Along with this, the `topdown` component tests all ran successfully too.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/topdown/topdown.c
+++ b/src/components/topdown/topdown.c
@@ -568,14 +568,14 @@ _topdown_init_control_state(hwd_control_state_t *ctl)
 	int metrics_fd = -1;
 	void *slots_p, *metrics_p;
 
-	/* set up slots */
+	// Set up slots
 	memset(&slots, 0, sizeof(slots));
 	slots.type = PERF_TYPE_RAW;
 	slots.size = sizeof(struct perf_event_attr);
 	slots.config = 0x0400ull;
 	slots.exclude_kernel = 1;
 
-	/* open slots */
+	// Open slots
 	slots_fd = perf_event_open(&slots, 0, -1, -1, 0);
 	if (slots_fd < 0)
 	{
@@ -583,7 +583,7 @@ _topdown_init_control_state(hwd_control_state_t *ctl)
 		goto fn_fail;
 	}
 
-	/* memory mapping the fd to permit _rdpmc calls from userspace */
+	// Memory mapping the fd to permit _rdpmc calls from userspace
 	slots_p = mmap(0, getpagesize(), PROT_READ, MAP_SHARED, slots_fd, 0);
 	if (slots_p == (void *) -1L)
 	{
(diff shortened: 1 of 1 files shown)


deater:
sorry for the delay reviewing this

the answer to both your questions is probably "historical reasons".  The perf_event component is really old, predating even my start of work on PAPI.
I think you can add events at init time on topdown because you always know in advance which events are going to be opened, while with regular perf_event events can be added/removed at any time so we do it at update time.  For Q#2  it is possible the events aren't possibly removed (when would that be done, at eventset destroy time?).  We probably just trust Linux to close everything out properly when the whole program exits.

For overall review of this changeset, as far as I can tell it looks like it will be fine.  It would be nice if there was some sort of new test added that tested the reset functionality.

Some comments:

There were various things that made it hard to review the changes that maybe should have been a separate change first:
* There were a number of places where /* */ comments were changed to //.  Is this a PAPI style-guide thing?
* All the changes seem to be using windows-style line endings with ^M (carriage return) instead of the Linux style originally used
* In at least one function some variable declarations were moved around which didn't seem to make any changes to behavior but it's a bit unclear why it was done
* A typo, "enable the event's" should probably be "enable the events"

I tested on our raptor lake machine and all the tests passed though I didn't analyze deeply if the results were expected results or not

deater:
I personally have no strong opinions on the // vs /* */ change, just it seemed like it happened to half the file and not the rest and couldn't remember if we had some sort of style guide for this kind of thing.  There was a big "great re-indenting" event many years ago that reset everything to the same formatting but I am sure things have drifted since then.

In retrospect the ^M in the file probably came from our end, sorry about that.

dbarry9:
I tested this PR on a system containing the Intel Sapphire Rapids architecture. The `topdown` component tests and PAPI utilities behave properly.

PR #522, opened 2025-12-12
Pin GHA dependencies by hash

williamjallen:
## Pull Request Description

Hash-pinned dependencies are an important requirement enforced by the OpenSSF scorecard tool: https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies

This commit pins all GitHub Actions workflows to a commit hash.

### DIFF
--- a/.github/workflows/appio_component_workflow.yml
+++ b/.github/workflows/appio_component_workflow.yml
@@ -20,6 +20,6 @@ jobs:
     runs-on: [self-hosted, cpu_intel]
     timeout-minutes: 60
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - name: appio component tests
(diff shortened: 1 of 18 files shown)


Treece-Burgess:
@williamjallen Thanks for posting this PR. I have a few clarification questions. 

1. The idea behind pinning by commit hash rather than tag's is to the mitigate security risks, correct? As tags are mutable compared to commit hashes which are immutable. 

For an example, if we currently have the v4 tag pinned for `actions/checkout`  a malicious user could take over the `actions/checkout` repo and inject malicious code at that tag. Which then would result in that malicious code being checked out and ran through our systems. 

2. As stands we currently do not have dependabot setup. I saw that it is very common to use dependabot to automatically create PR's to update the pinned hashes. Is the major benefit of continuously updating just getting the new security features if any are added?

Treece-Burgess (approved):
From a security standpoint this change makes sense.

The commit sha's that are used for `actions/checkout` and `actions/upload-artifact` match what are in their respective repositories for v4.

PR #525, opened 2025-12-19
Support for NVML counters: GPU energy, instant GPU power, avg memory power

mpatrou:
## Pull Request Description
Sibling PR (1/2) split into #529

The code 
* adds support  for 3 NVML energy counters:
  - energy_consumption
  - gpu_inst_power
  - gpu_memory_avg_power

## Author Checklist
- [ ] **Description**
_Why_ this PR exists. Reference all relevant information, including _background_, _issues_, _test failures_, etc
- [ ] **Commits**
_Commits_ are self contained and only do one thing
_Commits_ have a header of the form: `module: short description`
_Commits_ have a body (whenever relevant) containing a detailed description of the addressed problem and its solution
- [ ] **Tests**
The PR needs to pass all the tests
*co-authored with Oscar Hernandez

PR #529, opened 2026-01-06
 papi_hl region-based calculations

mpatrou:
## Pull Request Description

Sibling PR (2/2) split from #525
It updates papi_hl:

- region-based energy counter calculations: delta for energy_consumption above
- region-based energy counter calculations: average between before and after, used in gpu_inst_power

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/high-level/papi_hl.c
+++ b/src/high-level/papi_hl.c
@@ -600,10 +600,23 @@ static int _internal_hl_create_components()
       }
 
       /* change event type to instantaneous for specific events */
-      /* we consider all nvml events as instantaneous values */
-      if( (strstr(requested_event_names[i], "nvml:::") != NULL) ) {
+      // NVML events
+      if( strstr(requested_event_names[i], "nvml:::") != NULL) {
+        // for nvml total_energy_consumption, we calculate a delta value
+        if( strstr(requested_event_names[i], "total_energy_consumption") != NULL ) {
+            event_type = 0;
+            verbose_fprintf(stdout, "PAPI-HL Info: The event \"%s\" will be stored as delta value.\n", requested_event_names[i]);
+        }
+        else if( strstr(requested_event_names[i], "gpu_inst_power") != NULL || strstr(requested_event_names[i], "gpu_memory_avg_power") != NULL) {
+       	// for nvml gpu_inst_power and gpu_memory_avg_power, we calculate an average value
+           event_type = 2;
+           verbose_fprintf(stdout, "PAPI-HL Info: The event \"%s\" will be stored as average value.\n", requested_event_names[i]);
+        }
+        else {
+        // all the rest nvml events, we consider them as instantaneous values
          event_type = 1;
          verbose_fprintf(stdout, "PAPI-HL Info: The event \"%s\" will be stored as instantaneous value.\n", requested_event_names[i]);
+        }
       }
 
       /* check if event is supported on current machine */
@@ -883,10 +896,13 @@ static inline int _internal_hl_add_values_to_region( regions_t *node, enum regio
          for ( j = 0; j < components[i].num_of_events; j++ ) {
             if ( ( read_node = _internal_hl_insert_read_node(&node->values[cmp_iter].read_values) ) == NULL )
                return ( PAPI_ENOMEM );
-            if ( components[i].event_types[j] == 1 )
+            if ( components[i].event_types[j] == 1 ) //instantaneous
                read_node->value = _local_components[i].values[j];
-            else
+            else if ( components[i].event_types[j] == 2 ) //region-average
+               read_node->value = (_local_components[i].values[j] + node->values[cmp_iter].begin)/2;
+            else //delta
                read_node->value = _local_components[i].values[j] - node->values[cmp_iter].begin;
+                 
             cmp_iter++;
          }
       }
@@ -898,11 +914,13 @@ static inline int _internal_hl_add_values_to_region( regions_t *node, enum regio
       for ( i = 0; i < num_of_components; i++ )
          for ( j = 0; j < components[i].num_of_events; j++ ) {
             /* if event type is instantaneous only save last value */
-            if ( components[i].event_types[j] == 1 ) {
+            if ( components[i].event_types[j] == 1 )  //instantaneous
                node->values[cmp_iter].region_value = _local_components[i].values[j];
-            } else {
-               node->values[cmp_iter].region_value = _local_components[i].values[j] - node->values[cmp_iter].begin;
-            }
+            else if ( components[i].event_types[j] == 2 )  //region-average
+               node->values[cmp_iter].region_value = (_local_components[i].values[j] + node->values[cmp_iter].begin)/2;                  
+            else //delta
+               node->values[cmp_iter].region_value = _local_components[i].values[j] - node->values[cmp_iter].begin;               
+            
             cmp_iter++;
          }
    }
@@ -1237,6 +1255,9 @@ static void _internal_hl_json_definitions(FILE* f, bool beautifier)
          const char *event_type = "delta";
          if ( components[i].event_types[j] == 1 )
             event_type = "instant";
+         if ( components[i].event_types[j] == 2 )
+            event_type = "region-average";
+
          const PAPI_component_info_t* cmpinfo;
          cmpinfo = PAPI_get_component_info( components[i].component_id );
 


src/high-level/papi_hl.c
```diff
@@ -606,6 +606,18 @@ static int _internal_hl_create_components()
          verbose_fprintf(stdout, "PAPI-HL Info: The event \"%s\" will be stored as instantaneous value.\n", requested_event_names[i]);
       }
 
+      // except from nvml energy_consumption delta
+      if( (strstr(requested_event_names[i], "nvml:::") != NULL) && (strstr(requested_event_names[i], "energy_consumption") != NULL) ) {
```
Treece-Burgess: Change `energy_consumption` to `total_energy_consumption` as that was the finalized event in PR #525.

src/high-level/papi_hl.c
```diff
@@ -883,10 +898,13 @@ static inline int _internal_hl_add_values_to_region( regions_t *node, enum regio
          for ( j = 0; j < components[i].num_of_events; j++ ) {
             if ( ( read_node = _internal_hl_insert_read_node(&node->values[cmp_iter].read_values) ) == NULL )
                return ( PAPI_ENOMEM );
-            if ( components[i].event_types[j] == 1 )
+            if ( components[i].event_types[j] == 1 ) //instantaneous
                read_node->value = _local_components[i].values[j];
-            else
+            else if ( components[i].event_types[j] == 2 ) // average
```
Treece-Burgess: nit: Could you change the comment average to region-average to be inline with the `event_type` value below?

src/high-level/papi_hl.c
```diff
@@ -898,11 +916,13 @@ static inline int _internal_hl_add_values_to_region( regions_t *node, enum regio
       for ( i = 0; i < num_of_components; i++ )
          for ( j = 0; j < components[i].num_of_events; j++ ) {
             /* if event type is instantaneous only save last value */
-            if ( components[i].event_types[j] == 1 ) {
+            if ( components[i].event_types[j] == 1 )  //instantaneous
                node->values[cmp_iter].region_value = _local_components[i].values[j];
-            } else {
-               node->values[cmp_iter].region_value = _local_components[i].values[j] - node->values[cmp_iter].begin;
-            }
+            else if ( components[i].event_types[j] == 2 )  // average
```
Treece-Burgess: Likewise as above.

src/high-level/papi_hl.c
```diff
@@ -790,15 +802,16 @@ static int _internal_hl_start_counters()
 
    if ( state == PAPIHL_ACTIVE ) {
       for ( i = 0; i < num_of_components; i++ ) {
-         if ( ( retval = PAPI_start( _local_components[i].EventSet ) ) != PAPI_OK )
+         if ( ( retval = PAPI_start( _local_components[i].EventSet ) ) != PAPI_OK ){
```
Treece-Burgess: There are a few locations in the PR where syntax has been updated. Can you revert these changes to the original?

I will comment "Here" for other occurrences.

src/high-level/papi_hl.c
```diff
@@ -790,15 +802,16 @@ static int _internal_hl_start_counters()
...
             return (retval );
-
+         }
```
Treece-Burgess: Here.

src/high-level/papi_hl.c
```diff
@@ -790,15 +802,16 @@ static int _internal_hl_start_counters()
...
          /* warm up PAPI code paths and data structures */
          if ( ( retval = PAPI_read_ts( _local_components[i].EventSet, _local_components[i].values, &cycles ) != PAPI_OK ) ) {
             return (retval );
          }
       }
       _papi_hl_events_running = 1;
+
```
Treece-Burgess: Here.

src/high-level/papi_hl.c
```diff
@@ -853,6 +866,7 @@ static inline reads_t* _internal_hl_insert_read_node(reads_t** head_node)
    return new_node;
 }
 
+
```
Treece-Burgess: Here.

src/high-level/papi_hl.c
```diff
@@ -868,8 +882,9 @@ static inline int _internal_hl_add_values_to_region( regions_t *node, enum regio
       node->values[1].begin = ts;
       /* events from components */
       for ( i = 0; i < num_of_components; i++ )
-         for ( j = 0; j < components[i].num_of_events; j++ )
+         for ( j = 0; j < components[i].num_of_events; j++ ){
```
Treece-Burgess: Here.

src/high-level/papi_hl.c
```diff
@@ -868,8 +882,9 @@ static inline int _internal_hl_add_values_to_region( regions_t *node, enum regio
...
             node->values[cmp_iter++].begin = _local_components[i].values[j];
+         }
```
Treece-Burgess: Here.

src/high-level/papi_hl.c
Treece-Burgess: Q: Why not restructure the conditionals to be as follows:
```
      // For the nvml event total_energy_consumption we consider it a delta value
      if ( (strstr(requested_event_names[i], "nvml:::") != NULL) && (strstr(requested_event_names[i], "total_energy_consumption") != NULL) ) {
         event_type = 0; 
         verbose_fprintf(stdout, "PAPI-HL Info: The event \"%s\" will be stored as delta value.\n", requested_event_names[i]);
      }    
      // For the nvml event gpu_inst_power we consider it a region-average value
      else if ( (strstr(requested_event_names[i], "nvml:::") != NULL) && (strstr(requested_event_names[i], "gpu_inst_power") != NULL) ){
         event_type = 2; 
         verbose_fprintf(stdout, "PAPI-HL Info: The event \"%s\" will be stored as average value.\n", requested_event_names[i]);
      }    
      // For all nvml events besides the above listed two we consider them instantaneous values
      else {
         event_type = 1; 
         verbose_fprintf(stdout, "PAPI-HL Info: The event \"%s\" will be stored as instantaneous value.\n", requested_event_names[i]);
      } 
```

From my testing this will result in the same output as the current setup, but is a bit more fluid to myself atleast.

src/high-level/papi_hl.c
```diff
@@ -606,6 +606,18 @@ static int _internal_hl_create_components()
...
       }
 
+      // except from nvml energy_consumption delta
+      if( (strstr(requested_event_names[i], "nvml:::") != NULL) && (strstr(requested_event_names[i], "energy_consumption") != NULL) ) {
+         event_type = 0;
+         verbose_fprintf(stdout, "PAPI-HL Info: The event \"%s\" will be stored as delta value.\n", requested_event_names[i]);
+      }
+
+      // except from nvml gpu_inst_power average
+      if( (strstr(requested_event_names[i], "nvml:::") != NULL) && (strstr(requested_event_names[i], "gpu_inst_power") != NULL) ) {
```
Treece-Burgess: Q: As you are wanting to calculate a region average for event type 2, should `gpu_inst_power` actually be `gpu_memory_avg_power`?

src/high-level/papi_hl.c
```diff
@@ -1055,20 +1075,19 @@ static int _internal_hl_read_counters()
...
       HLDBG("Thread-ID:%lu, Component-ID:%d\n", PAPI_thread_id(), components[i].component_id);
       for ( j = 0; j < components[i].num_of_events; j++ ) {
         HLDBG("Thread-ID:%lu, %s:%lld\n", PAPI_thread_id(), components[i].event_names[j], _local_components[i].values[j]);
       }
 
-      if ( retval != PAPI_OK )
-         return ( retval );
    }
+   /* get cycles for last component */
+   retval = PAPI_read_ts( _local_components[num_of_components - 1].EventSet, _local_components[num_of_components - 1].values, &_local_cycles );
```
Treece-Burgess: Q: Why is the change to this section needed? It does not appear to me that it is behaving correctly. 

For instance, if I look solely at `gpu_memory_avg_power` between the master branch and your branch this event remains as an instant event type. However, in the master branch I get a value of 3472328295419250878, but in your branch I get a value of `35295`. If I take your branch and change this section back to how it is in master then I get a comparable value again.

src/high-level/papi_hl.c
```diff
@@ -883,10 +898,13 @@ static inline int _internal_hl_add_values_to_region( regions_t *node, enum regio
          for ( j = 0; j < components[i].num_of_events; j++ ) {
             if ( ( read_node = _internal_hl_insert_read_node(&node->values[cmp_iter].read_values) ) == NULL )
                return ( PAPI_ENOMEM );
-            if ( components[i].event_types[j] == 1 )
+            if ( components[i].event_types[j] == 1 ) //instantaneous
                read_node->value = _local_components[i].values[j];
-            else
+            else if ( components[i].event_types[j] == 2 ) // average
+               read_node->value = (_local_components[i].values[j] + node->values[cmp_iter].begin)/2;
```
Treece-Burgess, line 902: Q: What is your exact use case for wanting to use this formula instead of say a running average? Do you have extremely tiny regions that you are wanting to measure?

We have two utilities that measure average power in a more fine grained approach. One being in [powercap](https://github.com/icl-utk-edu/papi/blob/master/src/components/powercap/utils/powercap_plot.c) and another being in [rapl](https://github.com/icl-utk-edu/papi/blob/master/src/components/rapl/utils/rapl_plot.c). I know this is not the nvml component, but I did want to share.

Treece-Burgess (changes requested):

src/high-level/papi_hl.c
mpatrou: Exactly. This number: 3472328295419250878 does not look like a valid number. For example gpu_memory_avg_power represents power in mW. Something like that: 18198 mW or 35295 mW, makes more sense to be reported.

src/high-level/papi_hl.c
mpatrou: This API, supports region-based calculations (before and after region execution). From what I saw in case of instantaneous values only one is reported. This might not make sense for different types of counters, no matter how large or small a region is. Since both values (before and after) are measured and can be different, considering their average can give a more detailed information and more accurate information, especially for instantaneous counters that change quite fast.

Thanks for the tools! I am also aware of them.

src/high-level/papi_hl.c
mpatrou: I updated that block.The recommended change is not quite equivalent. Non- NVML counters will be updated, too, if we do the above.

mpatrou: thank you! done!

src/high-level/papi_hl.c
mpatrou: thank you! done.

src/high-level/papi_hl.c
mpatrou: thank you! done.

src/high-level/papi_hl.c
mpatrou: thank you! done.

src/high-level/papi_hl.c
mpatrou: done

src/high-level/papi_hl.c
mpatrou: done

src/high-level/papi_hl.c
mpatrou: done

src/high-level/papi_hl.c
mpatrou: done

src/high-level/papi_hl.c
mpatrou: done

src/high-level/papi_hl.c
mpatrou: thank you. I included gpu_memory_avg_power for region-based average, too, for higher accuracy.

Treece-Burgess:
> Hello @Treece-Burgess. Do we have any news on this?

Hello @mpatrou I will do a final run through of the PR either tomorrow or this weekend.

src/high-level/papi_hl.c
Treece-Burgess: The issue with this change is that the last component will have two reads occur for it which are:
1. `PAPI_read`
2. `PAPI_read_ts`

The latter will overwrite the first `PAPI_read` value.  When a user calls `PAPI_hl_region_begin`, `PAPI_hl_read`, or `PAPI_hl_region_end` they would expect a single read to occur similar to if they called the low level API with `PAPI_read`. Therefore, this change needs to be reverted back to what it is currently in the master branch.

The reason for this extremely large value is because of the internal code introduced in PR #525 specifically the following [block](https://github.com/icl-utk-edu/papi/pull/525/changes#diff-976b5c1b311e59fe6326d329f33d6f5df7b720cd6de922d9553c98e3d2b40c66R341). The value returned is a `field_value->value.ullVal`; however, for a `fieldId` of `NVML_FI_DEV_POWER_INSTANT` you need to use a `valueType` of `uiVal`.

I have tested the code in PR #564 with this current PR and I have not seen an extremely large value output after 10 runs.

mpatrou: I see. changes reverted! thank you!!

Treece-Burgess: Looks good to me!

mpatrou:
example usage on GH200 with kokkos:
image

Treece-Burgess:
> example usage on GH200 with kokkos: image

I have access to a GH200, but it appears I did not test on that machine for PR 564. Thanks for sending the output to me, I concur that it works on that architecture.

Issue #524, opened 2025-12-19
[Issue]: rocp_sdk component test 'advanced' fails


dbarry9:
### Issue Description

When using ROCm 7.1.1 on the MI300A architecture, the `rocp_sdk` component test `advanced` fails with the following error message. This test passes when using ROCm 6.4.0 on the same device.

> W20251219 11:23:12.645857 140519483372672 ioctl.cpp:68] Device 57270 could not be locked for profiling due to lack of permissions (capability SYS_PERFMON). PMC Counters may be inaccurate and System Counter Collection will be degraded.
> W20251219 11:23:12.645893 140519483372672 ioctl.cpp:68] Device 57271 could not be locked for profiling due to lack of permissions (capability SYS_PERFMON). PMC Counters may be inaccurate and System Counter Collection will be degraded.
> W20251219 11:23:12.645898 140519483372672 ioctl.cpp:68] Device 57272 could not be locked for profiling due to lack of permissions (capability SYS_PERFMON). PMC Counters may be inaccurate and System Counter Collection will be degraded.
> W20251219 11:23:12.645908 140519483372672 ioctl.cpp:68] Device 57273 could not be locked for profiling due to lack of permissions (capability SYS_PERFMON). PMC Counters may be inaccurate and System Counter Collection will be degraded.
> FAILED!!!
> Line # 21 Error in PAPI_add_named_event: Invalid argument
> Some tests require special hardware, permissions, OS, compilers
> or library versions. PAPI may still function perfectly on your
> system without the particular feature being tested here.

Note the warnings printed here match those in Issue #507 and are unrelated to the component test failure. I ran this same test on the MI250X architecture on Frontier with ROCm 7.0.1 and did not encounter the failure, despite the warnings still being present on Frontier.

### Operating System

RHEL 8.10 (Ootpa)

### Architecture

x86_64

### CPU Architecture

AMD Instinct MI300A Accelerator

### Compiler

GCC 14.0.1

### PAPI Version

master

### Component

rocp_sdk

### Steps to Reproduce

1. ./configure --prefix=${INSTDIR} --with-components="rocp_sdk"
2. make && make install
3. ./components/rocp_sdk/tests/advanced

PR #530, opened 2026-01-08
rocp_sdk tests: Add a runscript which can toggle desired events to be used and allow each test to be ran without command line args

Treece-Burgess:
## Pull Request Description
This PR updates the `rocp_sdk` component tests to:
1. Take optional command line arguments to pass `rocp_sdk` native events
2. If no optional `rocp_sdk` native events are provided via the command line we internally will enumerate through the available events on the architecture and ROCm version to get one

Along with the above two updates, the PR adds a run script which will run all four of the current tests for the `rocp_sdk` component. An optional flag `--with-desired-events` can be provided to this run script. Doing so will see the component tests ran with desired events listed in `run_rocp_sdk_tests.sh`.

Note: The desired events that had the `DIMENSION_SHADER_ENGINE` qualifier appended with a value greater than 3 were removed. As from testing with ROCm versions 7.0.1 and up these values did not exist anymore. 

This partially closes Issue #524 as one other PR to update error codes is needed.

### Setup
Testing was done on Odyssey at Oregon with the following setup:
- OS: RHEL 8.10
- CPU: MI300A
- GPU: MI300A
- ROCm versions: 6.4.4, 7.1.1, and 7.2.0 (pre-release)
- PAPI Configure: `./configure --prefix=$PWD/test-install --with-components="rocp_sdk" --with-debug=yes`

### Results
- `advanced`, `simple`, `simple_sampling`, and `two_eventsets` all passed without providing event names on the command line*
- `advanced`, `simple`, `simple_sampling`, and `two_eventsets` all passed when providing event names on the command line*

__*__ - For ROCm 7.2.0 `two_eventsets.c` shows zero values for events which previously did not have zero values this is a separate issue and will be looked into)

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/components/rocp_sdk/tests/run_rocp_sdk_tests.sh
+++ b/src/components/rocp_sdk/tests/run_rocp_sdk_tests.sh
@@ -1,0 +1,96 @@
+#!/bin/bash
+#Note: It is possible that the listed desired events for each test 
+# may not exist on the architecture you are running on or with the 
+# ROCm version you are using. If you run with the option --with-desired-events
+# and the test fails, verify with utils/papi_native_avail the desired event exists with
+# the appended qualifiers.
+
+absolute_path_to_script_dir=$(cd "$(dirname "$0")" && pwd)
+cd "$absolute_path_to_script_dir"
+
+advanced_desired_events="rocp_sdk:::SQ_CYCLES:device=0,\
+rocp_sdk:::SQ_BUSY_CYCLES:DIMENSION_INSTANCE=0:DIMENSION_SHADER_ENGINE=0:device=0,\
+rocp_sdk:::SQ_BUSY_CYCLES:DIMENSION_INSTANCE=0:DIMENSION_SHADER_ENGINE=1:device=0,\
+rocp_sdk:::SQ_BUSY_CYCLES:DIMENSION_INSTANCE=0:DIMENSION_SHADER_ENGINE=2:device=0,\
+rocp_sdk:::SQ_BUSY_CYCLES:DIMENSION_INSTANCE=0:DIMENSION_SHADER_ENGINE=3:device=0,\
+rocp_sdk:::SQ_BUSY_CYCLES:DIMENSION_INSTANCE=0,\
+rocp_sdk:::SQ_WAVE_CYCLES:DIMENSION_SHADER_ENGINE=0:device=0,\
+rocp_sdk:::SQ_WAVE_CYCLES:DIMENSION_SHADER_ENGINE=1:device=0,\
+rocp_sdk:::SQ_WAVE_CYCLES:DIMENSION_SHADER_ENGINE=2:device=0,\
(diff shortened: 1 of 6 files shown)


djwoun:
I am reviewing this PR.

src/components/rocp_sdk/tests/simple_sampling.c
```diff
@@ -1,89 +1,142 @@
...
     if (papi_errno != PAPI_VER_CURRENT) {
         test_fail(__FILE__, __LINE__, "PAPI_library_init", papi_errno);
     }
 
+    // If a user does not provide an event or events, then we go get an event to add
+    if (total_event_count == 0) {
+        enumerate_and_store_rocp_sdk_native_events(&rocp_sdk_native_event_names, &total_event_count);
+    }
+
+    int eventset = PAPI_NULL;
```
djwoun: eventset is already a global used by thread_main(). Redeclaring it here creates a separate local EventSet, so the thread still reads the global PAPI_NULL handle at line 98 and PAPI_read() fails. Either use the global here or pass the created EventSet into the thread. The thread-side PAPI_read() error is also unchecked right now, which makes this miss.

src/components/rocp_sdk/tests/run_rocp_sdk_tests.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
```
djwoun, line 1: The script assumes it’s being run from src/components/rocp_sdk/tests. The make ... and ./... calls are all relative to the current directory, so running it from anywhere else will break. Can we grab the script’s directory at the top and cd there first?

```
script_dir="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)"
cd "$script_dir"
```

src/components/rocp_sdk/tests/simple_sampling.c
Treece-Burgess: Good catch. I am opting for using the existing global `eventset` var and the next push will reflect this change.

Though I opt to use the global `eventset` var `PAPI_read` returns `PAPI_ECMP` for samples 1 - 13 (ran on Odyssey with MI300A with ROCm 7.2.0). I experience this same behavior on the master branch as well. I will make a note to look into this further and address it in a separate PR if needed.

src/components/rocp_sdk/tests/run_rocp_sdk_tests.sh
Treece-Burgess: Good catch. With the most recent push, I now have a variable `absolute_path_to_script_dir` which is set equal to `$(cd "$(dirname "$0")" && pwd)` and then I `cd`. From my testing the following cases hold:

- Running from the `papi` dir
- Running from `papi/src` dir
- Running from `components/rocp_sdk/tests` dir

PR #535, opened 2026-01-14
amd_smi: PAPI_ENOSUPP as warning in EXIT_WARNING_ON_ADD

djwoun:
## Pull Request Description
Add PAPI_ENOSUPP to EXIT_WARNING_ON_ADD so tests explicitly report unsupported event-add cases (e.g., amd_smi:::power_cap) instead of silently skipping.

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/components/amd_smi/tests/test_harness.h
+++ b/src/components/amd_smi/tests/test_harness.h
@@ -217,12 +217,11 @@ static inline int harness_eval_result(const char *file, int line, HarnessOpts op
  * @brief If adding the event set fails due to unsupported or hardware/resource
  *        limits, exit as "PASSED with WARNING".
  *
- * Recognizes PAPI_ENOEVNT, PAPI_ECNFLCT, and PAPI_EPERM.
+ * Recognizes PAPI_ENOEVNT, PAPI_ECNFLCT, PAPI_EPERM, and PAPI_ENOSUPP.
  */
 #define EXIT_WARNING_ON_ADD(rc, evname) do { \
-    if ((rc) == PAPI_ENOEVNT || (rc) == PAPI_ECNFLCT || (rc) == PAPI_EPERM) { \
-        EXIT_WARNING("Event unavailable (%s): %s", \
-            ((rc) == PAPI_ENOEVNT ? "ENOEVNT" : (rc) == PAPI_ECNFLCT ? "ECNFLCT" : "EPERM"), (evname)); \
+    if ((rc) == PAPI_ENOEVNT || (rc) == PAPI_ECNFLCT || (rc) == PAPI_EPERM || (rc) == PAPI_ENOSUPP) { \
+        EXIT_WARNING("Event unavailable (%s): %s", PAPI_strerror(rc), (evname)); \
     } \
 } while (0)
 


src/components/amd_smi/tests/test_harness.h
```diff
@@ -220,7 +220,7 @@ static inline int harness_eval_result(const char *file, int line, HarnessOpts op
  * Recognizes PAPI_ENOEVNT, PAPI_ECNFLCT, and PAPI_EPERM.
  */
 #define EXIT_WARNING_ON_ADD(rc, evname) do { \
-    if ((rc) == PAPI_ENOEVNT || (rc) == PAPI_ECNFLCT || (rc) == PAPI_EPERM) { \
+    if ((rc) == PAPI_ENOEVNT || (rc) == PAPI_ECNFLCT || (rc) == PAPI_EPERM || (rc) == PAPI_ENOSUPP) { \
         EXIT_WARNING("Event unavailable (%s): %s", \
             ((rc) == PAPI_ENOEVNT ? "ENOEVNT" : (rc) == PAPI_ECNFLCT ? "ECNFLCT" : "EPERM"), (evname)); \
```
Treece-Burgess: When I run `sudo ./amdsmi_set_test` on Odyssey with ROCm 7.2.0 I do now get output that I am unable to add the power cap event:
```
[tburgess@odyssey tests]$ sudo ./amdsmi_set_test 
[sudo] password for tburgess: 
WARNING: Event unavailable (EPERM): amd_smi:::power_cap
```

When testing the error I would get is -18 or `PAPI_ENOSUPP`. Right now the message indicates it is `PAPI_EPERM`. Therefore, the ternary operator needs to be updated to account for `PAPI_ENOSUPP`.

Also, instead of mapping the PAPI error codes to a string equivalent with the ternary operator, why not use `PAPI_strerror()`? As an example:
```
#define EXIT_WARNING_ON_ADD(rc, evname) do { \
    if ((rc) == PAPI_ENOEVNT || (rc) == PAPI_ECNFLCT || (rc) == PAPI_EPERM || (rc) == PAPI_ENOSUPP) { \
        const char * str = PAPI_strerror(rc); \
        EXIT_WARNING("Event unavailable (%s): %s", \
            str, (evname)); \
    } \
} while (0)
```
For `PAPI_ENOSUPP` you would get back "Not supported" i.e. `WARNING: Event unavailable (Not supported): amd_smi:::power_cap`

src/components/amd_smi/tests/test_harness.h
```diff
@@ -220,7 +220,7 @@ static inline int harness_eval_result(const char *file, int line, HarnessOpts op
  * Recognizes PAPI_ENOEVNT, PAPI_ECNFLCT, and PAPI_EPERM.
  */
 #define EXIT_WARNING_ON_ADD(rc, evname) do { \
-    if ((rc) == PAPI_ENOEVNT || (rc) == PAPI_ECNFLCT || (rc) == PAPI_EPERM) { \
+    if ((rc) == PAPI_ENOEVNT || (rc) == PAPI_ECNFLCT || (rc) == PAPI_EPERM || (rc) == PAPI_ENOSUPP) { \
```
Treece-Burgess, line 223: Q: For the conditions, why check for specific PAPI error codes?

djwoun: Those codes represent expected “event unavailable / cannot conduct on this system” conditions (no such event, conflict, permission, not supported) that should be treated as PASS-with-warning. Other error codes may indicate real bugs.

src/components/amd_smi/tests/test_harness.h
djwoun: Good catch, the ternary was still hardcoded, so PAPI_ENOSUPP could be misreported as EPERM. I switched EXIT_WARNING_ON_ADD to PAPI_strerror(rc), so amd_smi:::power_cap now reports Not supported when it returns PAPI_ENOSUPP.

src/components/amd_smi/tests/test_harness.h
Treece-Burgess: I understand, thank you for the clarification.

Treece-Burgess (approved):
I tested this PR on Odyssey at Oregon with ROCm 7.2.0 pre-release. 

PAPI Utilities:
- `papi_component_avail`: ✅ 
- `papi_command_line`: ✅ 
- `papi_native_avail`: ✅ 

`amd_smi` tests:
- `amdsmi_basics`: ✅ 
- `amdsmi_energy_monotonic`: ✅ 
- `amdsmi_hello`: ✅ 
- `amdsmi_gemm`: ✅ 
- `amdsmi_set_test`: Now shows:
```
[tburgess@odyssey tests]$ sudo ./amdsmi_set_test 
[sudo] password for tburgess: 
WARNING: Event unavailable (Not supported): amd_smi:::power_cap
PASSED with WARNING
```

PR #538, opened 2026-01-20
2026.01.13 cat bf16 kernels

dbarry9:
## Pull Request Description

This pull request:
- revises FLOPs kernels to accounts for speculative FLOPs and
- introduces kernels that use dot product intrinsics to trigger BF16 operations.

These changes have been tested on the ARM Neoverse V2 and Intel Sapphire Rapids architectures.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**


src/counter_analysis_toolkit/Makefile
```diff
@@ -1,7 +1,7 @@
-PAPIDIR?=${PAPI_DIR}
+PAPI_DIR?=${PAPI_DIR}
```
Treece-Burgess: Q: Why not just use `PAPI_DIR` set by the user instead of creating a new variable in the Makefile?

src/counter_analysis_toolkit/cat_arch.h
```diff
@@ -3,174 +3,284 @@
...
+void  test_sp_arm_VEC( int instr_per_loop, int EventSet, FILE *fp );
+void  test_dp_arm_VEC( int instr_per_loop, int EventSet, FILE *fp );
+void  test_fp16_arm_VEC_FMA( int instr_per_loop, int EventSet, FILE *fp );
+void  test_sp_arm_VEC_FMA( int instr_per_loop, int EventSet, FILE *fp );
+void  test_dp_arm_VEC_FMA( int instr_per_loop, int EventSet, FILE *fp );
+void  test_bf16_arm_VEC_DOT( int EventSet, FILE *fp );
+
+/* There are two vector instruction sets available on this architecture: SVE and NEON. */
+#define CAT_DEV_SVE
+//#define CAT_DEV_NEON
```
Treece-Burgess, line 151: Q: This define is used in line 186 below, should this be uncommented?

src/counter_analysis_toolkit/cat_arch.h
```diff
@@ -3,174 +3,284 @@
...
-#define SET_VEC_SH(_I_)         _I_ ;
-#define ADD_VEC_SH(_I_,_J_)     vaddh_f16( _I_ , _J_ );
-#define MUL_VEC_SH(_I_,_J_)     vmulh_f16( _I_ , _J_ );
-#define SQRT_VEC_SH(_I_)        vsqrth_f16( _I_ );
-#define FMA_VEC_SH(_out_,_I_,_J_,_K_) _out_ = _I_ * _J_ + _K_;
-
-#define SET_VEC_SS(_I_)         _I_ ;
-#define ADD_VEC_SS(_I_,_J_)     _I_ + _J_ ;
-#define MUL_VEC_SS(_I_,_J_)     _I_ * _J_ ;
+#include <arm_fp16.h>
```
Treece-Burgess, line 219: Q: Should this be underneath a `#if defined`?

src/counter_analysis_toolkit/Makefile
```diff
@@ -31,42 +31,38 @@ ifndef ARCH
...
-    VEC512=-mavx512f -O0 -DX86 $(VEC_META) -DX86_VEC_WIDTH_512B
-    VEC128_FMA=-mfma4 -mfma -O0 -DX86 $(VEC_META) -DX86_VEC_WIDTH_128B
-    VEC256_FMA=-mfma4 -mfma -O0 -DX86 $(VEC_META) -DX86_VEC_WIDTH_256B
-    VEC512_FMA=-mfma4 -mfma -mavx512f -O0 -DX86 $(VEC_META) -DX86_VEC_WIDTH_512B
-    VEC=-mavx -O0 -DX86
-    VEC_FMA=-mfma4 -mfma -O0 -DX86
-    VEC_ALL=$(VEC) $(VEC_FMA) -O0 -DX86
+    VECSRC=vec_dot_bf16 vec_fma_fp16 vec_fma_sp vec_fma_dp vec_nonfma_fp16 vec_nonfma_sp vec_nonfma_dp
+    VEC_META=-DAVX128_AVAIL -DAVX256_AVAIL -DAVX512_AVAIL -DAVX512_FP16_AVAIL -DAVX512_BF16_AVAIL
+    VEC128=-mavx -mavx512vl -mavx512fp16 -mavx512bf16 -O0 -DX86 $(VEC_META) -DX86_VEC_WIDTH_128B
```
Treece-Burgess, line 38: Q: Is it a guarantee that if you have fp16 available then you will have bf16 available and vice versa?

src/counter_analysis_toolkit/cat_arch.h
```diff
@@ -3,174 +3,284 @@
...
+
+void test_bf16_x86_128B_VEC_DOT( int EventSet, FILE *fp );
+void test_bf16_x86_256B_VEC_DOT( int EventSet, FILE *fp );
+void test_bf16_x86_512B_VEC_DOT( int EventSet, FILE *fp );
 
 #include <immintrin.h>
 
-typedef __m128  SP_SCALAR_TYPE;
-typedef __m128d DP_SCALAR_TYPE;
+#if defined(AVX512_BF16_AVAIL)
```
Treece-Burgess, line 36: Q: It appears that for a defined architecture e.g. x86 you have multiple `#if defined`'s for AVX512_BF16_AVAIL and AVX512_FP16_AVAIL. Why take this approach and not place them all under one AVX512_BF16_AVAIL or AVX512_FP16_AVAIL?

src/counter_analysis_toolkit/flops.c
```diff
@@ -46,14 +46,14 @@ void exec_single_gemm( int EventSet, FILE *fp );
...
 
-#if defined(ARM)
-half normalize_half( int n, half *xh );
-void cholesky_half( int n, half *lh, half *ah );
-void exec_half_norm( int EventSet, FILE *fp );
-void exec_half_cholesky( int EventSet, FILE *fp );
-void exec_half_gemm( int EventSet, FILE *fp );
-void keep_half_vec_res( int n, half *xh );
-void keep_half_mat_res( int n, half *lh );
+#if defined(FP16_AVAIL) || defined(AVX512_FP16_AVAIL)
```
Treece-Burgess, line 49: Q: We enter this define only for the ARM and x86 case?

src/counter_analysis_toolkit/flops.c
```diff
@@ -46,14 +46,14 @@ void exec_single_gemm( int EventSet, FILE *fp );
...
+fp16_half normalize_fp16_half( int n, fp16_half *xh );
+void cholesky_fp16_half( int n, fp16_half *lh, fp16_half *ah );
+void exec_fp16_half_norm( int EventSet, FILE *fp );
```
Treece-Burgess, line 52: Q: Why not a separate header file to store the forward declarations? I.e. flops.h.

src/counter_analysis_toolkit/flops.c
```diff
@@ -643,23 +643,23 @@ void keep_single_mat_res( int n, float *ls ) {
     }
 }
 
-#if defined(ARM)
-void exec_half_norm( int EventSet, FILE *fp ) {
+#if defined(FP16_AVAIL) || defined(AVX512_FP16_AVAIL)
```
Treece-Burgess, line 646: Q: Same as a previous comment, but the following:
```
#if defined(FP16_AVAIL) || defined(AVX512_FP16_AVAIL)
```

Has multiple occurrences, why not just utilize one and pull all the functions inside it?

src/counter_analysis_toolkit/gen_seq_dlopen.sh
```diff
@@ -1,8 +1,13 @@
 #!/bin/bash
 
-DRV_F=icache_seq.c
-KRN_F=icache_seq_kernel.c
-HEAD_F=icache_seq.h
+if [ $# -lt 1 ]
```
Treece-Burgess, line 3: Q: Were you running into not finding `icache_seq.c`, `icache_seq_kernel.c`, and `icache_seq.h`? As to me this seems unnecessary if you are always just passing PWD via the Makefile.

src/counter_analysis_toolkit/vec.c
```diff
@@ -46,153 +46,179 @@ void vec_driver(char* papi_event_name, hw_desc_t *hw_desc, char* outdir)
...
-    fprintf(ofp_papi, "# HP Non-FMA Scalar\n");
-    test_hp_scalar_VEC_24( ITER, EventSet, ofp_papi );
-    test_hp_scalar_VEC_48( ITER, EventSet, ofp_papi );
-    test_hp_scalar_VEC_96( ITER, EventSet, ofp_papi );
-
-    fprintf(ofp_papi, "# HP Non-FMA Vector AVX128\n");
-    test_hp_x86_128B_VEC( 24, ITER, EventSet, ofp_papi );
-    test_hp_x86_128B_VEC( 48, ITER, EventSet, ofp_papi );
-    test_hp_x86_128B_VEC( 96, ITER, EventSet, ofp_papi );
+#if defined(AVX512_FP16_AVAIL)
```
Treece-Burgess, line 49: nit: I would either follow the same linear pattern you have done with previous `if defined`'s or restructure for readability.
```c
#if defined(AVX128_AVAIL)
    #if defined(AVX512_FP16_AVAIL)
        // FP16 Non-FMA instruction trials.
        fprintf(ofp_papi, "# FP16 Non-FMA Scalar\n");
        test_fp16_scalar_VEC_24( EventSet, ofp_papi );
        test_fp16_scalar_VEC_48( EventSet, ofp_papi );
        test_fp16_scalar_VEC_96( EventSet, ofp_papi );

        fprintf(ofp_papi, "# FP16 Non-FMA Vector AVX128\n");
        test_fp16_x86_128B_VEC( 24, EventSet, ofp_papi );
        test_fp16_x86_128B_VEC( 48, EventSet, ofp_papi );
        test_fp16_x86_128B_VEC( 96, EventSet, ofp_papi );
    #if defined(AVX256_AVAIL)
        fprintf(ofp_papi, "# FP16 Non-FMA Vector AVX256\n");
        test_fp16_x86_256B_VEC( 24, EventSet, ofp_papi );
        test_fp16_x86_256B_VEC( 48, EventSet, ofp_papi );
        test_fp16_x86_256B_VEC( 96, EventSet, ofp_papi );
        .
        .
        .
 ```

File: src/counter_analysis_toolkit/vec.c [hunk 11]
```diff
@@ -46,153 +46,179 @@ void vec_driver(char* papi_event_name, hw_desc_t *hw_desc, char* outdir)
...
+    // FP16 Non-FMA instruction trials.
+    fprintf(ofp_papi, "# FP16 Non-FMA Scalar\n");
```
Treece-Burgess, line 51: Q: ofp_papi is a file? I just want to make sure we are not printing out to the terminal.

src/counter_analysis_toolkit/vec.c
Treece-Burgess: There are also a few more instances of this below as well.

src/counter_analysis_toolkit/cat_arch.h
```diff
@@ -3,174 +3,284 @@
 typedef unsigned long long uint64;
 
 #if defined(X86)
-void test_hp_x86_128B_VEC( int instr_per_loop, uint64 iterations, int EventSet, FILE *fp );
```
Treece-Burgess, line 6: Q: What were the iterations used for previously?

src/counter_analysis_toolkit/vec_dot_bf16.c
```diff
@@ -0,0 +1,98 @@
...
+    return test_bf16_VEC_DOT( EventSet, fp );
+}
+#elif defined(ARM)
+void test_bf16_arm_VEC_DOT( int EventSet, FILE *fp ) {
+    return test_bf16_VEC_DOT( EventSet, fp );
+}
+#endif
+
+#if ( defined(BF16_AVAIL) && defined(CAT_DEV_SVE) ) || defined(AVX512_BF16_AVAIL)
+static
```
Treece-Burgess, line 28: Q: Is this static suppose to go with `float test_bf16_VEC_DOT_internal( int EventSet, FILE *fp )`?

src/counter_analysis_toolkit/vec_dot_bf16.c
```diff
@@ -0,0 +1,98 @@
...
+    #else /* Cannot happen due to ifdef guards. */
+    #endif
+    SP_VEC_TYPE result = SET_VEC_PS(0.0);
+    double values = 0.0;
+    long long iterValues[1]; iterValues[0] = 0;
+    for (int iter=0; iter<ITERS; ++iter) {
+
+    if ( PAPI_start( EventSet ) != PAPI_OK ) {
+        fprintf(stderr, "Problem.\n");
+        return -1;
```
Treece-Burgess, line 55: Missing PAPI error code.

src/counter_analysis_toolkit/vec_dot_bf16.c
```diff
@@ -0,0 +1,98 @@
...
+    }
+    #else /* Cannot happen due to ifdef guards. */
+    #endif
+    SP_VEC_TYPE result = SET_VEC_PS(0.0);
+    double values = 0.0;
+    long long iterValues[1]; iterValues[0] = 0;
+    for (int iter=0; iter<ITERS; ++iter) {
+
+    if ( PAPI_start( EventSet ) != PAPI_OK ) {
+        fprintf(stderr, "Problem.\n");
```
Treece-Burgess, line 54: Will print to the terminal, need a SUBDBG.

src/counter_analysis_toolkit/vec_dot_bf16.c
```diff
@@ -0,0 +1,98 @@
...
+#if ( defined(BF16_AVAIL) && defined(CAT_DEV_SVE) ) || defined(AVX512_BF16_AVAIL)
+static
+
+float test_bf16_VEC_DOT_internal( int EventSet, FILE *fp ){
+
+    #if defined(BF16_AVAIL) && defined(CAT_DEV_SVE)
+    bf16_half two = 2.0;
+    BF16_VEC_TYPE vec1 = SET_VEC_PBF16(two);
+    BF16_VEC_TYPE vec2 = SET_VEC_PBF16(two);
+    #elif defined(AVX512_BF16_AVAIL)
```
Treece-Burgess, line 35: Q: Could this just not be the else block based off your comment below?

src/counter_analysis_toolkit/vec_dot_bf16.c
```diff
@@ -0,0 +1,98 @@
...
+    bf16_half one = 1.0;
+    BF16_VEC_TYPE vec1;
+    BF16_VEC_TYPE vec2;
+    int i;
+    const int CAP = sizeof(BF16_VEC_TYPE)/sizeof(bf16_half);
+    for(i = 0; i < CAP; ++i) {
+        vec1[i] = one;
+        vec2[i] = one;
+    }
+    #else /* Cannot happen due to ifdef guards. */
```
Treece-Burgess, line 45: If you choose to keep the structure as is, I would suggest adding a SUBDBG message here and a PAPI error code returned. I understand that this should never occur, but if an anomaly occurs it would be better to error at this stage with a clear indication of why rather than continuing on to the next block of code.

src/counter_analysis_toolkit/vec_dot_bf16.c
```diff
@@ -0,0 +1,98 @@
...
+    for (int iter=0; iter<ITERS; ++iter) {
+
+    if ( PAPI_start( EventSet ) != PAPI_OK ) {
+        fprintf(stderr, "Problem.\n");
+        return -1;
+    }
+
+    result = DOT_VEC_PBF16(result, vec1, vec2);
+
+    usleep(1);
```
Treece-Burgess, line 60: Q: Is this sleep needed?

src/counter_analysis_toolkit/vec_dot_bf16.c
```diff
@@ -0,0 +1,98 @@
...
+    #if defined(BF16_AVAIL) && defined(CAT_DEV_SVE)
+    float CAP = 8.0;
+    #elif defined(AVX512_BF16_AVAIL)
+    float CAP = 2.0;
+    #else /* Cannot happen due to ifdef guards. */
+    #endif
+    float sum = test_bf16_VEC_DOT_internal( EventSet, fp );
+
+    if( sum != CAP ) {
+        fprintf(stderr, "BF16 DOT: Inconsistent FLOP results detected! %f vs %f\n", sum, CAP);
```
Treece-Burgess, line 95: Printing to the terminal, SUBDBG.

src/counter_analysis_toolkit/vec_dot_bf16.c
```diff
@@ -0,0 +1,98 @@
...
+}
+
+static
+void test_bf16_VEC_DOT( int EventSet, FILE *fp )
+{
+    #if defined(BF16_AVAIL) && defined(CAT_DEV_SVE)
+    float CAP = 8.0;
+    #elif defined(AVX512_BF16_AVAIL)
+    float CAP = 2.0;
+    #else /* Cannot happen due to ifdef guards. */
```
Treece-Burgess, line 90: Same as above.

src/counter_analysis_toolkit/vec_scalar_verify.h
```diff
@@ -3,46 +3,46 @@
 #include <stdlib.h>
 #include "cat_arch.h"
 
-#define ITER 1
+#define ITERS 10000
```
Treece-Burgess, line 6: Q: Are you wanting to keep this at 10,000 compared to 1?

src/counter_analysis_toolkit/vec_nonfma_sp.c
```diff
@@ -90,13 +92,20 @@ float test_sp_mac_VEC_24( uint64 iterations, int EventSet, FILE *fp ){
             rA = ADD_VEC_PS(rA,rD);
             rB = MUL_VEC_PS(rB,rC);
 
-            i++;
-        }
-        c++;
+    /* Stop PAPI counters */
+    if ( NULL != fp && PAPI_stop(EventSet, iterValues) != PAPI_OK ) {
```
Treece-Burgess, line 97: Q: Why not store and return the PAPI error code that is return from PAPI_stop if it fails?

src/counter_analysis_toolkit/vec_nonfma_sp.c
```diff
@@ -1,39 +1,46 @@
...
-float test_sp_mac_VEC_24( uint64 iterations, int EventSet, FILE *fp ){
-    register SP_VEC_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+float test_sp_mac_VEC_24( int EventSet, FILE *fp ){
+
+    #if defined(ARM) && defined(CAT_DEV_SVE)
+    svbool_t pg = svptrue_b32();
+    #endif
+    volatile SP_VEC_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+    double values = 0.0;
+    long long iterValues[1]; iterValues[0] = 0;
```
Treece-Burgess, line 42: Just as a suggestion, if all you are worried about is a single value instead of declaring an array of size 1. You could just declare long long iterValues = 0 and then pass `&iterValues` to `PAPI_stop`.

src/counter_analysis_toolkit/vec_nonfma_sp.c
```diff
@@ -1,39 +1,46 @@
...
+    for (int iter=0; iter<ITERS; ++iter) {
```
Treece-Burgess: Declare `iter` outside of the for loop.

src/counter_analysis_toolkit/vec_nonfma_sp.c
```diff
@@ -384,13 +404,20 @@ float test_sp_mac_VEC_96( uint64 iterations, int EventSet, FILE *fp ){
             rA = ADD_VEC_PS(rA,rD);
             rB = MUL_VEC_PS(rB,rC);
 
-            i++;
-        }
-        c++;
+    /* Stop PAPI counters */
+    if ( NULL != fp && PAPI_stop(EventSet, iterValues) != PAPI_OK ) {
```
Treece-Burgess, line 411: Same as above.

src/counter_analysis_toolkit/vec_nonfma_sp.c
```diff
@@ -248,8 +266,15 @@ float test_sp_mac_VEC_48( uint64 iterations, int EventSet, FILE *fp ){
...
-    register SP_VEC_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+float test_sp_mac_VEC_96( int EventSet, FILE *fp ){
+
+    #if defined(ARM) && defined(CAT_DEV_SVE)
+    svbool_t pg = svptrue_b32();
+    #endif
+    volatile SP_VEC_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+    double values = 0.0;
+    long long iterValues[1]; iterValues[0] = 0;
+    for (int iter=0; iter<ITERS; ++iter) {
```
Treece-Burgess: Declare `iter` outside of the for loop.

src/counter_analysis_toolkit/vec_nonfma_sp.c
```diff
@@ -211,13 +222,20 @@ float test_sp_mac_VEC_48( uint64 iterations, int EventSet, FILE *fp ){
             rA = ADD_VEC_PS(rA,rD);
             rB = MUL_VEC_PS(rB,rC);
 
-            i++;
-        }
-        c++;
+    /* Stop PAPI counters */
+    if ( NULL != fp && PAPI_stop(EventSet, iterValues) != PAPI_OK ) {
```
Treece-Burgess, line 228: Same as above.

src/counter_analysis_toolkit/vec_nonfma_sp.c
```diff
@@ -418,25 +445,25 @@ float test_sp_mac_VEC_96( uint64 iterations, int EventSet, FILE *fp ){
...
     else if ( instr_per_loop == 96 ) {
-        sum += test_sp_mac_VEC_96( iterations, EventSet, fp );
-        scalar_sum += test_sp_scalar_VEC_96( iterations, EventSet, NULL );
+        sum += test_sp_mac_VEC_96( EventSet, fp );
+        scalar_sum += test_sp_scalar_VEC_96( EventSet, NULL );
     }
 
     if( sum/4.0 != scalar_sum ) {
-        fprintf(stderr, "Inconsistent FLOP results detected!\n");
+        fprintf(stderr, "SP: Inconsistent FLOP results detected!\n");
```
Treece-Burgess, line 470: Instance of printing to the terminal.

src/counter_analysis_toolkit/vec_scalar_verify.c
```diff
@@ -5,404 +5,401 @@ void papi_stop_and_print_placeholder(long long theory, FILE *fp)
...
-#if defined(ARM)
-half test_hp_scalar_VEC_24( uint64 iterations, int EventSet, FILE *fp ){
-    register half r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+#if defined(FP16_AVAIL) || defined(AVX512_FP16_AVAIL)
+fp16_half test_fp16_scalar_VEC_24( int EventSet, FILE *fp ){
+
+    volatile FP16_SCALAR_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+    double values = 0.0;
+    long long iterValues[1]; iterValues[0] = 0;
+    for (int iter=0; iter<ITERS; ++iter) {
```
Treece-Burgess: Declare `iter` outside of the for loop.

src/counter_analysis_toolkit/vec_scalar_verify.c
```diff
@@ -5,404 +5,401 @@ void papi_stop_and_print_placeholder(long long theory, FILE *fp)
...
+    rD = SET_VEC_SFP16(0.14);
+    rE = SET_VEC_SFP16(0.15);
+    rF = SET_VEC_SFP16(0.16);
 
     /* Start PAPI counters */
-    if ( NULL != fp ) {
-      if ( PAPI_start( EventSet ) != PAPI_OK ) {
-        return -1;
-      }
+    if ( NULL != fp && PAPI_start( EventSet ) != PAPI_OK ) {
```
Treece-Burgess, line 41: Q: Why not store and then return the PAPI error code of `PAPI_start` if a failure occurs?

src/counter_analysis_toolkit/vec_scalar_verify.c
```diff
@@ -5,404 +5,401 @@ void papi_stop_and_print_placeholder(long long theory, FILE *fp)
...
+    r5 = MUL_VEC_SFP16(r5,rE);
+    r6 = ADD_VEC_SFP16(r6,rD);
+    r7 = MUL_VEC_SFP16(r7,rC);
+    r8 = ADD_VEC_SFP16(r8,rF);
+    r9 = MUL_VEC_SFP16(r9,rE);
+    rA = ADD_VEC_SFP16(rA,rD);
+    rB = MUL_VEC_SFP16(rB,rC);
 
     /* Stop PAPI counters */
+    if ( NULL != fp && PAPI_stop(EventSet, iterValues) != PAPI_OK ) {
```
Treece-Burgess, line 73: Likewise as above.

src/counter_analysis_toolkit/vec_scalar_verify.c
```diff
@@ -5,404 +5,401 @@ void papi_stop_and_print_placeholder(long long theory, FILE *fp)
...
 }
 
-half test_hp_scalar_VEC_48( uint64 iterations, int EventSet, FILE *fp ){
-    register half r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+fp16_half test_fp16_scalar_VEC_48( int EventSet, FILE *fp ){
+
+    volatile FP16_SCALAR_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+    double values = 0.0;
+    long long iterValues[1]; iterValues[0] = 0;
+    for (int iter=0; iter<ITERS; ++iter) {
```
Treece-Burgess: Declare `iter` outside of the for loop.

src/counter_analysis_toolkit/vec_scalar_verify.c
```diff
@@ -5,404 +5,401 @@ void papi_stop_and_print_placeholder(long long theory, FILE *fp)
...
+    rD = SET_VEC_SFP16(0.14);
+    rE = SET_VEC_SFP16(0.15);
+    rF = SET_VEC_SFP16(0.16);
 
     /* Start PAPI counters */
-    if ( NULL != fp ) {
-      if ( PAPI_start( EventSet ) != PAPI_OK ) {
-        return -1;
-      }
+    if ( NULL != fp && PAPI_start( EventSet ) != PAPI_OK ) {
```
Treece-Burgess, line 136: Likewise as above.

src/counter_analysis_toolkit/vec_scalar_verify.c
```diff
@@ -5,404 +5,401 @@ void papi_stop_and_print_placeholder(long long theory, FILE *fp)
...
+    r5 = MUL_VEC_SFP16(r5,rE);
+    r6 = ADD_VEC_SFP16(r6,rD);
+    r7 = MUL_VEC_SFP16(r7,rC);
+    r8 = ADD_VEC_SFP16(r8,rF);
+    r9 = MUL_VEC_SFP16(r9,rE);
+    rA = ADD_VEC_SFP16(rA,rD);
+    rB = MUL_VEC_SFP16(rB,rC);
 
     /* Stop PAPI counters */
+    if ( NULL != fp && PAPI_stop(EventSet, iterValues) != PAPI_OK ) {
```
Treece-Burgess, line 194: Likewise as above.

src/counter_analysis_toolkit/vec_scalar_verify.c
```diff
@@ -5,404 +5,401 @@ void papi_stop_and_print_placeholder(long long theory, FILE *fp)
...
 }
 
-half test_hp_scalar_VEC_96( uint64 iterations, int EventSet, FILE *fp ){
-    register half r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+fp16_half test_fp16_scalar_VEC_96( int EventSet, FILE *fp ){
+
+    volatile FP16_SCALAR_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+    double values = 0.0;
+    long long iterValues[1]; iterValues[0] = 0;
+    for (int iter=0; iter<ITERS; ++iter) {
```
Treece-Burgess: Declare `iter` outside of the for loop.

src/counter_analysis_toolkit/vec_scalar_verify.c
```diff
@@ -5,404 +5,401 @@ void papi_stop_and_print_placeholder(long long theory, FILE *fp)
...
+    r5 = MUL_VEC_SFP16(r5,rE);
+    r6 = ADD_VEC_SFP16(r6,rD);
+    r7 = MUL_VEC_SFP16(r7,rC);
+    r8 = ADD_VEC_SFP16(r8,rF);
+    r9 = MUL_VEC_SFP16(r9,rE);
+    rA = ADD_VEC_SFP16(rA,rD);
+    rB = MUL_VEC_SFP16(rB,rC);
 
     /* Stop PAPI counters */
+    if ( NULL != fp && PAPI_stop(EventSet, iterValues) != PAPI_OK ) {
```
Treece-Burgess, line 367: Likewise as above.

src/counter_analysis_toolkit/vec_scalar_verify.c
```diff
@@ -440,8 +435,12 @@ float test_hp_scalar_VEC_96( uint64 iterations, int EventSet, FILE *fp ){
...
 /* Loop unrolling:  24 instructions */
 /************************************/
-float test_sp_scalar_VEC_24( uint64 iterations, int EventSet, FILE *fp ){
-    register SP_SCALAR_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+float test_sp_scalar_VEC_24( int EventSet, FILE *fp ){
+
+    volatile SP_SCALAR_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+    double values = 0.0;
+    long long iterValues[1]; iterValues[0] = 0;
+    for (int iter=0; iter<ITERS; ++iter) {
```
Treece-Burgess: Declare `iter` outside of the for loop.

src/counter_analysis_toolkit/vec_scalar_verify.c
```diff
@@ -462,52 +461,50 @@ float test_sp_scalar_VEC_24( uint64 iterations, int EventSet, FILE *fp ){
     rF = SET_VEC_SS(0.16);
 
     /* Start PAPI counters */
-    if ( NULL != fp ) {
-      if ( PAPI_start( EventSet ) != PAPI_OK ) {
-        return -1;
-      }
+    if ( NULL != fp && PAPI_start( EventSet ) != PAPI_OK ) {
```
Treece-Burgess, line 468: Likewise as above.

src/counter_analysis_toolkit/vec_nonfma_dp.c
```diff
@@ -380,13 +400,20 @@ double test_dp_mac_VEC_96( uint64 iterations, int EventSet, FILE *fp ){
...
-    /* Stop PAPI counters */
-    papi_stop_and_print(96, EventSet, fp);
+    values += iterValues[0];
+
+} // end of ITERS
+
+    values /= ITERS;
+
+    if ( NULL != fp ) {
+      papi_print(96, fp, values);
```
Treece-Burgess, line 418: Q: Does this correspond to the instructions per loop?

nit: I would suggest possibly not hard coding this value and have a variable which stores it. There are a few other cases as well if you decide to take that approach.

src/counter_analysis_toolkit/vec_fma_sp.c
```diff
@@ -1,39 +1,46 @@
...
-    register SP_VEC_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+float test_sp_mac_VEC_FMA_12( int EventSet, FILE *fp ){
+
+    #if defined(ARM) && defined(CAT_DEV_SVE)
+    svbool_t pg = svptrue_b32();
+    #endif
+    volatile SP_VEC_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+    double values = 0.0;
+    long long iterValues[1]; iterValues[0] = 0;
+    for (int iter=0; iter<ITERS; ++iter) {
```
Treece-Burgess: Declare `iter` outside of for loop.

src/counter_analysis_toolkit/vec_fma_sp.c
```diff
@@ -110,8 +119,15 @@ float test_sp_mac_VEC_FMA_12( uint64 iterations, int EventSet, FILE *fp ){
...
-    register SP_VEC_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+float test_sp_mac_VEC_FMA_24( int EventSet, FILE *fp ){
+
+    #if defined(ARM) && defined(CAT_DEV_SVE)
+    svbool_t pg = svptrue_b32();
+    #endif
+    volatile SP_VEC_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+    double values = 0.0;
+    long long iterValues[1]; iterValues[0] = 0;
+    for (int iter=0; iter<ITERS; ++iter) {
```
Treece-Burgess: Declare `iter` outside of for loop.

src/counter_analysis_toolkit/vec_fma_sp.c
```diff
@@ -202,8 +220,15 @@ float test_sp_mac_VEC_FMA_24( uint64 iterations, int EventSet, FILE *fp ){
...
-    register SP_VEC_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+float test_sp_mac_VEC_FMA_48( int EventSet, FILE *fp ){
+
+    #if defined(ARM) && defined(CAT_DEV_SVE)
+    svbool_t pg = svptrue_b32();
+    #endif
+    volatile SP_VEC_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+    double values = 0.0;
+    long long iterValues[1]; iterValues[0] = 0;
+    for (int iter=0; iter<ITERS; ++iter) {
```
Treece-Burgess: Declare `iter` outside of for loop.

src/counter_analysis_toolkit/vec_fma_sp.c
```diff
@@ -78,13 +80,20 @@ float test_sp_mac_VEC_FMA_12( uint64 iterations, int EventSet, FILE *fp ){
             r4 = FMA_VEC_PS(r4,r9,rB);
             r5 = FMA_VEC_PS(r5,r8,rA);
 
-            i++;
-        }
-        c++;
+    /* Stop PAPI counters */
+    if ( NULL != fp && PAPI_stop(EventSet, iterValues) != PAPI_OK ) {
```
Treece-Burgess, line 85: Same as previous comments.

src/counter_analysis_toolkit/vec_fma_sp.c
```diff
@@ -170,13 +181,20 @@ float test_sp_mac_VEC_FMA_24( uint64 iterations, int EventSet, FILE *fp ){
...
```
Treece-Burgess, line 187: Same as previous comments.

src/counter_analysis_toolkit/vec_fma_sp.c
```diff
@@ -290,13 +310,20 @@ float test_sp_mac_VEC_FMA_48( uint64 iterations, int EventSet, FILE *fp ){
...
```
Treece-Burgess, line 317: Same as previous comments.

src/counter_analysis_toolkit/vec_nonfma_dp.c
```diff
@@ -1,39 +1,46 @@
...
-    register DP_VEC_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+double test_dp_mac_VEC_24( int EventSet, FILE *fp ){
+
+    #if defined(ARM) && defined(CAT_DEV_SVE)
+    svbool_t pg = svptrue_b64();
+    #endif
+    volatile DP_VEC_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+    double values = 0.0;
+    long long iterValues[1]; iterValues[0] = 0;
+    for (int iter=0; iter<ITERS; ++iter) {
```
Treece-Burgess: Declare `iter` outside of for loop.

src/counter_analysis_toolkit/vec_nonfma_dp.c
```diff
@@ -90,13 +92,20 @@ double test_dp_mac_VEC_24( uint64 iterations, int EventSet, FILE *fp ){
             rA = ADD_VEC_PD(rA,rD);
             rB = MUL_VEC_PD(rB,rC);
 
-            i++;
-        }
-        c++;
+    /* Stop PAPI counters */
+    if ( NULL != fp && PAPI_stop(EventSet, iterValues) != PAPI_OK ) {
```
Treece-Burgess, line 97: Same as previous comments.

src/counter_analysis_toolkit/vec_nonfma_dp.c
```diff
@@ -125,8 +134,15 @@ double test_dp_mac_VEC_24( uint64 iterations, int EventSet, FILE *fp ){
...
-    register DP_VEC_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+double test_dp_mac_VEC_48( int EventSet, FILE *fp ){
+
+    #if defined(ARM) && defined(CAT_DEV_SVE)
+    svbool_t pg = svptrue_b64();
+    #endif
+    volatile DP_VEC_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+    double values = 0.0;
+    long long iterValues[1]; iterValues[0] = 0;
+    for (int iter=0; iter<ITERS; ++iter) {
```
Treece-Burgess: Declare `iter` outside of for loop.

src/counter_analysis_toolkit/vec_nonfma_dp.c
```diff
@@ -209,13 +220,20 @@ double test_dp_mac_VEC_48( uint64 iterations, int EventSet, FILE *fp ){
             rA = ADD_VEC_PD(rA,rD);
             rB = MUL_VEC_PD(rB,rC);
 
-            i++;
-        }
-        c++;
+    /* Stop PAPI counters */
+    if ( NULL != fp && PAPI_stop(EventSet, iterValues) != PAPI_OK ) {
```
Treece-Burgess, line 226: Same as previous comments.

src/counter_analysis_toolkit/vec_nonfma_dp.c
```diff
@@ -244,8 +262,15 @@ double test_dp_mac_VEC_48( uint64 iterations, int EventSet, FILE *fp ){
...
-    register DP_VEC_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+double test_dp_mac_VEC_96( int EventSet, FILE *fp ){
+
+    #if defined(ARM) && defined(CAT_DEV_SVE)
+    svbool_t pg = svptrue_b64();
+    #endif
+    volatile DP_VEC_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+    double values = 0.0;
+    long long iterValues[1]; iterValues[0] = 0;
+    for (int iter=0; iter<ITERS; ++iter) {
```
Treece-Burgess: Declare `iter` outside of for loop.

src/counter_analysis_toolkit/vec_nonfma_dp.c
```diff
@@ -380,13 +400,20 @@ double test_dp_mac_VEC_96( uint64 iterations, int EventSet, FILE *fp ){
             rA = ADD_VEC_PD(rA,rD);
             rB = MUL_VEC_PD(rB,rC);
 
-            i++;
-        }
-        c++;
+    /* Stop PAPI counters */
+    if ( NULL != fp && PAPI_stop(EventSet, iterValues) != PAPI_OK ) {
```
Treece-Burgess, line 407: Same as previous comments.

src/counter_analysis_toolkit/vec_scalar_verify.c
```diff
@@ -462,52 +461,50 @@ float test_sp_scalar_VEC_24( uint64 iterations, int EventSet, FILE *fp ){
...
+    r5 = MUL_VEC_SS(r5,rE);
+    r6 = ADD_VEC_SS(r6,rD);
+    r7 = MUL_VEC_SS(r7,rC);
+    r8 = ADD_VEC_SS(r8,rF);
+    r9 = MUL_VEC_SS(r9,rE);
+    rA = ADD_VEC_SS(rA,rD);
+    rB = MUL_VEC_SS(rB,rC);
 
     /* Stop PAPI counters */
+    if ( NULL != fp && PAPI_stop(EventSet, iterValues) != PAPI_OK ) {
```
Treece-Burgess, line 500: Same as above.

src/counter_analysis_toolkit/vec_scalar_verify.c
```diff
@@ -535,8 +532,12 @@ float test_sp_scalar_VEC_24( uint64 iterations, int EventSet, FILE *fp ){
...
 /* Loop unrolling:  48 instructions */
 /************************************/
-float test_sp_scalar_VEC_48( uint64 iterations, int EventSet, FILE *fp ){
-    register SP_SCALAR_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+float test_sp_scalar_VEC_48( int EventSet, FILE *fp ){
+
+    volatile SP_SCALAR_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+    double values = 0.0;
+    long long iterValues[1]; iterValues[0] = 0;
+    for (int iter=0; iter<ITERS; ++iter) {
```
Treece-Burgess: Declare `iter` outside of for loop.

src/counter_analysis_toolkit/vec_scalar_verify.c
```diff
@@ -557,78 +558,76 @@ float test_sp_scalar_VEC_48( uint64 iterations, int EventSet, FILE *fp ){
     rF = SET_VEC_SS(0.16);
 
     /* Start PAPI counters */
-    if ( NULL != fp ) {
-      if ( PAPI_start( EventSet ) != PAPI_OK ) {
-        return -1;
-      }
+    if ( NULL != fp && PAPI_start( EventSet ) != PAPI_OK ) {
```
Treece-Burgess, line 566: Same as above.

src/counter_analysis_toolkit/vec_scalar_verify.c
```diff
@@ -557,78 +558,76 @@ float test_sp_scalar_VEC_48( uint64 iterations, int EventSet, FILE *fp ){
...
+    r5 = MUL_VEC_SS(r5,rE);
+    r6 = ADD_VEC_SS(r6,rD);
+    r7 = MUL_VEC_SS(r7,rC);
+    r8 = ADD_VEC_SS(r8,rF);
+    r9 = MUL_VEC_SS(r9,rE);
+    rA = ADD_VEC_SS(rA,rD);
+    rB = MUL_VEC_SS(rB,rC);
 
     /* Stop PAPI counters */
+    if ( NULL != fp && PAPI_stop(EventSet, iterValues) != PAPI_OK ) {
```
Treece-Burgess, line 624: Same as above. This is the last instance I will put such that I don't continue to clog up the review, but there are a few more instances that I will not comment on that will need to be changed if you do decide to return the error code returned from `PAPI_stop`.

src/counter_analysis_toolkit/vec_scalar_verify.c
```diff
@@ -924,8 +927,12 @@ double test_dp_scalar_VEC_24( uint64 iterations, int EventSet, FILE *fp ){
...
 /* Loop unrolling:  48 instructions */
 /************************************/
-double test_dp_scalar_VEC_48( uint64 iterations, int EventSet, FILE *fp ){
-    register DP_SCALAR_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+double test_dp_scalar_VEC_48( int EventSet, FILE *fp ){
+
+    volatile DP_SCALAR_TYPE r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,rA,rB,rC,rD,rE,rF;
+    double values = 0.0;
+    long long iterValues[1]; iterValues[0] = 0;
+    for (int iter=0; iter<ITERS; ++iter) {
```
Treece-Burgess: Declare `iter` outside of for loop.

src/counter_analysis_toolkit/Makefile
```diff
@@ -1,7 +1,7 @@
-PAPIDIR?=${PAPI_DIR}
+PAPI_DIR?=${PAPI_DIR}
 USEMPI?=false
-LDFLAGS=-L$(PAPIDIR)/lib -lpapi -lm -lpthread -ldl -lrt
-INCFLAGS=-I$(PAPIDIR)/include
+LDFLAGS=-L$(PAPI_DIR)/lib -lpapi -lm -lpthread -ldl -lrt
+INCFLAGS=-I$(PAPI_DIR)/include
 CFLAGS+=-g -Wall -Wextra
 OPT0=-O0
 OPT1=-O1
```
Treece-Burgess, line 6: A behavior I noticed in this PR, but appears to occur in master is the following:
1. On x86, build CAT for the first time. This will create `cat_collect`.
2. Move to another architecture lets say aarch, go into the `counter_analysis_toolkit` run `make clean` and then run `make`. Everything will build successfully.
3. When I attempt to run `./cat_collect` I run into the following:
```
[tburgess@hopper1 counter_analysis_toolkit]$ ./cat_collect -in event_list.txt -out OUTPUT_DIRECTORY -flops
-bash: ./cat_collect: cannot execute binary file: Exec format error
```

I have to then remove `cat_collect` and re run `make` such that step 3 will run.

Treece-Burgess (changes requested):

src/counter_analysis_toolkit/cat_arch.h
dbarry9: On certain architectures, such as ARM Neoverse V2 and Fujitsu A64FX, there are two instruction sets: NEON and SVE. Both instruction sets are fully functional on these architectures; however, when the FLOPs kernels in CAT use the NEON intrinsics, they do not trigger any events that the SVE instructions do not. Furthermore, the counts of the FLOPs events counts are stochastically well behaved when using the SVE kernels. For flexibility, I decided to keep the ifdefs related to NEON present in case the developers or users decide to experiment more with them. Only one of SVE or NEON should be used at a time in CAT.

src/counter_analysis_toolkit/Makefile
dbarry9: This is not guaranteed. Certain processors, such as the AMD Ryzen 9 9950X 16-Core (Zen5 architecture), has bf16 but not fp16.

src/counter_analysis_toolkit/gen_seq_dlopen.sh
dbarry9: This is unnecessary when using the Makefile. However, down the road, we plan to use a more robust build system. This change accommodates such a transition. For example, in CMake, something like the following:
```
add_custom_command(
    OUTPUT ${CMAKE_SOURCE_DIR}/src/icache_seq.h
           ${CMAKE_SOURCE_DIR}/src/icache_seq.c
           ${CMAKE_SOURCE_DIR}/src/icache_seq_kernel.c
    COMMAND bash ${CMAKE_SOURCE_DIR}/src/gen_seq_dlopen.sh ${CMAKE_SOURCE_DIR}/src
    DEPENDS ${CMAKE_SOURCE_DIR}/src/gen_seq_dlopen.sh
    VERBATIM
)
```
would need to be done in order for the shell script to know where exactly where the source lives.

src/counter_analysis_toolkit/vec.c
dbarry9: It is indeed a file.

src/counter_analysis_toolkit/cat_arch.h
dbarry9: Previously, the iterations were used to run the kernels multiple times in case there was slight run-to-run variation in the measured counts of FLOPs events. Historically, there was no such variation on the tested architectures. However, on ARM Neoverse V2 for example, the FLOPs events are _speculative_, and there was significant run-to-run variation with the kernel structure that used the `iterations` variables and while-loops. Removing them is part of re-designing the kernels to be less prone to this variation.

src/counter_analysis_toolkit/vec_dot_bf16.c
dbarry9: It could be, but I am partial to writing out all cases explicitly for the time being until we migrate to a more robust build system.
Edit to Add: I will follow the advice you give below w.r.t. placing `SUBDBG` calls in the else-cases.

src/counter_analysis_toolkit/vec_dot_bf16.c
dbarry9: Yes, it is. I will remove the extra space in the next push.

src/counter_analysis_toolkit/Makefile
dbarry9: I have followed this suggestion in my latest push.

src/counter_analysis_toolkit/flops.c
dbarry9: Correct, half-precision is not available on the available IBM architectures.

src/counter_analysis_toolkit/vec_scalar_verify.h
dbarry9: Yes, now that there is run-to-run variation, the event counts are averaged over 10k kernel executions.

src/counter_analysis_toolkit/vec_scalar_verify.c
dbarry9: I have addressed this in my latest push.

src/counter_analysis_toolkit/vec_scalar_verify.c
dbarry9: I have addressed this in my latest push.

src/counter_analysis_toolkit/vec_nonfma_sp.c
dbarry9: I have addressed this in my latest push.

src/counter_analysis_toolkit/vec_nonfma_dp.c
dbarry9: I have addressed this in my latest push.

src/counter_analysis_toolkit/vec_nonfma_sp.c
dbarry9: I have addressed this in my latest push.

src/counter_analysis_toolkit/vec_scalar_verify.c
dbarry9: I have addressed this in my latest push.

src/counter_analysis_toolkit/vec_scalar_verify.c
dbarry9: I have addressed this in my latest push.

src/counter_analysis_toolkit/vec_scalar_verify.c
dbarry9: I have addressed this in my latest push.

src/counter_analysis_toolkit/vec_scalar_verify.c
dbarry9: I have addressed this in my latest push.

src/counter_analysis_toolkit/vec_fma_sp.c
dbarry9: I have addressed this in my latest push.

src/counter_analysis_toolkit/vec_fma_sp.c
dbarry9: I have addressed this in my latest push.

src/counter_analysis_toolkit/vec_fma_sp.c
dbarry9: I have addressed this in my latest push.

src/counter_analysis_toolkit/vec_nonfma_dp.c
dbarry9: I have addressed this in my latest push.

src/counter_analysis_toolkit/vec_nonfma_dp.c
dbarry9: I have addressed this in my latest push.

PR #541, opened 2026-01-23
2026.01.22 initial rocm presets

dbarry9:
## Pull Request Description

This pull request adds the following to the `rocm` component:
- functions to facilitate ROCm presets, and
- presets for FMA instructions of various precisions.

These changes have been tested on the AMD MI250X architecture.

Here is the relevant output from `papi_avail` on the Frontier supercomputer:
```
================================================================================
  PAPI Component Preset Events
================================================================================
       Name           Code    Avail Deriv Description (Note)
PAPI_ROCM_FP16_FMA 0x80000080  Yes   No   ROCM Half precision (FP16) FMA instructions
    :device=0
        Mandatory device qualifier [0,1,2,3,4,5,6,7]
PAPI_ROCM_BF16_FMA 0x80000081  No    No   ROCM Half precision (BF16) FMA instructions
PAPI_ROCM_FP32_FMA 0x80000082  Yes   No   ROCM Single precision (FP32) FMA instructions
    :device=0
        Mandatory device qualifier [0,1,2,3,4,5,6,7]
PAPI_ROCM_FP64_FMA 0x80000083  Yes   No   ROCM Double precision (FP64) FMA instructions
    :device=0
        Mandatory device qualifier [0,1,2,3,4,5,6,7]
PAPI_ROCM_FP_FMA   0x80000084  Yes   Yes  ROCM floating-point FMA instructions
    :device=0
        Mandatory device qualifier [0,1,2,3,4,5,6,7]
PAPI_ROCM_FP8_OPS  0x80000085  No    No   ROCM 8-bit precision floating-point operations
================================================================================
```

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/papi_events.csv
+++ b/src/papi_events.csv
@@ -2409,6 +2409,14 @@ PRESET,PAPI_CUDA_FP_FMA,DERIVED_POSTFIX,N0|N1|+|N2|+|,cuda:::sm__sass_thread_ins
 cuda,GH100
 PRESET,PAPI_CUDA_FP8_OPS,NOT_DERIVED,cuda:::sm__ops_path_tensor_src_fp8:stat=sum
 
+# AMD MI250X architecture
+rocm,gfx90a
+#
+PRESET,PAPI_ROCM_FP16_FMA,NOT_DERIVED,rocm:::SQ_INSTS_VALU_FMA_F16:device=0
+PRESET,PAPI_ROCM_FP32_FMA,NOT_DERIVED,rocm:::SQ_INSTS_VALU_FMA_F32:device=0
+PRESET,PAPI_ROCM_FP64_FMA,NOT_DERIVED,rocm:::SQ_INSTS_VALU_FMA_F64:device=0
+PRESET,PAPI_ROCM_FP_FMA,DERIVED_POSTFIX,N0|N1|+|N2|+|,rocm:::SQ_INSTS_VALU_FMA_F16:device=0,rocm:::SQ_INSTS_VALU_FMA_F32:device=0,rocm:::SQ_INSTS_VALU_FMA_F64:device=0
+
 #########################
 # ARM Neoverse V2       #
 #########################


src/papi_events.csv
```diff
@@ -2387,6 +2387,14 @@ PRESET,PAPI_CUDA_FP_FMA,DERIVED_POSTFIX,N0|N1|+|N2|+|,cuda:::sm__sass_thread_ins
 cuda,GH100
 PRESET,PAPI_CUDA_FP8_OPS,NOT_DERIVED,cuda:::sm__ops_path_tensor_src_fp8:stat=sum
 
+# AMD MI250X architecture
+rocm,gfx90a
+#
+PRESET,PAPI_ROCM_FP16_FMA,NOT_DERIVED,rocm:::SQ_INSTS_VALU_FMA_F16:device=0
```
Treece-Burgess, line 2416: Q: In `papi_rocm_presets.h` you have 6 listed events, but here only 4 are defined. Are the other two for a separate architecture?

src/papi_events.csv
```diff
@@ -2387,6 +2387,14 @@ PRESET,PAPI_CUDA_FP_FMA,DERIVED_POSTFIX,N0|N1|+|N2|+|,cuda:::sm__sass_thread_ins
...
+PRESET,PAPI_ROCM_FP32_FMA,NOT_DERIVED,rocm:::SQ_INSTS_VALU_FMA_F32:device=0
+PRESET,PAPI_ROCM_FP64_FMA,NOT_DERIVED,rocm:::SQ_INSTS_VALU_FMA_F64:device=0
```
Treece-Burgess, line 2418: Q: It appears that none of the events listed will need an instance qualifier appended to them. For the cuda component, you show the stats qualifier as well in the `papi_avail` output as those events do need the stats qualifier.

Say for a future case you add an event which does require an instance value, is that already handled with this update or would more work need to be done?

src/components/rocm/roc_common.c
```diff
@@ -322,3 +322,37 @@ init_thread_id_fn(void)
     thread_id_fn = (_papi_hwi_thread_id_fn) ?
         _papi_hwi_thread_id_fn : _papi_getpid;
 }
+
+/* misc. */
+hsa_status_t
+get_chip_name( hsa_agent_t agent, void *data ) {
+
+    char *name = (char *) data;
```
Treece-Burgess, line 330: nit: I would personally be more descriptive with the variable naming not just here, but in other instances as well.

`name` should correspond to the `arch_name`, but it is not immediately clear this is what we actually mean. Long term, I believe it would help others that come along and work on the component that may not be as familiar with it.

src/components/rocm/roc_common.c
```diff
@@ -322,3 +322,37 @@ init_thread_id_fn(void)
...
+
+    hsa_device_type_t type;
+    hsa_status_t hsa_errno = hsa_agent_get_info_p(agent, HSA_AGENT_INFO_DEVICE, &type);
+    if ( HSA_STATUS_SUCCESS != hsa_errno ) {
+        return hsa_errno;
+    }
+
+    if (type == HSA_DEVICE_TYPE_GPU) {
+        hsa_errno = hsa_agent_get_info_p(agent, HSA_AGENT_INFO_NAME, name);
```
Treece-Burgess, line 339: From the docs `HSA_AGENT_INFO_NAME` is a null terminated `char[64]`. The variable you are passing is `arch_name[PAPI_2MAX_STR_LEN]` ([256]), I would swap to `PAPI_MIN_STR_LEN`.

src/components/rocm/rocm.c
```diff
@@ -345,6 +346,32 @@ rocm_dispatch_timer(int n __attribute__((unused)), hwd_siginfo_t *info, void *uc
...
+
+    /* Setup presets. */
+    char arch_name[PAPI_2MAX_STR_LEN];
+
+    /* Load preset table for every device type available on the system.
+     * As long as one of the cards has presets defined, then they should
+     * be available. */
+    int retval = rocd_get_chip_name(arch_name);
+    if ( retval != PAPI_OK ) {
+        SUBDBG("EXIT: Failed to get ROCm device name.\n");
```
Treece-Burgess, line 362: nit: Should ROCm not be AMD here?

src/components/rocm/rocm.c
```diff
@@ -345,6 +346,32 @@ rocm_dispatch_timer(int n __attribute__((unused)), hwd_siginfo_t *info, void *uc
...
+        return retval;
+    }
+
+    retval = _papi_load_preset_table_component( cname, arch_name, cidx );
+    if ( retval != PAPI_OK ) {
+        SUBDBG("EXIT: Failed to init ROCm component presets.\n");
```
Treece-Burgess, line 368: nit: I would take ROCm and make it lowercase as our component name is `rocm` and AMD themselves normally use ROCm.

src/components/rocm/roc_common.c
```diff
@@ -322,3 +322,37 @@ init_thread_id_fn(void)
...
+        }
+    }
+
+    return HSA_STATUS_SUCCESS;
+}
+
+int
+rocc_get_chip_name(char *name) {
+
+    hsa_status_t hsa_errno = hsa_iterate_agents_p(&get_chip_name, name);
```
Treece-Burgess, line 352: Q: From my understanding, `hsa_iterate_agents` will enumerate over the available agents and invoke the callback you have defined. Therefore, if we have 7 AMD devices this will iterate over all 7 calling your callback function. What if there are multiple mixed devices, i.e. MI250x and MI300A? From my testing you would only capture the very last one.

src/components/rocm/rocm.c
```diff
@@ -345,6 +346,32 @@ rocm_dispatch_timer(int n __attribute__((unused)), hwd_siginfo_t *info, void *uc
...
 
+static int rocm_init_comp_presets(void) {
+    SUBDBG("ENTER: Init ROCm component presets.\n");
+    int cidx = _rocm_vector.cmp_info.CmpIdx;
+    char *cname = _rocm_vector.cmp_info.name;
+
+    /* Setup presets. */
+    char arch_name[PAPI_2MAX_STR_LEN];
+
+    /* Load preset table for every device type available on the system.
```
Treece-Burgess, line 357: nit: I would change the comments from /* to // as then it becomes easier to comment blocks of code out later if you are debugging.

However, you can use `#if 0 #endif`.

This is something we should probably discuss internally as a team though.

src/components/rocm/papi_rocm_presets.h
```diff
@@ -0,0 +1,140 @@
+#ifndef __PAPI_ROCM_PRESETS_H__
+#define __PAPI_ROCM_PRESETS_H__
+
+hwi_presets_t _rocm_presets[PAPI_MAX_rocm_PRESETS] = {
```
Treece-Burgess, line 4: When I use `PAPI_get_event_info` with either the preset rocm define or the preset rocm event code the member variable `component_index` from the structure `PAPI_event_info_t` shows a value of 0. This should not be the case for the rocm component.

src/components/rocm/papi_rocm_presets.h
```diff
@@ -0,0 +1,140 @@
...
+/*  0 */ {"PAPI_ROCM_FP16_FMA",
+	  "ROCM FP16 FMA instr",
+	  "ROCM Half precision (FP16) FMA instructions", 0,
+	  0, PAPI_PRESET_BIT_MSC,
```
Treece-Burgess, line 8: On Frontier, when I call `PAPI_enum_event` with the modifier `PAPI_PRESET_ENUM_AVAIL` I do not see `PAPI_ROCM_FP16_FMA` output even though this preset is available on the architecture. When I used the modifier `PAPI_ENUM_EVENTS` it did appear in the output.

src/components/rocm/rocm.c
```diff
@@ -688,7 +715,22 @@ int
...
 {
     if (!_rocm_vector.cmp_info.initialized) {
-        return rocm_init_private();
+        int papi_errno = rocm_init_private();
+        if( PAPI_OK != papi_errno ) {
+            return papi_errno;
+        }
+
+        // Setup the presets.
+        papi_errno = rocm_init_comp_presets();
```
Treece-Burgess, line 724: nit: If you add a SUBDBG to this I could see the nested approach; however, one option would be to:
```
papi_errno = rocm_init_comp_presets();
if (papi_errno == PAPI_ENOIMPL) {
    papi_errno = PAPI_OK;
}

return papi_errno;
```

src/components/rocm/rocm.c
dbarry9: That's a good question. Taking one example: sometimes NVIDIA GPUs are referred to as CUDA devices. Additionally, elsewhere in the PAPI `rocm` component, the phrasing "ROCm device" is used. My goal was to be consistent with this and to distinguish a generic "device" (which could refer to CPU or GPU) from a device that is specifically ROCm-compatible (for example, Frontier has both AMD CPUs and AMD GPUs). I suppose that part of the confusion is that "device" is an overloaded term. Although, historically, device has had a GPU/non-host processor connotation.

PR #543, opened 2026-01-25
Restrict workflow permissions to minimum required

williamjallen:
## Pull Request Description
Limiting the scope of the GITHUB_TOKEN via the [permissions](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions) attribute is best practice to prevent a leaked GITHUB_TOKEN from allowing repository write access.

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/.github/workflows/amd_smi_component_workflow.yml
+++ b/.github/workflows/amd_smi_component_workflow.yml
@@ -8,6 +8,9 @@ on:
   # allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   component_tests:
     strategy:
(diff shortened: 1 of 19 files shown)


Treece-Burgess:
Hello @williamjallen, thank you for submitting the PR. 

I do not believe this change is necessary as [pull requests from forks](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/) are already set to `read`. 
> The token has write permissions to a number of [API endpoints](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token) except in the case of pull requests from forks which are always read.

To add, the entire set of permissions would already be set to `read`, this change just sets `contents` to `read`. 

In terms of a workflow_dispatch, this can be write, but to trigger a workflow_dispatch you need to have write access to the repository i.e. owner, collaborator, or admin. Is there another reason that you believe this change is warranted?

Treece-Burgess (approved):

PR #545, opened 2026-01-28
Add gaudi2 component for Intel Habana Gaudi2 AI accelerator

tokey-tahmid:
## Pull Request Description
This PR adds a new component for accessing hardware performance counters on Intel Habana Gaudi2 AI accelerator via the SPMU interface. Currently supports 189 native events covering TPC (81), EDMA (50), MME CTRL (8), and PDMA (50) engines with 6 simultaneous hardware counters per SPMU unit.

## Implementation
Uses `hlthunk_debug()` ioctl interface with:
- `HL_DEBUG_OP_SET_MODE` - Enable debug mode
- `HL_DEBUG_OP_SPMU` - Configure SPMU event selection
- `HL_DEBUG_OP_READBLOCK` - Read counter values

### Component:
- `linux-gaudi2.c` - Main component implementation
- `gaudi2_events.h` - Event definitions and SPMU base addresses
- `Rules.gaudi2` - Build configuration
- `README.md` - Component documentation

### Tests (ctypes-based):

- `tests/Makefile` - Build system for tests
- `tests/python/run_tests.sh` - Test runner script
- `tests/python/test_component_and_events.py` - Event enumeration and component availability tests
- `tests/python/test_start_stop_read.py` - Basic start/stop/read lifecycle test on TPC events
- `tests/python/test_multidevice.py` - Multi-device counter isolation test
- `tests/python/test_mme_events.py` - MME CTRL SPMU event test
- `tests/python/test_edma_events.py` - EDMA SPMU event test

### Tests (cyPAPI-based):

- `tests/cypapi_tests/Makefile` - Builds cyPAPI dependency
- `tests/cypapi_tests/run_tests.sh` - Test runner script
- `tests/cypapi_tests/test_start_stop_read_cypapi.py` - Basic lifecycle test using cyPAPI
- `tests/cypapi_tests/test_multidevice_cypapi.py` - Multi-device test using cyPAPI
- `tests/cypapi_tests/test_mme_events_cypapi.py` - MME CTRL event test using cyPAPI
- `tests/cypapi_tests/test_edma_events_cypapi.py` - EDMA event test using cyPAPI 

## Testing
Tested on Voyager machine with Gaudi2 (HL-225)
- All of the following PAPI utilities successfully work with the component:
    - `papi_component_avail` -  ✅
    - `papi_avail` -  ✅
    - `papi_native_avail` -  ✅
    - `papi_command_line` -  ✅
- Successfully instrumented workloads with the `gaudi2` component to test the events and counter values:
    - Python example -  ✅
    - Custom TPC kernel -  ✅

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**


src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+    {"TPC_NUM_OF_OPCODE3_EXECUTED", "Number of opcode3 executed", GAUDI2_ENGINE_TPC, TPC_SPMU_NUM_OF_OPCODE3_EXECUTED},
+    {"TPC_NUM_OF_OPCODE4_EXECUTED", "Number of opcode4 executed", GAUDI2_ENGINE_TPC, TPC_SPMU_NUM_OF_OPCODE4_EXECUTED},
+
+    /* TPC execution */
+    {"TPC_KERNEL_EXECUTED", "TPC kernels executed", GAUDI2_ENGINE_TPC, TPC_SPMU_KERNEL_EXECUTED},
+    {"TPC_SCALAR_PIPE_EXEC", "Scalar pipe execution cycles", GAUDI2_ENGINE_TPC, TPC_SPMU_SCALAR_PIPE_EXEC},
+    {"TPC_VECTOR_PIPE_EXEC", "Vector pipe execution cycles", GAUDI2_ENGINE_TPC, TPC_SPMU_VECTOR_PIPE_EXEC},
+
+    /* TPC cache */
+    {"TPC_ICACHE_MISS", "Instruction cache miss", GAUDI2_ENGINE_TPC, TPC_SPMU_ICACHE_MISS},
```
Treece-Burgess: Questions:
1. Can you not programmatically get the events on the architecture?
2. At what point do you check if these hard coded events exist on the architecture and if they do not exist will they just be removed from the event pool that outputs with `papi_native_avail`?

src/components/gaudi2/Rules.gaudi2
```diff
@@ -0,0 +1,32 @@
...
+# Expected directory structure under PAPI_GAUDI2_ROOT:
+#   $(PAPI_GAUDI2_ROOT)/include/habanalabs/hlthunk.h
+#   $(PAPI_GAUDI2_ROOT)/lib/habanalabs/libhl-thunk.so
+
+# Component source files
+COMPSRCS += components/gaudi2/linux-gaudi2.c
+COMPOBJS += linux-gaudi2.o
+
+# Require PAPI_GAUDI2_ROOT to be set
+ifndef PAPI_GAUDI2_ROOT
```
Treece-Burgess: That I am aware of no other component has a message print out if the component env is not set:
```
components/gaudi2/Rules.gaudi2:16: *** PAPI_GAUDI2_ROOT must be set. Example: export PAPI_GAUDI2_ROOT=/usr.  Stop.
```

Also, you can set `PAPI_GAUDI2_ROOT` to have a default value if not set as from the README it appears it could be `/usr`:
```
PAPI_GAUDI2_ROOT ?=  /usr
```

src/components/gaudi2/Rules.gaudi2
```diff
@@ -0,0 +1,32 @@
+# Rules for the Gaudi2 component
```
Treece-Burgess, line 1: Q: Not that I can run them, but are you planning to push a tests directory?

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+#include <linux/types.h>
+
+#include "papi.h"
+#include "papi_internal.h"
+#include "papi_vector.h"
+#include "papi_memory.h"
+
+#include "gaudi2_events.h"
+
+#define GAUDI2_COMPONENT_NAME "gaudi2"
```
Treece-Burgess: This define does not appear to be used.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+
+    SUBDBG("Initializing Gaudi2 component (cidx=%d)\n", cidx);
+
+    _gaudi2_vector.cmp_info.CmpIdx = cidx;
+
+    papi_errno = load_hlthunk_library();
+    if (papi_errno != PAPI_OK) {
+        snprintf(_gaudi2_vector.cmp_info.disabled_reason,
+                 PAPI_MAX_STR_LEN, "Failed to load libhl-thunk.so");
+        _gaudi2_vector.cmp_info.disabled = 1;
```
Treece-Burgess: For `.disabled` this is normally set to the error code that caused the component to be disabled.  In this case just `_gaudi2_vector.cmp_info.disabled = papi_errno`.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+
+    /* Try to find existing fd from PyTorch, else open ourselves */
+    gaudi2_device_fd = find_gaudi2_device_fd();
+    if (gaudi2_device_fd < 0)
+        gaudi2_device_fd = p_hlthunk_open(HLTHUNK_DEVICE_DONT_CARE, NULL);
+
+    if (gaudi2_device_fd < 0) {
+        snprintf(_gaudi2_vector.cmp_info.disabled_reason,
+                 PAPI_MAX_STR_LEN, "No Gaudi2 device found");
+        _gaudi2_vector.cmp_info.disabled = 1;
```
Treece-Burgess: Same situation as above, `.disabled` should be `PAPI_ENOSUPP`.

In the case where you have multiple possible conditionals that could cause a failure you could set `papi_errno` to the error code and then have a `goto` statement. That is how a lot of the components do it.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+
+    /* MME */
+    {"MME_NUM_OUTER_PRODUCTS", "MME outer products", GAUDI2_ENGINE_MME, MME_CTRL_SPMU_NUM_OUTER_PRODUCTS},
+    {"MME_STALL_ON_A", "MME stall waiting for A matrix", GAUDI2_ENGINE_MME, MME_CTRL_SPMU_OUTER_PRODUCT_STALL_ON_A},
+    {"MME_STALL_ON_B", "MME stall waiting for B matrix", GAUDI2_ENGINE_MME, MME_CTRL_SPMU_OUTER_PRODUCT_STALL_ON_B},
+
+    {NULL, NULL, 0, 0}
+};
+
+static int num_native_events = 0;
```
Treece-Burgess: nit: Instead of having this global variable that is carried around, you could just use `_gaudi2_vector.cmp_info.num_native_events`.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+} gaudi2_control_t;
+
+/* Per-thread context */
+typedef struct {
+    int device_fd;
+    int debug_mode_enabled;
+    int spmu_enabled;
+} gaudi2_context_t;
+
+static int gaudi2_initialized = 0;
```
Treece-Burgess: Same as above.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+        .default_domain = PAPI_DOM_USER,
+        .available_domains = PAPI_DOM_USER | PAPI_DOM_KERNEL,
+        .default_granularity = PAPI_GRN_THR,
+        .available_granularities = PAPI_GRN_THR,
+        .hardware_intr_sig = PAPI_INT_SIGNAL,
+        .fast_real_timer = 0,
+        .fast_virtual_timer = 0,
+        .attach = 0,
+        .attach_must_ptrace = 0,
+        .initialized = 0,
```
Treece-Burgess: I cannot verify this programmatically, but based off the code I do not believe this value will be changed even if `gaudi2` is  initialized. You change `gaudi2_initialized`, but I never see component info `.initialized` changed.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+    DIR *dir;
+    struct dirent *entry;
+    char link_path[64];
+    char target[512];
+    ssize_t len;
+    int found_fd = -1;
+
+    dir = opendir("/proc/self/fd");
+    if (!dir)
+        return -1;
```
Treece-Burgess, line 542: I would add a `SUBDBG` and use `PAPI_ESYS` instead of -1.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+
+    while ((entry = readdir(dir)) != NULL) {
+        if (entry->d_name[0] == '.')
+            continue;
+
+        snprintf(link_path, sizeof(link_path), "/proc/self/fd/%.20s", entry->d_name);
```
Treece-Burgess: The return value of `snprintf` needs to be checked. 
```
int strLen = snprintf(..., string_length, ..., ...)
if (strLen < 0 || strLen >= string_length) {
    SUBDBG(...)
    return PAPI_EBUF;
}
```

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+        len = readlink(link_path, target, sizeof(target) - 1);
+        if (len < 0)
+            continue;
+        target[len] = '\0';
+
+        /* Look for /dev/accel/accel* (not control device) */
+        if (strstr(target, "/dev/accel/accel") != NULL &&
+            strstr(target, "control") == NULL) {
+            found_fd = atoi(entry->d_name);
+            SUBDBG("Found Gaudi2 device: fd=%d -> %s\n", found_fd, target);
```
Treece-Burgess: Q: Can there be multiple Gaudi2 devices? As it appears you break right when you find a match. For cuda there can be multiple devices likewise for amd_smi and rocp_sdk.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+            break;
+        }
+    }
+
+    closedir(dir);
```
Treece-Burgess: `closedir` can fail via [EBADF and EINTR](https://pubs.opengroup.org/onlinepubs/009604499/functions/closedir.html). To be safe, I would check the return status of `closedir` and return a PAPI error code if it fails, i.e.:
```
int status = closedir()
if (status == -1) {
    SUBDBG(...)
    return PAPI_ESYS;
}
```

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+
+/*
+ * Find existing Gaudi2 device fd from /proc/self/fd.
+ * When PyTorch has already opened the device.
+ */
+static int find_gaudi2_device_fd(void)
+{
+    DIR *dir;
+    struct dirent *entry;
+    char link_path[64];
```
Treece-Burgess: Can use `PAPI_MIN_STR_LEN`.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+static int gaudi2_device_fd = -1;
+static unsigned int gaudi2_lock;
+
+papi_vector_t _gaudi2_vector;
+
+/* Load hlthunk library */
+static int load_hlthunk_library(void)
+{
+    const char *lib_paths[] = {
+        "/usr/lib/habanalabs/libhl-thunk.so",
```
Treece-Burgess: Currently you are hard coding these paths, but you should take into account the user setting `PAPI_GAUDI2_ROOT` as the top priority to construct the search path initially. 

However, you can definitely have fall back paths you want to search after that. These can be hard coded based off your knowledge of where `libhl-thunk.so` could be. There is similar logic in the cuda component.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+    int papi_errno = PAPI_OK;
+    int i;
+
+    for (i = 0; i < num_counters; i++)
+        values[i] = 0;
+
+    read_buffer = mmap(NULL, 4096, PROT_READ | PROT_WRITE,
+                       MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+    if (read_buffer == MAP_FAILED)
+        return PAPI_ENOMEM;
```
Treece-Burgess: nit: Huge nitpick, but I would follow the return workflow that you use below for here.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+
+cleanup:
+    if (read_buffer != MAP_FAILED)
+        munmap(read_buffer, 4096);
+    return papi_errno;
+}
+
+static uint64_t get_spmu_base_address(gaudi2_engine_type_t engine, int dcore, int instance)
+{
+    static const uint64_t tpc_spmu_bases[4][6] = {
```
Treece-Burgess: Q: Could the values of 4 and 6 here and 4 and 2 below ever change? 

If so I would probably not hard code these values and instead define them with variables.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+
+    if (count > GAUDI2_MAX_COUNTERS)
+        return PAPI_ECOUNT;
+
+    gaudi2_ctl->num_counters = count;
+
+    for (int i = 0; i < count; i++) {
+        int event_idx = native[i].ni_event;
+
+        if (event_idx < 0 || event_idx >= num_native_events)
```
Treece-Burgess: For conditionals that will result in a return of an errorcode, I would add a `SUBDBG` to each with a message indicating what has failed. It helps with anyone coming along to debug by setting `export PAPI_DEBUG=SUBSTRATE`.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+    if (papi_errno != PAPI_OK)
+        return papi_errno;
+    gaudi2_ctx->spmu_enabled = 1;
+
+    for (int i = 0; i < gaudi2_ctl->num_counters; i++) {
+        gaudi2_ctl->counters[i].last_value = 0;
+        gaudi2_ctl->counters[i].accumulated = 0;
+    }
+
+    gaudi2_ctl->running = 1;
```
Treece-Burgess: You could add `#define`'s to indicate the status of a Gaudi2 eventset. Other components follow this:
```
#define GAUDI2_EVENTS_RUNNING        (0x2)
```

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+            }
+        }
+    }
+
+    if (gaudi2_ctx->spmu_enabled) {
+        disable_spmu(gaudi2_ctx->device_fd, 0);
+        gaudi2_ctx->spmu_enabled = 0;
+    }
+
+    gaudi2_ctl->running = 0;
```
Treece-Burgess: Likewise as above.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+        return PAPI_ENOEVNT;
+
+    strncpy(name, gaudi2_native_events[EventCode].name, len);
+    name[len - 1] = '\0';
+    return PAPI_OK;
+}
+
+static int gaudi2_ntv_name_to_code(const char *name, unsigned int *EventCode)
+{
+    for (int i = 0; i < num_native_events; i++) {
```
Treece-Burgess: This design is similar to the `appio` component.

From my understanding you are hard coding the events into this file and there are not an over abundance of them. Therefore, the for loop is fine and we would not incur a large overhead here. However, down the line what if you end up adding more events or even an enumerator function is added to the library? In that case a hash table would be a better suited design long term.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+            return PAPI_EINVAL;
+    }
+}
+
+static int gaudi2_ntv_code_to_name(unsigned int EventCode, char *name, int len)
+{
+    if (EventCode >= (unsigned int)num_native_events)
+        return PAPI_ENOEVNT;
+
+    strncpy(name, gaudi2_native_events[EventCode].name, len);
```
Treece-Burgess: Use `snprintf` over `strncpy`.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+    }
+    return PAPI_ENOEVNT;
+}
+
+static int gaudi2_ntv_code_to_descr(unsigned int EventCode, char *descr, int len)
+{
+    if (EventCode >= (unsigned int)num_native_events)
+        return PAPI_ENOEVNT;
+
+    strncpy(descr, gaudi2_native_events[EventCode].description, len);
```
Treece-Burgess: Use `snprintf` over `strncpy`.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+}
+
+static int gaudi2_ntv_code_to_info(unsigned int EventCode, PAPI_event_info_t *info)
+{
+    if (EventCode >= (unsigned int)num_native_events)
+        return PAPI_ENOEVNT;
+
+    gaudi2_native_event_t *event = &gaudi2_native_events[EventCode];
+
+    strncpy(info->symbol, event->name, sizeof(info->symbol) - 1);
```
Treece-Burgess: Use `snprintf` over `strncpy`.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+    info->symbol[sizeof(info->symbol) - 1] = '\0';
+
+    strncpy(info->long_descr, event->description, sizeof(info->long_descr) - 1);
```
Treece-Burgess: Use `snprintf` over `strncpy`.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -0,0 +1,773 @@
...
+    info->long_descr[sizeof(info->long_descr) - 1] = '\0';
+
+    strncpy(info->short_descr, event->description, sizeof(info->short_descr) - 1);
```
Treece-Burgess: Use `snprintf` over `strncpy`.

Treece-Burgess (changes requested):

src/components/gaudi2/Rules.gaudi2
tokey-tahmid: I agree, the message is not required in Rules.gaudi2. I removed it and added PAPI_GAUDI2_ROOT ?= /usr as the default.

src/components/gaudi2/Rules.gaudi2
tokey-tahmid: I plan to push a tests directory in another PR after this one is merged.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: Removed.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: Thanks for this suggestion. I updated the code accordingly for this and the following cases.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: fixed.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: fixed.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: removed from vector table static init.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: fixed.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: added.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: added.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: As of right now, we only worked with one Gaudi2 device. I have noted a TODO for multiple devices. I plan to handle this case in another PR.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: added.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: thanks, added.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: updated.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: sure!

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: updated.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: added.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: added.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: added.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: I noted a TODO here. I plan to add the full list of Gaudi2 events in a future PR. I will update the code to a hash table in that PR.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: updated.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: updated.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: updated.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: updated.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: updated.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: 1. It is a two-step process to get the event ids/names and the event descriptions. I am still working on this to get the events (names + description) programmatically from the device.
2. At its current state, the gaudi2_native_events[] table is entirely hardcoded. During init_component, the code just counts entries in the table without validating them against the hardware. All events always appear in papi_native_avail output. There's also currently no check that the opened device is actually a Gaudi2. I am also working on this to validate the device, and if not Gaudi2, then disable the component with an appropriate message.

Are any of these two features a requirement to merge this PR? If not, I plan to add them in a future PR.

Treece-Burgess: > Are any of these two features a requirement to merge this PR? If not, I plan to add them in a future PR.

Having hard coded events is fine if there is no enumerator function to get the events from. `rocm_smi` and `amd_smi` have hard coded events as there is not an enumerator function.

What I think should be included in this PR to merge are:
1. A check to make sure those hard coded events are viable for the user to add and if not then they are removed from the event pool. 
    -    The above is ideal, but it may also be sufficient that it can be checked at runtime when a user adds an event. I believe `amd_smi` checks to make sure the event is available on the architecture before making them available in the event pool; however, for `perf_event` this does not happen and some events that output with `papi_native_avail` are unable to be added successfully.

2. Making sure the opened device is a Gaudi2, as you do not want the component to be active if the device isn't even Gaudi2. This I think is much less flexible and should be added.

src/components/gaudi2/Rules.gaudi2
Treece-Burgess: I think it would be best to add the tests directory here in the same PR as well. Even if it is just a singular test that is fine, you can always add more later.

src/components/gaudi2/linux-gaudi2.c
Treece-Burgess: If the component does not support multiple devices as of now, then you may want to move forward with adding it to this PR or disable the component.

My thought for the latter is that if we can't confirm it runs correctly with multiple devices or the code is not there then disabling it makes sense.

src/components/gaudi2/linux-gaudi2.c
Treece-Burgess: I think that is fair to do in a separate PR in the future.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: These two features have been added with the recent commits. They work as follows:

1. The code now checks if the opened device is Gaudi2 and disables the component for other devices.
2. Added validation of hard-coded events - Queries engine masks and filters events; as a result, papi_native_avail would only show events for engines actually present on the hardware

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -73,21 +73,73 @@ struct hl_debug_params_read_block {
...
 typedef int (*hlthunk_open_fn)(int device_type, const char *busid);
 typedef int (*hlthunk_close_fn)(int fd);
 typedef int (*hlthunk_debug_fn)(int fd, struct hl_debug_args *debug);
 typedef int (*hlthunk_get_device_name_from_fd_fn)(int fd);
 
+/* hlthunk_hw_ip_info - hardware IP information structure */
+#define HL_INFO_VERSION_MAX_LEN_LOCAL  128
+#define HL_INFO_CARD_NAME_MAX_LEN_LOCAL 16
+
+struct hlthunk_hw_ip_info {
```
Treece-Burgess: Q: Why do we need to have our own structure define that is then passed to `p_hlthunk_get_hw_ip_info`? Is there not an existing structure in the library?

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -516,9 +565,88 @@ static int gaudi2_init_component(int cidx)
...
+        SUBDBG("Failed to allocate filtered event table\n");
+        snprintf(_gaudi2_vector.cmp_info.disabled_reason,
+                 PAPI_MAX_STR_LEN, "Failed to allocate event table");
+        _gaudi2_vector.cmp_info.disabled = papi_errno;
+        return papi_errno;
+    }
+
+    /* Second pass: copy available events into filtered table */
+    int idx = 0;
+    for (int i = 0; i < catalog_size; i++) {
```
Treece-Burgess: Declare `i` outside of the for loop.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -516,9 +565,88 @@ static int gaudi2_init_component(int cidx)
...
+
+    /* First pass: count how many catalog events are available */
+    int catalog_size = 0;
+    while (gaudi2_event_catalog[catalog_size].name != NULL)
+        catalog_size++;
+
     num_events = 0;
-    while (gaudi2_native_events[num_events].name != NULL)
-        num_events++;
+    for (int i = 0; i < catalog_size; i++) {
```
Treece-Burgess: Declare `i` outside of the for loop.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -516,9 +565,88 @@ static int gaudi2_init_component(int cidx)
...
+    /* Query hardware IP info to determine which engines are enabled.
+     * Only expose events for engine types that 
+     * have at least one enabled instance */
+    struct hlthunk_hw_ip_info hw_ip;
+    memset(&hw_ip, 0, sizeof(hw_ip));
+
+    if (p_hlthunk_get_hw_ip_info(gaudi2_device_fd, &hw_ip) != 0) {
+        papi_errno = PAPI_ESYS;
+        SUBDBG("Failed to get hw_ip_info from device\n");
+        snprintf(_gaudi2_vector.cmp_info.disabled_reason,
```
Treece-Burgess: Two items here:

1. You need to check the return value of `snprintf`
```
int strLen = snprintf(..., strings_length, ..., ...)'
if (strLen < 0 || strLen >= strings_length) error_out
```

2. I would default to use `PAPI_HUGE_STR_LEN` over PAPI_MAX_STR_LEN` as that is the max allocated space `disabled_reason` has.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -516,9 +565,88 @@ static int gaudi2_init_component(int cidx)
...
+            case GAUDI2_ENGINE_MME:  if (mme_avail)  num_events++; break;
+            case GAUDI2_ENGINE_PDMA: if (pdma_avail) num_events++; break;
+            default: break;
+        }
+    }
+
+    if (num_events == 0) {
+        papi_errno = PAPI_ENOSUPP;
+        SUBDBG("No available engine types on this device\n");
+        snprintf(_gaudi2_vector.cmp_info.disabled_reason,
```
Treece-Burgess: Same as above.

src/components/gaudi2/linux-gaudi2.c
```diff
@@ -516,9 +565,88 @@ static int gaudi2_init_component(int cidx)
...
+        return papi_errno;
+    }
+
+    /* Allocate filtered event table */
+    gaudi2_native_events = (gaudi2_native_event_t *)
+        papi_calloc(num_events, sizeof(gaudi2_native_event_t));
+    if (!gaudi2_native_events) {
+        papi_errno = PAPI_ENOMEM;
+        SUBDBG("Failed to allocate filtered event table\n");
+        snprintf(_gaudi2_vector.cmp_info.disabled_reason,
```
Treece-Burgess: Same as above.

Treece-Burgess (dismissed):

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: We don't actually. I intended to remove these structs and include the hlthunk header where those structs exist, but forgot. Fixed it in the latest commit.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: Fixed it in the latest commit.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: Fixed it in the latest commit.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: Fixed it in the latest commit.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: Fixed it in the latest commit.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: Fixed it in the latest commit.

src/components/gaudi2/linux-gaudi2.c
tokey-tahmid: Added multi-device support with the latest commit. `papi_native_avail` output shows:

```
 Native Events in Component: gaudi2
===============================================================================
| gaudi2:::TPC_MEMORY2SB_BP                                                    |
|            Memory to SB backpressure                                         |
|            Mandatory device qualifier [0,1,2,3]                              |
| gaudi2:::TPC_SB2MEMORY_BP                                                    |
|            SB to memory backpressure                                         |
|            Mandatory device qualifier [0,1,2,3]                              |

```

src/components/gaudi2/Rules.gaudi2
tokey-tahmid: Component tests directory added with 3 Python test scripts so far:

1. Component availability and events enumeration test
2. Running a PyTorch workload and collecting event data with PAPI start/stop on one device
3. Running a PyTorch workload on multi-device

dbarry9:
I tested this PR on the Voayger system at SDSC, which contains the Gaudi2 architecture. The build of this feature branch includes functional PAPI utilities that expose Gaudi2 events. All five of the `gaudi2` component tests pass. 
[papi_native_avail.txt](https://github.com/user-attachments/files/30012382/papi_native_avail.txt)
[papi_mem_info.txt](https://github.com/user-attachments/files/30012383/papi_mem_info.txt)
[papi_hardware_avail.txt](https://github.com/user-attachments/files/30012385/papi_hardware_avail.txt)
[papi_component_avail.txt](https://github.com/user-attachments/files/30012386/papi_component_avail.txt)
[papi_command_line.txt](https://github.com/user-attachments/files/30012387/papi_command_line.txt)
[papi_avail.txt](https://github.com/user-attachments/files/30012388/papi_avail.txt)
[hl-smi_-L.txt](https://github.com/user-attachments/files/30012389/hl-smi_-L.txt)
[component_tests.txt](https://github.com/user-attachments/files/30012390/component_tests.txt)

dbarry9 (approved):

PR #548, opened 2026-01-30
resolve contentious AMD components

### LINKED ISSUE #407: `rocm`/`rocp_sdk`: `hsa_init` Behavior
Both the `rocm` and `rocp_sdk` components call `hsa_init` and `hsa_shut_down`.  Configuring PAPI with BOTH `rocm` and `rocp_sdk` will lead to issues in the `rocm` component, specifically the call to `hsa_init` will fail with the error `HSA_STATUS_ERROR_OUT_OF_RESOURCES`. 

We were able to resolve this issue by restructuring where `hsa_shut_down` and `dlclose(rocm_dlp)` were called in the `rocp_sdk` component. This can be seen done in PR #401. 

However, restructuring where `hsa_shut_down` and `dlclose(rocm_dlp)` are called in the `rocp_sdk` component leads to `PAPI_read` failing. Below is output from the test `simple_sampling` using PR #401 on a machine with 2*MI210's. 
```
FAILED!!!
Line # 91 Error in PAPI_read: Not supported by component
Some tests require special hardware, permissions, OS, compilers
or library versions. PAPI may still function perfectly on your 
system without the particular feature being tested here.  
```

### LINKED ISSUE #416: rocp_sdk: provide a way to disable component in rocprofiler_configure
Because rocprofiler-sdk searches for symbols named `rocprofiler_configure` and invokes them during runtime initialization, a lot of the rocp_sdk PAPI code is invoked before a call to PAPI_disable_component_by_name can be called. It'd be nice if there were some PAPI environment variable users could set in order to short-circuit PAPI's rocp_sdk `rocprofiler_configure` code. Something at the top of `rocprofiler_configure` like the following:

```
    const char* env_val = getenv("PAPI_ROCP_SDK_DISABLE");
    if (env_val != NULL && 
        (strcmp(env_val, "1") == 0 || strcasecmp(env_val, "yes") == 0)) {
        // PAPI rocp_sdk component is disabled
        return NULL;
    }
```

This would prevent further logic from being executed that may cause unintentional side-effects (I have personally seen some benign logs to stdout due to PAPI's rocp_sdk component invoking some rocprofiler-sdk functions).

### LINKED ISSUE #478: Support a PAPI_DISABLE_COMPONENTS environment variable
### Description of Feature Request

Sometimes a PAPI component configured for the PAPI library can exhibit unexpected behavior or else otherwise not execute properly. This usually is during initialization. There is currently no way to disable a configured component outside of using the API. Locally we have enhance PAPI with a `PAPI_DISABLE_COMPONENTS` environment variable to disable selected components from being activated at runtime. This is particularly helpful for PAPI utilities `papi_component_avail`, `papi_avail`, etc.

Below is a GIT `diff` of how we implemented this environment variable. It has proven useful on a number of occasions where issues in a component prevented the successful execution of a PAPI utility, or even a program using PAPI.

Thanks for the consideration.

```
@@ -1047,6 +1051,20 @@ PAPI_library_init( int version )
 {
     APIDBG( "Entry: version: %#x\n", version);
 
+#if defined(_CRAY)
+       /* Disable components early */
+
+       char *penv = getenv ("PAPI_DISABLE_COMPONENTS");
+
+       if (penv != NULL) {
+               char *p;
+
+               for (p = strtok (penv, ",:"); p != NULL; p = strtok (NULL, ",:")) {
+                       (void) PAPI_disable_component_by_name (p);
+               }
+       }
+#endif /* _CRAY */
+
        int tmp = 0;
 
        /* This is a poor attempt at a lock. 
```

### Component

_No response_

dbarry9:
## Pull Request Description

- framework: PAPI_DISABLE_COMPONENTS env var
- amd comps: contentious components in same config

This pull request resolves issues #407, #416, and #478.

ROCm version >= 6.3.2 being the "cutoff" for making `rocp_sdk` active by default over `rocm` was chosen due to known bugs in the ROCProfiler SDK in prior releases.

These changes have been tested using ROCm 7.0.2 on the Frontier supercomputer, which contains the AMD MI250X architecture.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/amd_smi/linux-amd-smi.c
+++ b/src/components/amd_smi/linux-amd-smi.c
@@ -57,11 +57,40 @@ static int _amd_smi_init_component(int cidx) {
     _amd_smi_vector.cmp_info.num_mpx_cntrs = -1;
     _amd_smi_lock = PAPI_NUM_LOCK + NUM_INNER_LOCK + cidx;
 
-    CHECK_SNPRINTF(doNotAllowTruncation, _amd_smi_vector.cmp_info.disabled_reason, PAPI_MAX_STR_LEN,
-             "Not initialized. Access an AMD SMI event to initialize.");
-    _amd_smi_vector.cmp_info.disabled = PAPI_EDELAY_INIT;
+    /* Manage contension between rocm_smi and amd_smi components. */
+    int use_amd_smi = 0;
+    #if defined(DEFAULT_TO_AMD_SMI)
+    use_amd_smi = 1;
+    #endif
+    #if defined(DEFAULT_TO_ROCM_SMI)
+        char *disabledComps = getenv("PAPI_DISABLE_COMPONENTS");
+        if (disabledComps != NULL) {
+            char *penv = strdup(disabledComps);
+            char *p;
+            for (p = strtok (penv, ",:"); p != NULL; p = strtok (NULL, ",:")) {
+                if(!strcmp(p, "rocm_smi")) use_amd_smi = 1;
+            }
+            free(penv);
+        } else {
+            SUBDBG("amd_smi: getenv(PAPI_DISABLE_COMPONENTS) was not set.\n");
+        }
+    #endif
 
-    return PAPI_EDELAY_INIT;
+    int papi_errno;
+    if (use_amd_smi) {
+        CHECK_SNPRINTF(_amd_smi_vector.cmp_info.disabled_reason, PAPI_HUGE_STR_LEN,
+                 "Not initialized. Access an AMD SMI event to initialize.");
+        papi_errno = PAPI_EDELAY_INIT;
+        _amd_smi_vector.cmp_info.disabled = papi_errno;
+        return papi_errno;
+    } else {
+        CHECK_SNPRINTF(_amd_smi_vector.cmp_info.disabled_reason, PAPI_HUGE_STR_LEN,
+                 "Not active while rocm_smi component is active. Set 'export PAPI_DISABLE_COMPONENTS=rocm_smi' to override.");
+        papi_errno = PAPI_ECOMBO;
+        _amd_smi_vector.cmp_info.disabled = papi_errno;
+        return papi_errno;
+    }
+
 }
 
 static int evt_get_count(int *count) {
@@ -92,9 +121,14 @@ static int _amd_smi_init_private(void) {
         amds_err_get_last(&error_str);
         if (!error_str || !error_str[0])
             error_str = "AMD SMI component initialization failed";
+<<<<<<< HEAD
         CHECK_SNPRINTF(doNotAllowTruncation, _amd_smi_vector.cmp_info.disabled_reason,
                  sizeof _amd_smi_vector.cmp_info.disabled_reason, "%s",
                  error_str);
+=======
+        CHECK_SNPRINTF(_amd_smi_vector.cmp_info.disabled_reason,
+                 PAPI_HUGE_STR_LEN, "%s", error_str);
+>>>>>>> e27d70a7e (amd comps: contentious components in same config)
         goto fn_fail;
     }
 
--- a/src/components/rocm/rocm.c
+++ b/src/components/rocm/rocm.c
@@ -133,24 +133,57 @@ rocm_init_component(int cid)
     _rocm_lock = PAPI_NUM_LOCK + NUM_INNER_LOCK + cid;
     SUBDBG("ENTER: cid: %d\n", cid);
 
-    int papi_errno = rocd_init_environment();
-    if (papi_errno != PAPI_OK) {
-        _rocm_vector.cmp_info.initialized = 0;
-        _rocm_vector.cmp_info.disabled = papi_errno;
-        const char *err_string;
-        rocd_err_get_last(&err_string);
-        int expect = snprintf(_rocm_vector.cmp_info.disabled_reason,
-                              PAPI_MAX_STR_LEN, "%s", err_string);
-        if (expect > PAPI_MAX_STR_LEN) {
+    /* Manage contension between rocm and rocp_sdk components. */
+    int use_rocm = 0;
+    #if defined(DEFAULT_TO_ROCM)
+    use_rocm = 1;
+    #endif
+    #if defined(DEFAULT_TO_ROCP_SDK)
+        char *disabledComps = getenv("PAPI_DISABLE_COMPONENTS");
+        if (disabledComps != NULL) {
+            char *penv = strdup(disabledComps);
+            char *p;
+            for (p = strtok (penv, ",:"); p != NULL; p = strtok (NULL, ",:")) {
+                if(!strcmp(p, "rocp_sdk")) use_rocm = 1;
+            }
+            free(penv);
+        } else {
+            SUBDBG("rocm: getenv(PAPI_DISABLE_COMPONENTS) was not set.\n");
+        }
+    #endif
+
+    int papi_errno, expect;
+    if (use_rocm) {
+        papi_errno = rocd_init_environment();
+        if (papi_errno != PAPI_OK) {
+            _rocm_vector.cmp_info.initialized = 1;
+            _rocm_vector.cmp_info.disabled = papi_errno;
+            const char *err_string;
+            rocd_err_get_last(&err_string);
+            expect = snprintf(_rocm_vector.cmp_info.disabled_reason,
+                              PAPI_HUGE_STR_LEN, "%s", err_string);
+            if (expect < 0 || expect >= PAPI_HUGE_STR_LEN) {
+                SUBDBG("disabled_reason truncated");
+            }
+            goto fn_fail;
+        }
+
+        expect = snprintf(_rocm_vector.cmp_info.disabled_reason, PAPI_HUGE_STR_LEN, "%s",
+                         "Not initialized. Access component events to initialize it.");
+        if (expect < 0 || expect >= PAPI_HUGE_STR_LEN) {
             SUBDBG("disabled_reason truncated");
         }
-        goto fn_fail;
+        papi_errno = PAPI_EDELAY_INIT;
+        _rocm_vector.cmp_info.disabled = papi_errno;
+    } else {
+        expect = snprintf(_rocm_vector.cmp_info.disabled_reason, PAPI_HUGE_STR_LEN, "%s",
+                          "Not active while rocp_sdk component is active. Set 'export PAPI_DISABLE_COMPONENTS=rocp_sdk' to override.");
+        if (expect < 0 || expect >= PAPI_HUGE_STR_LEN) {
+            SUBDBG("disabled_reason truncated");
+        }
+        papi_errno = PAPI_ECOMBO;
+        _rocm_vector.cmp_info.disabled = papi_errno;
     }
-
-    sprintf(_rocm_vector.cmp_info.disabled_reason,
-            "Not initialized. Access component events to initialize it.");
-    papi_errno = PAPI_EDELAY_INIT;
-    _rocm_vector.cmp_info.disabled = papi_errno;
 
   fn_exit:
     SUBDBG("EXIT: %s\n", PAPI_strerror(papi_errno));
@@ -209,8 +242,8 @@ rocm_init_private(void)
         const char *err_string;
         rocd_err_get_last(&err_string);
         int expect = snprintf(_rocm_vector.cmp_info.disabled_reason,
-                              PAPI_MAX_STR_LEN, "%s", err_string);
-        if (expect > PAPI_MAX_STR_LEN) {
+                              PAPI_HUGE_STR_LEN, "%s", err_string);
+        if (expect > PAPI_HUGE_STR_LEN) {
             SUBDBG("disabled_reason truncated");
         }
 
@@ -222,8 +255,8 @@ rocm_init_private(void)
     _rocm_vector.cmp_info.num_native_events = count;
     _rocm_vector.cmp_info.num_cntrs = count;
     _rocm_vector.cmp_info.initialized = 1;
-    int strLen = snprintf(_rocm_vector.cmp_info.disabled_reason, PAPI_MAX_STR_LEN, "%s", "");
-    if (strLen < 0 || strLen >= PAPI_MAX_STR_LEN) {
+    int strLen = snprintf(_rocm_vector.cmp_info.disabled_reason, PAPI_HUGE_STR_LEN, "%s", "");
+    if (strLen < 0 || strLen >= PAPI_HUGE_STR_LEN) {
         SUBDBG("Failed to fully write disabled_reason.\n");
     }
 
--- a/src/components/rocm_smi/linux-rocm-smi.c
+++ b/src/components/rocm_smi/linux-rocm-smi.c
@@ -68,11 +68,45 @@ _rocm_smi_init_component(int cidx)
     _rocm_smi_vector.cmp_info.num_cntrs = -1;
     _rocm_smi_lock = PAPI_NUM_LOCK + NUM_INNER_LOCK + cidx;
 
-    sprintf(_rocm_smi_vector.cmp_info.disabled_reason,
-            "Not initialized. Access component events to initialize it.");
-    _rocm_smi_vector.cmp_info.disabled = PAPI_EDELAY_INIT;
+    /* Manage contension between rocm_smi and amd_smi components. */
+    int use_rocm_smi = 0;
+    #if defined(DEFAULT_TO_ROCM_SMI)
+    use_rocm_smi = 1;
+    #endif
+    #if defined(DEFAULT_TO_AMD_SMI)
+        char *disabledComps = getenv("PAPI_DISABLE_COMPONENTS");
+        if (disabledComps != NULL) {
+            char *penv = strdup(disabledComps);
+            char *p;
+            for (p = strtok (penv, ",:"); p != NULL; p = strtok (NULL, ",:")) {
+                if(!strcmp(p, "amd_smi")) use_rocm_smi = 1;
+            }
+            free(penv);
+        } else {
+            SUBDBG("rocm_smi: getenv(PAPI_DISABLE_COMPONENTS) was not set.\n");
+        }
+    #endif
 
-    return PAPI_EDELAY_INIT;
+    int papi_errno, expect;
+    if (use_rocm_smi) {
+        expect = snprintf(_rocm_smi_vector.cmp_info.disabled_reason, PAPI_HUGE_STR_LEN, "%s",
+                          "Not initialized. Access component events to initialize it.");
+        if (expect < 0 || expect >= PAPI_HUGE_STR_LEN) {
+            SUBDBG("disabled_reason truncated");
+        }
+        papi_errno = PAPI_EDELAY_INIT;
+        _rocm_smi_vector.cmp_info.disabled = papi_errno;
+        return papi_errno;
+    } else {
+        expect = snprintf(_rocm_smi_vector.cmp_info.disabled_reason, PAPI_HUGE_STR_LEN, "%s",
+                          "Not active while amd_smi component is active. Set 'export PAPI_DISABLE_COMPONENTS=amd_smi' to override.");
+        if (expect < 0 || expect >= PAPI_HUGE_STR_LEN) {
+            SUBDBG("disabled_reason truncated");
+        }
+        papi_errno = PAPI_ECOMBO;
+        _rocm_smi_vector.cmp_info.disabled = papi_errno;
+        return papi_errno;
+    }
 }
 
 static int
@@ -108,7 +142,10 @@ _rocm_smi_init_private(void)
         _rocm_smi_vector.cmp_info.disabled = papi_errno;
         const char *error_str;
         rocs_err_get_last(&error_str);
-        sprintf(_rocm_smi_vector.cmp_info.disabled_reason, "%s", error_str);
+        int expect = snprintf(_rocm_smi_vector.cmp_info.disabled_reason, PAPI_HUGE_STR_LEN, "%s", error_str);
+        if (expect < 0 || expect >= PAPI_HUGE_STR_LEN) {
+            SUBDBG("disabled_reason truncated");
+        }
         goto fn_fail;
     }
 
@@ -125,6 +162,12 @@ _rocm_smi_init_private(void)
 
   fn_exit:
     _rocm_smi_vector.cmp_info.disabled = papi_errno;
+    if(PAPI_OK == papi_errno) {
+        int expect = snprintf(_rocm_smi_vector.cmp_info.disabled_reason, PAPI_HUGE_STR_LEN, "%s", "");
+        if (expect < 0 || expect >= PAPI_HUGE_STR_LEN) {
+            SUBDBG("disabled_reason truncated");
+        }
+    }
     PAPI_unlock(COMPONENT_LOCK);
     return papi_errno;
   fn_fail:
--- a/src/components/rocp_sdk/rocp_sdk.c
+++ b/src/components/rocp_sdk/rocp_sdk.c
@@ -133,28 +133,57 @@ rocp_sdk_init_component(int cid)
     _rocp_sdk_vector.cmp_info.num_cntrs = -1;
     _rocp_sdk_lock = PAPI_NUM_LOCK + NUM_INNER_LOCK + cid;
 
-    // We set this env variable to silence some unnecessary ROCprofiler-SDK debug messages.
-    // It is not critical, so if it fails to be set, we can safely ignore the error.
-    (void)setenv("ROCPROFILER_LOG_LEVEL","fatal",0);
+    /* Manage contension between rocm and rocp_sdk components. */
+    int use_rocp_sdk = 0;
+    #if defined(DEFAULT_TO_ROCP_SDK)
+    use_rocp_sdk = 1;
+    #endif
+    #if defined(DEFAULT_TO_ROCM)
+        char *disabledComps = getenv("PAPI_DISABLE_COMPONENTS");
+        if (disabledComps != NULL) {
+            char *penv = strdup(disabledComps);
+            char *p;
+            for (p = strtok (penv, ",:"); p != NULL; p = strtok (NULL, ",:")) {
+                if(!strcmp(p, "rocm")) use_rocp_sdk = 1;
+            }
+            free(penv);
+        } else {
+            SUBDBG("rocp_sdk: getenv(PAPI_DISABLE_COMPONENTS) was not set.\n");
+        }
+    #endif
 
-    int papi_errno = rocprofiler_sdk_init_pre();
-    if (papi_errno != PAPI_OK) {
-        _rocp_sdk_vector.cmp_info.initialized = 0;
+    int papi_errno, expect;
+    if( use_rocp_sdk) {
+        // We set this env variable to silence some unnecessary ROCprofiler-SDK debug messages.
+        // It is not critical, so if it fails to be set, we can safely ignore the error.
+        (void)setenv("ROCPROFILER_LOG_LEVEL","fatal",0);
+
+        papi_errno = rocprofiler_sdk_init_pre();
+        if (papi_errno != PAPI_OK) {
+            _rocp_sdk_vector.cmp_info.initialized = 1;
+            _rocp_sdk_vector.cmp_info.disabled = papi_errno;
+            const char *err_string;
+            rocprofiler_sdk_err_get_last(&err_string);
+            expect = snprintf(_rocp_sdk_vector.cmp_info.disabled_reason, PAPI_HUGE_STR_LEN, "%s", err_string);
+            if (expect < 0 || expect >= PAPI_HUGE_STR_LEN) {
+                SUBDBG("disabled_reason truncated");
+            }
+            return papi_errno;
+        }
+
+        // This component needs to be fully initialized from the beginning,
+        // because interleaving hip calls and PAPI calls leads to errors.
+        return check_n_initialize();
+    } else {
+        expect = snprintf(_rocp_sdk_vector.cmp_info.disabled_reason, PAPI_HUGE_STR_LEN, "%s",
+                          "Not active while rocm component is active. Set 'export PAPI_DISABLE_COMPONENTS=rocm' to override.");
+        if (expect < 0 || expect >= PAPI_HUGE_STR_LEN) {
+            SUBDBG("disabled_reason truncated");
+        }
+        papi_errno = PAPI_ECOMBO;
         _rocp_sdk_vector.cmp_info.disabled = papi_errno;
-        const char *err_string;
-        rocprofiler_sdk_err_get_last(&err_string);
-        snprintf(_rocp_sdk_vector.cmp_info.disabled_reason, PAPI_MAX_STR_LEN, "%s", err_string);
         return papi_errno;
     }
-
-    int strLen = snprintf(_rocp_sdk_vector.cmp_info.disabled_reason, PAPI_HUGE_STR_LEN, "%s", \
-                          "Not initialized. Access component events to initialize it.");
-    if (strLen < 0 || strLen >= PAPI_HUGE_STR_LEN) {
-        SUBDBG("Failed to fully write initial disabled_reason.\n");
-        return PAPI_EBUF;
-    }
-    _rocp_sdk_vector.cmp_info.disabled = PAPI_EDELAY_INIT;
-    return PAPI_EDELAY_INIT;
 }
 
 int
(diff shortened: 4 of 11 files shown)


src/configure
```diff
@@ -7295,6 +7159,165 @@ for comp in $components; do
   fi
 done
 
+# Select the components that match the available ROCm stack. amd_smi
+# should be used for ROCm >= 6.4.0, while rocm_smi remains the fallback
+# for older releases. rocp_sdk should be used for ROCm >= 6.2.1, while
```
Treece-Burgess: rocp_sdk should be used for ROCm >= 6.3.2. As it has known bugs before this version.

src/configure
```diff
@@ -7295,6 +7159,165 @@ for comp in $components; do
...
+        if test "x$selected_roc" = "xrocp_sdk"; then
+            CFLAGS="$CFLAGS -DDEFAULT_TO_ROCP_SDK"
+        else
+            CFLAGS="$CFLAGS -DDEFAULT_TO_ROCM"
+        fi
+    fi
+fi
+
+if (test $rocm_found -eq 1 && test $rocp_sdk_found -eq 1); then
+  echo "WARNING: Components rocm and rocp_sdk will not be simultaneously active. See components/rocm/README.md for more details."
```
Treece-Burgess: If you are going to keep this warning message and the below warning message. I would output them at the very bottom of the configure output. As of right now they output in the middle:
```
checking for components/sysdetect/papi_sysdetect_presets.h... no
WARNING: Components rocm and rocp_sdk will not be simultaneously active. See components/rocm/README.md for more details.
```

It will be easier for a user to see the Warning at the end rather than in the middle of configure output as it is fairly quick.

src/configure
```diff
@@ -7295,6 +7159,165 @@ for comp in $components; do
...
+
+    rocm_version_clean=`echo "$rocm_version" | sed -n 's/[^0-9]*\([0-9][0-9]*\.[0-9][0-9]*\(\.[0-9][0-9]*\)\?\).*/\1/p'`
+    if test "x$rocm_version_clean" = "x"; then
+        rocm_version_clean=`echo "$rocm_version" | tr -cd '0-9.'`
+    fi
+
+   if test $amd_smi_found -eq 1 && test $rocm_smi_found -eq 1; then
+        have_amd_smi_lib=0
+        have_rocm_smi_lib=0
+        for root in "$PAPI_AMDSMI_ROOT" "$PAPI_ROCM_ROOT" "PAPI_ROCP_SDK_ROOT"
```
Treece-Burgess: If we are checking for `libamd_smi.so` then I think we should only be checking `PAPI_AMDSMI_ROOT` only. The README's outline the need for a component's environment variable.

We want to be as flexible as possible, but we also need to make sure we are following the guidelines that we have already set in place.

src/configure
```diff
@@ -7295,6 +7159,165 @@ for comp in $components; do
...
+            if test "x$root" != "x"; then
+                if test -f "$root/lib/libamd_smi.so" || test -f "$root/lib64/libamd_smi.so"; then
+                    have_amd_smi_lib=1
+                fi
+            fi
+        done
+        for root in "$PAPI_ROCMSMI_ROOT" "$PAPI_ROCM_ROOT" "$PAPI_ROCP_SDK_ROOT"
+        do
+            if test "x$root" != "x"; then
+                if test -f "$root/lib/librocm_smi64.so" || test -f "$root/lib64/librocm_smi64.so"; then
```
Treece-Burgess: Same as above.

src/configure
```diff
@@ -7295,6 +7159,165 @@ for comp in $components; do
...
+            fi
+            rocm_version_value=$((10#$rocm_major * 10000 + 10#$rocm_minor * 100 + 10#$rocm_patch))
+            if test $rocm_version_value -lt 60400; then
+                selected_smi="rocm_smi"
+            fi
+        else
+            if test $have_amd_smi_lib -eq 0 && test $have_rocm_smi_lib -eq 1; then
+                selected_smi="rocm_smi"
+            elif test $have_amd_smi_lib -eq 0 && test $have_rocm_smi_lib -eq 0; then
+                echo "WARNING: Unable to determine ROCm version or locate AMD/ROCm SMI libraries. Defaulting to amd_smi when both components are requested."
```
Treece-Burgess: I would consider removing the echo's here and below. My reasons are:
1. It clogs up the configure output
2. The messages will get lost in the middle of the output instead of the end

I think this information would be better suited to be put in the component's README and in the component's  disabled message.

src/configure
```diff
@@ -7295,6 +7159,165 @@ for comp in $components; do
...
+        elif command -v rocminfo >/dev/null 2>&1; then
+            rocm_version=`rocminfo --version 2>/dev/null | head -n 1`
+        elif command -v amd-smi >/dev/null 2>&1; then
+            rocm_version=`amd-smi --version 2>/dev/null | head -n 1`
+        elif command -v rocm-smi >/dev/null 2>&1; then
+            rocm_version=`rocm-smi --version 2>/dev/null | head -n 1`
+        fi
+    fi
+
+    rocm_version_clean=`echo "$rocm_version" | sed -n 's/[^0-9]*\([0-9][0-9]*\.[0-9][0-9]*\(\.[0-9][0-9]*\)\?\).*/\1/p'`
```
Treece-Burgess, line 6918: This sed command is extremely robust, is this needed due to the above `hipconfig --rocm-version`?

If so, why not default to checking for the `rocm_version.h` file which has the defines for `ROCM_VERSION_MAJOR`, `ROCM_VERSION_MINOR`, and `ROCM_VERSION_PATCH`. You could just use:
```
find /opt/rocm -name rocm_version.h
```

This would just return the path to the file and then you can subsequently grab the versions.

src/papi.c
```diff
@@ -1049,6 +1049,16 @@ PAPI_library_init( int version )
 
 	int tmp = 0;
 
+    char *penv = getenv("PAPI_DISABLE_COMPONENTS");
```
Treece-Burgess: Both `putenv` and `PAPI_disable_component` would need to be set before `PAPI_library_init(...)`; however, what if for some reason I want to disable a component at a certain point in my script. For amd_smi and rocm_smi, maybe this is due to a specific device in use. The current construction would not allow for this.

Handling it at the component level would and I think could be a beneficial addition for users.

src/configure
```diff
@@ -7295,6 +7159,165 @@ for comp in $components; do
...
+fi
+if (test $rocm_smi_found -eq 1 && test $amd_smi_found -eq 0); then
+    CFLAGS="$CFLAGS -DDEFAULT_TO_ROCM_SMI"
+fi
+
+if (test $rocm_found -eq 1 && test $rocp_sdk_found -eq 1) || \
+   (test $amd_smi_found -eq 1 && test $rocm_smi_found -eq 1); then
+    rocm_version=""
+
+    for root in "$PAPI_AMDSMI_ROOT" "$PAPI_ROCMSMI_ROOT" "$PAPI_ROCM_ROOT" "$PAPI_ROCP_SDK_ROOT"
```
Treece-Burgess: Q: What if the environment variable ROOT's have different paths to different versions?

src/configure
```diff
@@ -1109,15 +1107,6 @@ do
   | -silent | --silent | --silen | --sile | --sil)
     silent=yes ;;
 
-  -runstatedir | --runstatedir | --runstatedi | --runstated \
-  | --runstate | --runstat | --runsta | --runst | --runs \
```
Treece-Burgess, line 1112: Using autoconf-archive on Methane should be able to make it such that the `runstatedir` values appear.

src/components/rocm_smi/linux-rocm-smi.c
```diff
@@ -68,11 +68,43 @@ _rocm_smi_init_component(int cidx)
...
+    #endif
+    #if defined(DEFAULT_TO_AMD_SMI)
+        char *penv = getenv("PAPI_DISABLE_COMPONENTS");
+        if (penv != NULL) {
+            char *p;
+            for (p = strtok (penv, ",:"); p != NULL; p = strtok (NULL, ",:")) {
+                if(!strcmp(p, "amd_smi")) use_rocm_smi = 1;
+            }
+        } else {
+            SUBDBG("rocm: getenv(PAPI_DISABLE_COMPONENTS) failed.\n");
```
Treece-Burgess: You are in `rocm_smi`.

src/components/amd_smi/linux-amd-smi.c
```diff
@@ -57,11 +57,38 @@ static int _amd_smi_init_component(int cidx) {
...
+    #endif
+    #if defined(DEFAULT_TO_ROCM_SMI)
+        char *penv = getenv("PAPI_DISABLE_COMPONENTS");
+        if (penv != NULL) {
+            char *p;
+            for (p = strtok (penv, ",:"); p != NULL; p = strtok (NULL, ",:")) {
+                if(!strcmp(p, "rocm_smi")) use_amd_smi = 1;
+            }
+        } else {
+            SUBDBG("rocm: getenv(PAPI_DISABLE_COMPONENTS) failed.\n");
```
Treece-Burgess: You are in `amd_smi.

src/components/rocp_sdk/rocp_sdk.c
```diff
@@ -133,23 +133,55 @@ rocp_sdk_init_component(int cid)
...
+    #endif
+    #if defined(DEFAULT_TO_ROCM)
+        char *penv = getenv("PAPI_DISABLE_COMPONENTS");
+        if (penv != NULL) {
+            char *p;
+            for (p = strtok (penv, ",:"); p != NULL; p = strtok (NULL, ",:")) {
+                if(!strcmp(p, "rocm")) use_rocp_sdk = 1;
+            }
+        } else {
+            SUBDBG("rocm: getenv(PAPI_DISABLE_COMPONENTS) failed.\n");
```
Treece-Burgess: You are in `rocp_sdk`.

src/components/amd_smi/linux-amd-smi.c
```diff
@@ -57,11 +57,38 @@ static int _amd_smi_init_component(int cidx) {
...
+    int papi_errno;
+    if (use_amd_smi) {
+        CHECK_SNPRINTF(_amd_smi_vector.cmp_info.disabled_reason, PAPI_MAX_STR_LEN,
+                 "Not initialized. Access an AMD SMI event to initialize.");
+        papi_errno = PAPI_EDELAY_INIT;
+        _amd_smi_vector.cmp_info.disabled = papi_errno;
+        return papi_errno;
+    } else {
+        CHECK_SNPRINTF(_amd_smi_vector.cmp_info.disabled_reason, PAPI_MAX_STR_LEN,
+                 "Not active while rocm_smi component is active. Set PAPI_DISABLE_COMPONENTS=amd_smi to override.");
```
Treece-Burgess: nit: May want to say "Set 'export PAPI_DISABLE_COMPONENTS=amd_smi'" to override. Likewise with the others.

Also, I believe this should be rocm_smi instead of amd_smi.

src/components/amd_smi/linux-amd-smi.c
```diff
@@ -57,11 +57,38 @@ static int _amd_smi_init_component(int cidx) {
...
+                if(!strcmp(p, "rocm_smi")) use_amd_smi = 1;
+            }
+        } else {
+            SUBDBG("rocm: getenv(PAPI_DISABLE_COMPONENTS) failed.\n");
+        }
+    #endif
+
+    int papi_errno;
+    if (use_amd_smi) {
+        CHECK_SNPRINTF(_amd_smi_vector.cmp_info.disabled_reason, PAPI_MAX_STR_LEN,
```
Treece-Burgess: The `disabled_reason` can be up to a size of `PAPI_HUGE_STR_LEN`, I would suggest changing `PAPI_MAX_STR_LEN` to it here and in other cases.

src/components/amd_smi/linux-amd-smi.c
```diff
@@ -57,11 +57,38 @@ static int _amd_smi_init_component(int cidx) {
     _amd_smi_vector.cmp_info.num_mpx_cntrs = -1;
     _amd_smi_lock = PAPI_NUM_LOCK + NUM_INNER_LOCK + cidx;
 
-    CHECK_SNPRINTF(_amd_smi_vector.cmp_info.disabled_reason, PAPI_MAX_STR_LEN,
-             "Not initialized. Access an AMD SMI event to initialize.");
-    _amd_smi_vector.cmp_info.disabled = PAPI_EDELAY_INIT;
+    /* Manage contension between rocm_smi and amd_smi components. */
```
Treece-Burgess, line 60: Something to note is that if we compile `amd_smi` and `rocm_smi` with ROCm 6.3.4 then `amd_smi` causes a compilation failure:
```
components/amd_smi/amds.c:1049:62: error: ‘AMDSMI_TEMP_SHUTDOWN’ undeclared (first use in this function); did you mean ‘AMDSMI_TEMP_CRIT_MIN’?
       AMDSMI_TEMP_LOWEST,         AMDSMI_TEMP_HIGHEST,       AMDSMI_TEMP_SHUTDOWN};
                                                              ^~~~~~~~~~~~~~~~~~~~
                                                              AMDSMI_TEMP_CRIT_MIN

```

This is out of the range of support for `amd_smi`; however, we state that we default to `rocm_smi` in versions <= ROCm 6.4.0.

src/components/rocm/rocm.c
```diff
@@ -133,25 +133,56 @@ rocm_init_component(int cid)
...
     SUBDBG("ENTER: cid: %d\n", cid);
 
-    int papi_errno = rocd_init_environment();
-    if (papi_errno != PAPI_OK) {
-        _rocm_vector.cmp_info.initialized = 1;
-        _rocm_vector.cmp_info.disabled = papi_errno;
-        const char *err_string;
-        rocd_err_get_last(&err_string);
-        int expect = snprintf(_rocm_vector.cmp_info.disabled_reason,
+    /* Manage contension between rocm and rocp_sdk components. */
```
Treece-Burgess, line 136: When on Odyssey, with ROCm 6.4.4 the `amd_smi` and `rocp_sdk` are initially active. If I set `export PAPI_DISABLE_COMPONENTS="amd_smi,rocp_sdk"`, both components will be disabled, but only `rocm_smi` will become active. `rocm` still shows as disabled.

src/components/rocm_smi/linux-rocm-smi.c
```diff
@@ -68,11 +68,43 @@ _rocm_smi_init_component(int cidx)
...
+        }
+    #endif
+
+    int papi_errno, expect;
+    if (use_rocm_smi) {
+        expect = snprintf(_rocm_smi_vector.cmp_info.disabled_reason, PAPI_MAX_STR_LEN, "%s",
+                          "Not initialized. Access component events to initialize it.");
```
Treece-Burgess, line 93: After using `export PAPI_DISABLE_COMPONENTS="amd_smi,rocp_sdk"`, I attempt to get the component info for `rocm_smi`. To initialize it I use, `PAPI_enum_cmp_event` and the value of `num_native_events` changes from -1 to the number of events on the system. However, the disabled message still shows, "Not initialized. Access component events to initialize it".

Treece-Burgess (changes requested):

src/components/amd_smi/linux-amd-smi.c
dbarry9: Thank you for finding this. Should this go in an Issue/separate PR?

src/components/amd_smi/linux-amd-smi.c
dbarry9: This will be addressed in the next push.

src/components/amd_smi/linux-amd-smi.c
dbarry9: This will be addressed in the next push.

src/components/rocm_smi/linux-rocm-smi.c
dbarry9: Would the solution be for me to set the disabled reason to an empty string?

src/configure
dbarry9: Thank you. I will try this.

src/configure
dbarry9: Good question. I will defer this to @djwoun.

src/configure
dbarry9: Also pinging @djwoun on this one.

src/configure
dbarry9: @djwoun

src/configure
dbarry9: @djwoun

src/papi.c
dbarry9: > what if for some reason I want to disable a component at a certain point in my script.

Can't `PAPI_disable_component` be called at any time?

src/configure
dbarry9: I thought about this, too. On the one hand, I realize that this PR is placing a cut-off for which version should be used by default, and we should make the cut-off a useful and reasonable one (_i.e.,_ setting it past the versions with known bugs). On the other hand, because user can currently configure `rocp_sdk` in their build (while using ROCm 6.3.2) in the master branch and because technically the ROCProfiler-SDK is also shipped with versions between 6.2.1 and 6.3.2, despite them containing bugs, it could be placing an extra barrier for a user to utilize the version of ROCm they want. I suppose it doesn't matter too much, as the user can override whatever default we have in place.

src/components/rocm/rocm.c
dbarry9: Thank you for catching this. I am looking into it.

src/papi.c
Treece-Burgess: From my testing and per the function description no:
> * PAPI_disable_component() must be called before PAPI_library_init().

src/components/amd_smi/linux-amd-smi.c
Treece-Burgess: To actually make it to runtime with a ROCm version lower than 6.4.0 we would have to put if def's in a few locations. So yes, I would say it should be its own separate PR.

src/components/rocm_smi/linux-rocm-smi.c
Treece-Burgess: Yes! It just needs to be an empty string as at that point it has been initialized!

src/configure
Treece-Burgess: The issue I see with `rocp_sdk` having a default below 6.3.2 is that it invites the user to run into these known bugs that Anthony originally ran into himself. These bugs could be completely out of our control to fix as well by them needing to be fixed in a upstream ROCm version. Lastly, I believe ROCprofiler-SDK was a beta version until ROCm 7.0 as it finally jumped to 1.0.

src/configure
djwoun: That whole block is just a fallback chain to populate rocm_version when the version file probes don’t find anything. The sed/tr is only there to normalize whatever those tools print.

src/configure
djwoun: Agreed, the component  expect the component specific env vars, so checking only PAPI_AMDSMI_ROOT (and PAPI_ROCMSMI_ROOT for rocm_smi) is reasonable. I added PAPI_ROCM_ROOT as a convenience/flexibility fallback.

src/configure
djwoun: Fair point, I added the messages so users could tell why one component got dropped when both were requested, but I agree they’re noisy/mid-output. Putting it in the component disabled reason message is probably the best/most clear way for a user to know. I am slightly unsure how we would modify .cmp_info.disabled_reason tough.

src/papi.c
dbarry9: I understand. I will loop back to this.

src/configure
dbarry9: That makes sense to me. I will make the change.

src/components/rocm/rocm.c
dbarry9: The latest push should resolve this issue; however, please note that I am still addressing your other guidance in this pull request, so I will ping you once it is ready for re-review.

src/papi.c
dbarry9: Just so I understand properly: If the user is running a script, then they could change the environment variable between application runs, right? Or do you mean enable/disable components in the context of a single application?

sbkathpe (changes requested):
Yes, I believe this follows the spirit and intent of this env variable.

I would do a `free(penv)` though since this allocated space is no longer needed.

src/papi.c
```diff
@@ -1049,6 +1049,17 @@ PAPI_library_init( int version )
...
 
+    char *disabledComps = getenv("PAPI_DISABLE_COMPONENTS");
+    if (disabledComps != NULL) {
+        char *penv = strdup(disabledComps);
+        char *p;
+        for (p = strtok (penv, ",:"); p != NULL; p = strtok (NULL, ",:")) {
+            (void) PAPI_disable_component_by_name(p);
+        }
+    } else {
+        SUBDBG("PAPI_library_init: getenv(PAPI_DISABLE_COMPONENTS) failed.\n");
```
sbkathpe: Is this debug statement required or necessary? The message saying it "failed" seems misleading since it is just fine if it is NOT set in the environment. Perhaps change "failed" with "was not set" or similar or follow any prior art for issuing a `SUBDBG` for an unset environment variable.

src/papi.c
```diff
@@ -1049,6 +1049,17 @@ PAPI_library_init( int version )
 
 	int tmp = 0;
 
+    char *disabledComps = getenv("PAPI_DISABLE_COMPONENTS");
+    if (disabledComps != NULL) {
+        char *penv = strdup(disabledComps);
+        char *p;
+        for (p = strtok (penv, ",:"); p != NULL; p = strtok (NULL, ",:")) {
+            (void) PAPI_disable_component_by_name(p);
```
sbkathpe, line 1057: Add a `free(penv)`.

src/configure
dbarry9: Done.

src/configure
dbarry9: Done.

src/components/amd_smi/linux-amd-smi.c
dbarry9: Are the other cases you refer to only in the amd_smi component, or does this apply to the changes I have made to the other AMD components, as well?

src/papi.c
dbarry9: I have now added it.

src/papi.c
dbarry9: You are right; it is not truly an error. I have updated it to say "was not set."

src/components/amd_smi/linux-amd-smi.c
Treece-Burgess: Yes, they all currently use `PAPI_MAX_STR_LEN`, but I would update all to use and check for `PAPI_HUGE_STR_LEN` as that is the max room you can have with `.disabled_reason`. I think it is safer to avoid someone making a change later down the road that has > `PAPI_MAX_STR_LEN` as they see `.disabled_reason` supports `PAPI_HUGE_STR_LEN`.

src/components/rocm_smi/linux-rocm-smi.c
dbarry9: This should be fixed now.

src/papi.c
Treece-Burgess: I mean in the context of a single application. An example workflow would be like:

```
// Example configure would be: ./configure --prefix=$PWD/test-install --with-components="amd_smi rocm_smi"
// Example machine would have mixed devices with MI210 and MI300A
int main()
{
    // rocm_smi is initially disabled and I go to profile the MI300A
    PAPI_library_init();

    // do some sort of PAPI work here

   // now I want to profile the MI210, so I attempt to do
   putenv("PAPI_DISABLE_COMPONENTS="amd_smi");
   // However, this will not work as you have to make this change before PAPI_library_init() is called
}
```

Do you understand the workflow I mean now? If not I will ping you on slack.

I think overall this could be useful to a user, but I can also see an argument against it. Really you would have to disable the component at the component level.

dbarry9: I understand the workflow you mean; I will think it over. Thank you!

src/components/amd_smi/linux-amd-smi.c
dbarry9: I have now updated these to `PAPI_HUGE_STR_LEN`.

src/configure
Treece-Burgess: Okay great, then we will move forward for just checking the environment variable associated with the `.so` we are looking for.

src/configure
Treece-Burgess: Jun and I replied [here](https://github.com/icl-utk-edu/papi/pull/548#discussion_r2760757804).

src/configure
Treece-Burgess: I think handling all these checks at the component level would be ideal using a combination of the `Rules.component` files along with the source files `linux-component.c`. Keeping the checks here I do agree that having some of these message continue on to the component may not be possible.

However, I think we can cover all our bases using the component README's and `papi_component_avail`'s disabled message. 

My thought is either you update the AMD component's README limitations section or add a entirely new section that discusses the cases. In combination with this you could update the disabled messages to be the following:

```
// Note lets say my configure is ./configure --prefix=$PWD/test-install --with-components="amd_smi rocm_smi" and I used ROCm 6.4.4 which means amd_smi is used.

// The disabled message then for rocm_smi could be
The components rocm_smi and amd_smi have been configured together. Only one can be active at a given time. To make rocm_smi active, set 'export PAPI_DISABLE_COMPONENTS=amd_smi'. For more information see components/rocm_smi/README.md.
```

src/configure
Treece-Burgess: My suggestion would be to change the implementation to the less robust option that I believe searching and using the contents of `rocm_version.h` would be. However, if you prefer this option then we can move forward with it.

sbkathpe (commented):
Just to make sure, the "PAPI_DISABLE_COMPONENTS" code should NOT be compile-time dependent on "#if defined(DEFAULT_TO_ROCP_SDK)" but be unconditionally supported.
It is hard for me to tell if this condition is part of the final code or not.

src/papi.c
dbarry9: What would disabling the component at the component level look like? Would each component's init function parse `PAPI_DISABLE_COMPONENTS` for its own component's name?

src/configure
dbarry9: This is a really good point. Do you think it would be a good idea to determine the version for each component, based off of each component's `PAPI_*_ROOT` variable? I don't think that we should allow for different versions of ROCm to be used in a single PAPI build. It seems that this could cause issues during compilation and linking. At any rate, we could certainly check that the `PAPI_*_ROOT` variables ultimately all point to the same ROCm version number, and if they do not, then we can warn the user at the end of the configure.

src/configure
dbarry9: Can the component not perform an independent check for these library *.so's? In that case, we would not need to propagate the error/warning message all the way from configure, correct?

I agree that handling these checks at the component level would be ideal. If we go this route, however, I would like to keep the logic as-is for this PR, and we can raise a new issue to point out that this needs to be addressed in the component. Does that sound okay?

Edit to Add: I will see what I can do to make the output from the `echo`'s more appropriately located, and I will substitute them with `AC_MSG_WARN`.

src/configure
dbarry9: I have updated this accordingly.

src/configure
dbarry9: I have now changed the `echo`'s to `AC_MSG_WARN`'s and `AC_MSG_NOTICE`'s, as appropriate. They will now print at the end of the configuration. I can still update it if we decide to do something differently for this PR.

```
 ...
 config.status: creating config.h
 config.status: executing genpapifdef commands
 configure: Detected ROCm version 7.0.2. Enabling amd_smi by default.
 configure: Detected ROCm version 7.0.2. Enabling rocp_sdk by default.
 configure: WARNING: Components rocm and rocp_sdk will not be simultaneously active. See components/rocm/README.md for more details.
 configure: WARNING: Components rocm_smi and amd_smi will not be simultaneously active.
```

src/configure
dbarry9: @djwoun Do you agree with Treece's suggestion, or would you like to keep it the same for now/for this PR?

src/configure
dbarry9: This is done in the latest push.

src/configure
dbarry9: This is done in the latest push.

src/configure
dbarry9: I have updated the logic here to check for ROCm version mismatches among the `PAPI_*_ROOT` variables corresponding to the AMD components that are included in the configure command. If there is a version mismatch, then the last mismatch that was found gets reported to the user as a warning that this can lead to compilation errors. I think that, for now, it is reasonable to let the user proceed at their own risk.

```
 ...
 config.status: creating config.h
 config.status: executing genpapifdef commands
 configure: WARNING:
     ROCm version 7.0.2 in /opt/rocm-7.0.2 does not match version 6.4.2-120 in /opt/rocm-6.4.2.
     Using different ROCm versions could lead to compilation errors.
```

src/configure
Treece-Burgess: > Can the component not perform an independent check for these library *.so's? In that case, we would not need to propagate the error/warning message all the way from configure, correct?

The .so's should already be looked for at the component level. For example, see the function `obtain_function_pointers` in rocp_sdk.

Treece-Burgess: > I have now changed the echo's to AC_MSG_WARN's and AC_MSG_NOTICE's, as appropriate. They will now print at the end of the configuration. I can still update it if we decide to do something differently for this PR.

This looks better in the configure output. One thing I would suggest changing is the need to continue to read and condense the text. For example:
```
configure: Components rocm and rocp_sdk do not support being simultaneously configured. Enabling rocp_sdk by default due to ROCm 7.0.2. See x readme for more details.
```

Something along those lines, it is a longer sentence, but it is all right there for a user to read.

src/papi.c
Treece-Burgess: Yes, so how I would implement it would be to do the following:
1. As there does not appear to be an enumeration function for compiled in components, I would create one in PAPI_internal.c. In this function, I would collect all compiled in components.
2. At the component level I would get the values of `PAPI_DISABLE_COMPONENTS`, call the function I created in step 1, and then compare the values. If the function I was in, was in the list of `PAPI_DISABLE_COMPONENTS` then I would provide an error code and err string.

I have not personally tested this, but at a high level this is the workflow I would envision.

src/configure
Treece-Burgess: > This is a really good point. Do you think it would be a good idea to determine the version for each component, based off of each component's PAPI_*_ROOT variable? I don't think that we should allow for different versions of ROCm to be used in a single PAPI build. It seems that this could cause issues during compilation and linking. At any rate, we could certainly check that the PAPI_*_ROOT variables ultimately all point to the same ROCm version number, and if they do not, then we can warn the user at the end of the configure.

I do not think having different versions linked to different environment variables will lead to an issue at compilation.

> I have updated the logic here to check for ROCm version mismatches among the PAPI_*_ROOT variables corresponding to the AMD components that are included in the configure command. If there is a version mismatch, then the last mismatch that was found gets reported to the user as a warning that this can lead to compilation errors. I think that, for now, it is reasonable to let the user proceed at their own risk.

I cannot think of a better solution at this time, so I think this is the way forward for now.

src/configure
dbarry9: Okay, I have updated it to the following:
```
 ...
 config.status: creating config.h
 config.status: executing genpapifdef commands
 configure:
     Components rocm_smi and amd_smi cannot be simultaneously active. See components/amd_smi/README.md for details.
     Component amd_smi will be enabled by default due to ROCm version 7.0.2 being detected.
 configure:
     Components rocm and rocp_sdk cannot be simultaneously active. See components/rocp_sdk/README.md for details.
     Component rocp_sdk will be enabled by default due to ROCm version 7.0.2 being detected.
```

src/papi.c
dbarry9: > 2\. At the component level I would get the values of `PAPI_DISABLE_COMPONENTS`, call the function I created in step 1, and then compare the values. If the function I was in, was in the list of `PAPI_DISABLE_COMPONENTS` then I would provide an error code and err string.

Which function inside of the component would handle this? The reason I ask: If the component's init function does this, then wouldn't we have the same issue as we had in the sample workflow you posted?

src/configure
djwoun: Changing the output location and the changed message appears clear to me.

src/configure
djwoun: I agree with Treece’s direction, but I don’t think we should replace the current fallback chain with find /opt/rocm. /opt/rocm may not be the active root (modules/multiple installs), and some systems won’t have headers installed.

I suggest we add a rocm_version.h probe under the detected roots (e.g., $root/include/rocm_version.h / a couple common include paths). If found, parse ROCM_VERSION_{MAJOR,MINOR,PATCH} and use that. If not found, keep the existing CLI fallbacks + normalization.

src/papi.c
dbarry9: This seems like a corner case in the sense that it only affects a couple specific configurations (`rocm` _and_ `rocp_sdk`, `rocm_smi` _and_ `amd_smi`), and even then only given specific hardware configurations (having different AMD GPU/APU architectures in the same node), and even then only in the context of one execution of an application. I could be missing other cases, though.

As mentioned previously, disabling a component programmatically in the master branch of PAPI currently requires that `PAPI_disable_component` is called before `PAPI_library_init`. This PR's scope is not intended to extend this particular functionality (e.g., disabling components after `PAPI_library_init`); rather, its scope is to allow the existing functionality to be triggered automatically in the presence of `PAPI_DISABLE_COMPONENTS`.

I think your suggestion is the ideal approach to disabling components, but expanding the existing functionality in this manner is a large undertaking because each component would have to implement this mechanism. In my opinion, we should make this a separate Feature Request.

src/configure
dbarry9: I have now added this logic.

dbarry9:
Closing to create separate PRs #631 and #632.

PR #550, opened 2026-02-01
amd_smi: String description for each device for string meta data events

djwoun:
## Pull Request Description
Previously, the event description for string events for amd_smi in papi_native_avail only displayed the string for Device 0. 

Implemented per-device descriptions. Updated initialization to capture and store specific strings (UUIDs, serials, versions) for every device index. Modified amds_evt_code_to_info and amds_evt_code_to_descr to dynamically display the correct source string for the specific device variant being queried. Each :device=N variant now displays its own specific  string in the description, providing necessary context for the hash value.

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/components/amd_smi/amds.c
+++ b/src/components/amd_smi/amds.c
@@ -256,12 +256,20 @@ static void sanitize_name(const char *src, char *dst, size_t len) {
 }
 
 static int strip_device_qualifier(const char *name, char *base, size_t len) {
-  if (!name || !base || len == 0)
+  if (!name || !base || len == 0) {
+    SUBDBG("strip_device_qualifier: invalid arguments.\n");
     return PAPI_EINVAL;
+  }
   size_t nlen = strlen(name);
-  if (nlen >= len)
+  if (nlen >= len) {
+    SUBDBG("strip_device_qualifier: event name too long: '%s'\n", name);
     return PAPI_EBUF;
-  CHECK_SNPRINTF(base, len, "%s", name);
+  }
+  int strLen = snprintf(base, len, "%s", name);
+  if (strLen < 0 || (size_t)strLen >= len) {
+    SUBDBG("strip_device_qualifier: failed to fully write base event name.\n");
+    return PAPI_EBUF;
+  }
   char *qual = strstr(base, ":device=");
   if (qual) {
     char *next = strchr(qual + strlen(":device="), ':');
@@ -648,6 +656,21 @@ static int shutdown_event_table(void) {
   int i;
   for (i = 0; i < ntv_table.count; ++i) {
     htable_delete(htable, ntv_table.events[i].name);
+    if ((ntv_table.events[i].evtinfo_flags & AMDS_EVTINFO_FLAG_PER_DEVICE_DESCR) &&
+        ntv_table.events[i].per_device_descr) {
+      amds_per_device_descr_t *pd =
+          ntv_table.events[i].per_device_descr;
+      if (pd->descrs) {
+        int d;
+        for (d = 0; d < pd->num_devices; ++d) {
+          if (pd->descrs[d])
+            papi_free(pd->descrs[d]);
+        }
+        papi_free(pd->descrs);
+      }
+      papi_free(pd);
+      ntv_table.events[i].per_device_descr = NULL;
+    }
     papi_free(ntv_table.events[i].name);
     papi_free(ntv_table.events[i].descr);
   }
@@ -947,22 +970,30 @@ int amds_err_get_last(const char **err_string) {
 static int add_event(int *idx_ptr, const char *name, const char *descr, int device,
                      uint32_t variant, uint32_t subvariant, int mode,
                      amds_accessor_t access_func) {
-  if (!idx_ptr || !name || !descr)
+  if (!idx_ptr || !name || !descr) {
+    SUBDBG("add_event: invalid arguments.\n");
     return PAPI_EINVAL;
+  }
 
   char base_name[PAPI_MAX_STR_LEN];
   int papi_errno = strip_device_qualifier(name, base_name, sizeof(base_name));
-  if (papi_errno != PAPI_OK)
+  if (papi_errno != PAPI_OK) {
+    SUBDBG("add_event: failed to strip device qualifier for '%s'.\n", name);
     return papi_errno;
+  }
 
   native_event_t *existing = NULL;
   int hret = htable_find(htable, base_name, (void **)&existing);
   if (hret == HTABLE_SUCCESS) {
     papi_errno = amds_dev_set(&existing->device_map, device);
-    if (papi_errno != PAPI_OK)
+    if (papi_errno != PAPI_OK) {
+      SUBDBG("add_event: failed to set device bitmap for '%s' device=%d.\n",
+             base_name, device);
       return papi_errno;
+    }
     return PAPI_OK;
   } else if (hret != HTABLE_ENOVAL) {
+    SUBDBG("add_event: failed to find '%s' in the event hash table.\n", base_name);
     return PAPI_ECMP;
   }
 
@@ -970,17 +1001,24 @@ static int add_event(int *idx_ptr, const char *name, const char *descr, int devi
   ev->id = *idx_ptr;
   ev->name = strdup(base_name);
   ev->descr = strdup(descr);
-  if (!ev->name || !ev->descr)
+  if (!ev->name || !ev->descr) {
+    SUBDBG("add_event: failed to allocate memory for '%s'.\n", base_name);
     return PAPI_ENOMEM;
+  }
   ev->device = (device < 0) ? device : -1;
   ev->device_map = 0;
   papi_errno = amds_dev_set(&ev->device_map, device);
-  if (papi_errno != PAPI_OK)
+  if (papi_errno != PAPI_OK) {
+    SUBDBG("add_event: failed to set device bitmap for '%s' device=%d.\n",
+           base_name, device);
     return papi_errno;
+  }
   ev->value = 0;
   ev->mode = mode;
   ev->variant = variant;
   ev->subvariant = subvariant;
+  ev->evtinfo_flags = 0;
+  ev->per_device_descr = NULL;
   ev->priv = NULL;
   ev->open_func = open_simple;
   ev->close_func = close_simple;
@@ -992,20 +1030,108 @@ static int add_event(int *idx_ptr, const char *name, const char *descr, int devi
   return PAPI_OK;
 }
 
+static int add_event_with_perdev_descr(int *idx_ptr, const char *name,
+                                       const char *descr, int device,
+                                       uint32_t variant, uint32_t subvariant,
+                                       int mode, amds_accessor_t access_func,
+                                       uint32_t evtinfo_flags) {
+  if (!idx_ptr || !name || !descr) {
+    SUBDBG("add_event_with_perdev_descr: invalid arguments.\n");
+    return PAPI_EINVAL;
+  }
+
+  char base_name[PAPI_MAX_STR_LEN];
+  int papi_errno = strip_device_qualifier(name, base_name, sizeof(base_name));
+  if (papi_errno != PAPI_OK) {
+    SUBDBG("add_event_with_perdev_descr: failed to strip device qualifier for '%s'.\n",
+           name);
+    return papi_errno;
+  }
+
+  /* Create or expand the base event entry using the existing add_event(). */
+  papi_errno = add_event(idx_ptr, name, descr, device, variant, subvariant, mode,
+                         access_func);
+  if (papi_errno != PAPI_OK) {
+    SUBDBG("add_event_with_perdev_descr: add_event failed for '%s'.\n", base_name);
+    return papi_errno;
+  }
+
+  native_event_t *ev = NULL;
+  int hret = htable_find(htable, base_name, (void **)&ev);
+  if (hret != HTABLE_SUCCESS || !ev) {
+    SUBDBG("add_event_with_perdev_descr: failed to find '%s' in the event hash table.\n",
+           base_name);
+    return PAPI_ECMP;
+  }
+
+  ev->evtinfo_flags |= evtinfo_flags;
+
+  if (!(ev->evtinfo_flags & AMDS_EVTINFO_FLAG_PER_DEVICE_DESCR))
+    return PAPI_OK;
+
+  amds_per_device_descr_t *pd = ev->per_device_descr;
+  if (!pd) {
+    pd = (amds_per_device_descr_t *)papi_calloc(1, sizeof(*pd));
+    if (!pd) {
+      SUBDBG("add_event_with_perdev_descr: failed to allocate per-device description metadata for '%s'.\n",
+             base_name);
+      return PAPI_ENOMEM;
+    }
+    int limit = device_count;
+    if (limit <= 0 || limit > 64) {
+      SUBDBG("add_event_with_perdev_descr: unsupported device_count=%d for '%s'.\n",
+             limit, base_name);
+      papi_free(pd);
+      return PAPI_ENOSUPP;
+    }
+    pd->num_devices = limit;
+    pd->descrs = (char **)papi_calloc((size_t)limit, sizeof(*pd->descrs));
+    if (!pd->descrs) {
+      SUBDBG("add_event_with_perdev_descr: failed to allocate per-device description array for '%s'.\n",
+             base_name);
+      papi_free(pd);
+      return PAPI_ENOMEM;
+    }
+    ev->per_device_descr = pd;
+  }
+
+  if (device >= 0 && device < pd->num_devices) {
+    if (pd->descrs[device])
+      papi_free(pd->descrs[device]);
+    pd->descrs[device] = strdup(descr);
+    if (!pd->descrs[device]) {
+      SUBDBG("add_event_with_perdev_descr: failed to allocate per-device description for '%s' device=%d.\n",
+             base_name, device);
+      return PAPI_ENOMEM;
+    }
+  }
+
+  return PAPI_OK;
+}
+
 static int add_counter_event(int *idx_ptr, const char *name, const char *descr,
                              int device, uint32_t variant, uint32_t subvariant) {
   int papi_errno = add_event(idx_ptr, name, descr, device, variant, subvariant,
                              PAPI_MODE_READ, access_amdsmi_gpu_counter);
-  if (papi_errno != PAPI_OK)
+  if (papi_errno != PAPI_OK) {
+    SUBDBG("add_counter_event: add_event failed for '%s'.\n",
+           name ? name : "<null>");
     return papi_errno;
+  }
   char base_name[PAPI_MAX_STR_LEN];
   papi_errno = strip_device_qualifier(name, base_name, sizeof(base_name));
-  if (papi_errno != PAPI_OK)
+  if (papi_errno != PAPI_OK) {
+    SUBDBG("add_counter_event: failed to strip device qualifier for '%s'.\n",
+           name ? name : "<null>");
     return papi_errno;
+  }
   native_event_t *ev = NULL;
   int hret = htable_find(htable, base_name, (void **)&ev);
-  if (hret != HTABLE_SUCCESS || !ev)
+  if (hret != HTABLE_SUCCESS || !ev) {
+    SUBDBG("add_counter_event: failed to find '%s' in the event hash table.\n",
+           base_name);
     return PAPI_ECMP;
+  }
   ev->open_func = open_counter;
   ev->close_func = close_counter;
   ev->start_func = start_counter;
@@ -2263,7 +2389,8 @@ static int init_event_table(void) {
     }
   }
   /* GPU clock frequency levels for multiple clock domains */
-  for (int d = 0; d < gpu_count; ++d) {
+  int d;
+  for (d = 0; d < gpu_count; ++d) {
     const amdsmi_clk_type_t clk_types[] = {
         AMDSMI_CLK_TYPE_SYS,  AMDSMI_CLK_TYPE_GFX,  AMDSMI_CLK_TYPE_DF,
         AMDSMI_CLK_TYPE_DCEF, AMDSMI_CLK_TYPE_SOC,  AMDSMI_CLK_TYPE_MEM,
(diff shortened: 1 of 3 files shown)


src/components/amd_smi/amds.c
```diff
@@ -648,6 +648,20 @@ static int shutdown_event_table(void) {
   int i;
```
Treece-Burgess, line 656: Q: Output from `papi_native_avail` looks good on Odyssey. For the hash cases, I am seeing 4 values output in the description. However, some of the events just have the values repeated. For example, 
```
6410 --------------------------------------------------------------------------------
6411 | amd_smi:::board_product_name_hash                                            |
6412 |            Hash of Device 0 board product name string 'Aqua Vanjaram [Instinc|
6413 |            t MI300A]'                                                        |
6414 |            Hash of Device 1 board product name string 'Aqua Vanjaram [Instinc|
6415 |            t MI300A]'                                                        |
6416 |            Hash of Device 2 board product name string 'Aqua Vanjaram [Instinc|
6417 |            t MI300A]'                                                        |
6418 |            Hash of Device 3 board product name string 'Aqua Vanjaram [Instinc|
6419 |            t MI300A]'                                                        |
6420 |     :device=0                                                                |
6421 |            Mandatory device qualifier [0,1,2,3]                              |
6422 --------------------------------------------------------------------------------
```

Do you think it would be possible to condense these cases into a single description i.e. "Hash of Device 0,1,2,3 board product name string 'Aqua Vanjaram [Instinct MI300A]'"? I think this may over complicate the internal code a great deal, but wanted to get your thoughts.

src/components/amd_smi/amds.c
```diff
@@ -2668,15 +2741,15 @@ static int init_event_table(void) {
           uint32_t variant;
         } board_fields[] = {
             {binfo.product_serial, "board_serial_hash",
-             "Device %d board serial number hash of '%s'", 0},
+             "Hash of Device %d board serial number string '%s'", 0},
```
Treece-Burgess, line 2798: nit: I think this description could be clearer, something like "String representation for hash of device %d: '%s'". Likewise for the other cases below as well.

src/components/amd_smi/amds.c
```diff
@@ -648,6 +648,20 @@ static int shutdown_event_table(void) {
   int i;
   for (i = 0; i < ntv_table.count; ++i) {
     htable_delete(htable, ntv_table.events[i].name);
+    if ((ntv_table.events[i].evtinfo_flags & AMDS_EVTINFO_FLAG_PER_DEVICE_DESCR) &&
+        ntv_table.events[i].per_device_descr) {
+      amds_per_device_descr_t *pd =
+          ntv_table.events[i].per_device_descr;
+      if (pd->descrs) {
+        for (int d = 0; d < pd->num_devices; ++d) {
```
Treece-Burgess: Declare `d` outside of the for loop header.

src/components/amd_smi/amds.c
```diff
@@ -992,6 +1008,63 @@ static int add_event(int *idx_ptr, const char *name, const char *descr, int devi
...
+    return PAPI_OK;
+
+  amds_per_device_descr_t *pd = ev->per_device_descr;
+  if (!pd) {
+    pd = (amds_per_device_descr_t *)papi_calloc(1, sizeof(*pd));
+    if (!pd)
+      return PAPI_ENOMEM;
+    int limit = device_count;
+    if (limit <= 0 || limit > 64)
+      limit = 64;
```
Treece-Burgess: Q: If the value of `limit` goes below or above, why would this not be a strict failure instead of just hard setting it to be 64?

src/components/amd_smi/amds.c
```diff
@@ -992,6 +1008,63 @@ static int add_event(int *idx_ptr, const char *name, const char *descr, int devi
...
+    pd->descrs = (char **)papi_calloc((size_t)limit, sizeof(*pd->descrs));
+    if (!pd->descrs) {
+      papi_free(pd);
+      return PAPI_ENOMEM;
+    }
+    ev->per_device_descr = pd;
+  }
+
+  if (device >= 0 && device < pd->num_devices) {
+    if (pd->descrs[device])
```
Treece-Burgess, line 1099: Q: At line 1049, you allocate memory via `calloc` for the number of entries of `limit`. At this stage you free that block of memory and then reassign a new block of memory via `strdup`. That is correct, but why do you not just above do the `pd->descrs[device] = strdup(descr)` instead of allocating the block? That is not clear to me immediately.

src/components/amd_smi/amds.c
```diff
@@ -992,6 +1008,63 @@ static int add_event(int *idx_ptr, const char *name, const char *descr, int devi
...
+                                       const char *descr, int device,
+                                       uint32_t variant, uint32_t subvariant,
+                                       int mode, amds_accessor_t access_func,
+                                       uint32_t evtinfo_flags) {
+  if (!idx_ptr || !name || !descr)
+    return PAPI_EINVAL;
+
+  char base_name[PAPI_MAX_STR_LEN];
+  int papi_errno = strip_device_qualifier(name, base_name, sizeof(base_name));
+  if (papi_errno != PAPI_OK)
```
Treece-Burgess: Can you sprinkle SUBDBG's throughout the conditionals that will result in an error code being returned that is not `PAPI_OK`? It will be easier for debugging later if needed.

src/components/amd_smi/amds.c
```diff
@@ -992,6 +1008,63 @@ static int add_event(int *idx_ptr, const char *name, const char *descr, int devi
...
+
+  if (!(ev->evtinfo_flags & AMDS_EVTINFO_FLAG_PER_DEVICE_DESCR))
+    return PAPI_OK;
+
+  amds_per_device_descr_t *pd = ev->per_device_descr;
+  if (!pd) {
+    pd = (amds_per_device_descr_t *)papi_calloc(1, sizeof(*pd));
+    if (!pd)
+      return PAPI_ENOMEM;
+    int limit = device_count;
```
Treece-Burgess, line 1080: Q: Not pertaining to this PR, but what is the difference between device_count and gpu_count? You use device_count here, but gpu_count is what is used to enumerate through the for loops inside init_event_table.

src/components/amd_smi/amds_evtapi.c
```diff
@@ -87,6 +87,102 @@ static int device_next(uint64_t bitmap, int after) {
...
+   * string using fixed-width fields so tools like papi_native_avail fold each
+   * device descr onto its own output line.
+   */
+  if (!(event->evtinfo_flags & AMDS_EVTINFO_FLAG_PER_DEVICE_DESCR)) {
+    CHECK_SNPRINTF(buf, len, "%s", fallback);
+    return PAPI_OK;
+  }
+
+  if (!event->device_map) {
+    snprintf(buf, len, "%s", fallback);
```
Treece-Burgess: `snprintf` needs to have the return value checked otherwise you can run into compilation warnings. Why not use `CHECK_SNPRINTF`?

src/components/amd_smi/amds_evtapi.c
```diff
@@ -87,6 +87,102 @@ static int device_next(uint64_t bitmap, int after) {
...
+    return PAPI_OK;
+  }
+
+  size_t used = 0;
+  // Keep in sync with papi_native_avail's fixed-width folding (EVT_LINE - 12 - 2).
+  const size_t line_width = 80 - 12 - 2;
+
+  const amds_per_device_descr_t *pd = event->per_device_descr;
+  for (int dev = device_first(event->device_map); dev >= 0;) {
```
Treece-Burgess: Declare `dev` outside of the for loop header.

src/components/amd_smi/amds_evtapi.c
```diff
@@ -87,6 +87,102 @@ static int device_next(uint64_t bitmap, int after) {
...
+      buf[used] = '\0';
+      if (pad != line_width - rem)
+        break;
+    }
+
+    dev = next;
+  }
+
+  if (buf[0] == '\0')
+    snprintf(buf, len, "%s", fallback);
```
Treece-Burgess: `snprintf` needs to have the return value checked otherwise you can run into compilation warnings. Why not use `CHECK_SNPRINTF`?

src/components/amd_smi/amds_evtapi.c
```diff
@@ -87,6 +87,102 @@ static int device_next(uint64_t bitmap, int after) {
...
+    int next = device_next(event->device_map, dev);
+
+    const char *line = fallback;
+    if (pd && pd->descrs && pd->num_devices > 0 && dev < pd->num_devices &&
+        pd->descrs[dev]) {
+      line = pd->descrs[dev];
+    }
+
+    int strLen = snprintf(buf + used, len - used, "%s", line);
+    if (strLen < 0) {
```
Treece-Burgess: Q: Why add the null terminating character yourself? `snprintf` will add the null terminating character even if it will truncate the output as long as you do not pass a zero buffer size. Are you thinking it could be possible for `len - used` to result in the zero buffer?

I think it would be better to condense the two conditionals:
```
if (strLen < 0|| strLen >= len - used) {
    SUBDBG("Failed to fully fill the event description.\n")
}
```

src/components/amd_smi/amds_evtapi.c
```diff
@@ -87,6 +87,102 @@ static int device_next(uint64_t bitmap, int after) {
...
+    if ((size_t)strLen >= len - used) {
+      buf[len - 1] = '\0';
+      break;
+    }
+    used += (size_t)strLen;
+
+    if (next < 0)
+      break;
+
+    size_t rem = used % line_width;
```
Treece-Burgess, line 252: Q: Why do we need to fill out the line width? Is there an issue with the line width being shorter and if so do you have an example of the output for it you could post in response?

src/components/amd_smi/amds.c
djwoun: The event returns a numeric hash, while the description shows the input string it’s derived from. “String representation for hash …” sounds like the string is the hash value in string form, which isn’t the case here.

src/components/amd_smi/amds.c
djwoun: ack

src/components/amd_smi/amds.c
djwoun: you are completely right

src/components/amd_smi/amds.c
djwoun: pd->descrs = papi_calloc(limit, sizeof(*pd->descrs)); allocates the array of limit pointers (char *descrs[limit]) and sets them all to NULL.

Later, pd->descrs[device] = strdup(descr); allocates a separate string buffer for that one device, and stores its pointer into the array slot.

src/components/amd_smi/amds.c
djwoun: gpu_count = number of GPU
device_count = GPUs + CPU sockets

src/components/amd_smi/amds_evtapi.c
djwoun: right ack

src/components/amd_smi/amds_evtapi.c
djwoun: ack

src/components/amd_smi/amds_evtapi.c
djwoun: right ack

src/components/amd_smi/amds.c
djwoun: I'll have to think about this

src/components/amd_smi/amds.c
djwoun: yeah

src/components/amd_smi/amds_evtapi.c
djwoun: This padding is to match papi_native_avail’s fixed-width. papi_native_avail prints descriptions in chunks of EVT_LINE - 12 - 2 chars. Padding each device descr out to that width forces each device’s descr to start on a new folded line, so it displays as one device per line.

src/components/amd_smi/amds_evtapi.c
djwoun: snprintf(): truncation is NULL-terminated as long as remaining > 0, but if snprintf() returns < 0 the termination of the destination buffer aren’t guaranteed. Since we’re building buf incrementally, I force buf[used] = '\0' only on the < 0 error path to keep it valid.

Also, lots of devices means we could hit the truncation case (strLen >= remaining) because the combined descr exceeds the fixed buffer.

src/components/amd_smi/amds_evtapi.c
Treece-Burgess: Awesome, thank you for the explanation.

src/components/amd_smi/amds.c
Treece-Burgess: I tested this code again since your latest push and the code to condense the lines to a single line does appear:
```
| amd_smi:::compute_partition_hash                                             |    
|            Hash of Device 0,1,2,3 compute partition string 'SPX'             |    
|            Mandatory device qualifier [0,1,2,3]                              |    
```

While the cases for multiple lines being needed still holds true:
```
| amd_smi:::uuid_hash                                                          |
|            Hash of Device 0 UUID string 'f6ff74a0-0000-1000-808e-1753dc12b715|
|            '                                                                 |
|            Hash of Device 1 UUID string '5bff74a0-0000-1000-80d3-148dc76b831a|
|            '                                                                 |
|            Hash of Device 2 UUID string 'a5ff74a0-0000-1000-807d-04507c371187|
|            '                                                                 |
|            Hash of Device 3 UUID string 'cdff74a0-0000-1000-80cf-b00895cff553|
|            '                                                                 |
|            Mandatory device qualifier [0,1,2,3]                              |
```

Good work!

src/components/amd_smi/amds_evtapi.c
Treece-Burgess: Thanks for the further explanation and agreed with the later case.

src/components/amd_smi/amds.c
Treece-Burgess: Thanks for taking care of this!

src/components/amd_smi/amds.c
Treece-Burgess: Cool thanks for the explanation!

src/components/amd_smi/amds_evtapi.c
```diff
@@ -127,11 +149,62 @@ static int fill_event_description(const native_event_t *event, int device,
     return PAPI_OK;
   }
 
+  /*
+   * If all per-device description strings are identical aside from the device
+   * number, collapse them to a single line ("Device 0,1,2,3 ...") to avoid
+   * repeating the same value N times in tools like papi_native_avail.
+   */
+  const amds_per_device_descr_t *pd = event->per_device_descr;
+  int first = device_first(event->device_map);
```
Treece-Burgess: Q: If `device_first` returns -1, should this workflow continue on? From looking at `device_first` it looks to verify that the device is supported and if not -1 is returned.

src/components/amd_smi/amds_evtapi.c
```diff
@@ -87,6 +87,178 @@ static int device_next(uint64_t bitmap, int after) {
...
+    CHECK_SNPRINTF(buf, len, "%s", src);
+    return PAPI_OK;
+  }
+
+  /*
+   * Base event: if this event has per-device descrs, pack them into a single
+   * string using fixed-width fields so tools like papi_native_avail fold each
+   * device descr onto its own output line.
+   */
+  if (!(event->evtinfo_flags & AMDS_EVTINFO_FLAG_PER_DEVICE_DESCR)) {
```
djwoun, line 142: @Treece-Burgess so if device_first returns -1. It would have to end returning fallback, but in the case it does pass all thes conditions and hit `int first = device_first(event->device_map);`, it would return with "" in the buffer.

src/components/amd_smi/amds_evtapi.c
```diff
@@ -87,6 +87,178 @@ static int device_next(uint64_t bitmap, int after) {
...
+      line0 = pd->descrs[first];
+    }
+
+    size_t prefix_len0 = 0;
+    const char *suffix0 = NULL;
+    if (split_device_number_field(line0, &prefix_len0, &suffix0)) {
+      int can_collapse = 1;
+      int dev = device_next(event->device_map, first);
+      for (; dev >= 0; dev = device_next(event->device_map, dev)) {
+        const char *line = fallback;
```
djwoun, line 172: @Treece-Burgess  Right now if a per-device description isn’t present we just fall back to event->descr (best-effort), which feels like the reasonable behavior for this description-formatting feature. I’m thinking we keep line = fallback, and add a SUBDBG when device_map includes a device but pd->descrs[dev] is NULL so we can catch that inconsistency during debug. what do you think?

src/components/amd_smi/amds_evtapi.c
```diff
@@ -87,6 +87,178 @@ static int device_next(uint64_t bitmap, int after) {
...
+            memcmp(line, line0, prefix_len0) != 0 ||
+            strcmp(suffix, suffix0) != 0) {
+          can_collapse = 0;
+          break;
+        }
+      }
+
+      if (can_collapse) {
+        char devices[PAPI_MAX_STR_LEN] = {0};
+        if (format_device_bitmap(event->device_map, devices, sizeof(devices)) ==
```
djwoun: @Treece-Burgess Same idea here: If format_device_bitmap() can’t build the device list, we just skip collapsing and fall back to the normal per-device description formatting, so we don’t fail the event over a display issue. Maybe add a SUBDBG. what do you think?

src/components/amd_smi/amds_evtapi.c
```diff
@@ -87,6 +87,178 @@ static int device_next(uint64_t bitmap, int after) {
...
+            PAPI_OK) {
+          int strLen = snprintf(buf, len, "%.*s%s%s", (int)prefix_len0, line0,
```
djwoun, line 197: @Treece-Burgess  If the formatted string doesn’t fit, snprintf truncates to len-1 characters and NULL terminates. So not checking strLen >= len in the collapse fast-path just means “silently truncate the collapse description”, not overflow or corrupt memory. In line 223 we check that because it's more likely to truncate, and if it does it doesn't error out it just breaks so it doesn't continue try add more.

src/components/amd_smi/amds_evtapi.c
Treece-Burgess: I see, I just tested it as well by setting `first = -1` directly after the first call. The descriptions do still output and even the cases with multiple lines for the hash events show.

I do not think we will ever have -1 returned, but on the off chance that we do could you add an else block like so?
```
  else {
      SUBDBG("Incorrectly returned -1 for the device.\n");
      return PAPI_EBUG;
  }
```

src/components/amd_smi/amds_evtapi.c
Treece-Burgess: My main concern was if someone set `export CFLAGS="-O2 -Wformat-truncation" that this workflow would issue a compilation warning. 

From my testing that does not occur due to the values of devices, line0, and suffix0 being unknown quantities. 

Can you still add the case for if (strLen >= len) and then in that block just add a `SUBDBG`. I agree the likely hood of truncation with 1024 characters available is slim, but if it happen we can easily find out why.

src/components/amd_smi/amds_evtapi.c
Treece-Burgess: Yes, that sounds good with me!

src/components/amd_smi/amds_evtapi.c
Treece-Burgess: Adding the SUBDBG sounds good with me!

src/components/amd_smi/amds_evtapi.c
djwoun: Oh I was slightly mistaken, if nothing was written it explicitly falls back to event->descr or "". So the only way we end up with "" is when event->descr itself is empty.

I thought about it and this would be a pretty critical bug, so I think your suggestion makes sense.

src/components/amd_smi/amds_evtapi.c
djwoun: cool

src/components/amd_smi/amds_evtapi.c
djwoun: cool

src/components/amd_smi/amds_evtapi.c
djwoun: Ah I see that is fair.

Treece-Burgess (approved):
I tested these changes on Odyssey at Oregon with ROCm 7.2.0 pre-release, below are the results.

- PAPI build: ✅ 
- PAPI utilities*: ✅ 
- `amd_smi` component tests: ✅ (I lack the permissions for `amdsmi_set_test.c`)

For `papi_native_avail`, we now see the string description for each device and if applicable a collapsed version if all devices repeat the same value:
```
# Collapsed version
6325 --------------------------------------------------------------------------------
6326 | amd_smi:::board_product_name_hash                                            |
6327 |            Hash of Device 0,1,2,3 board product name string 'Aqua Vanjaram [I|
6328 |            nstinct MI300A]'                                                  |
6329 |     :device=0                                                                |
6330 |            Mandatory device qualifier [0,1,2,3]                              |
6331 --------------------------------------------------------------------------------

# Not collapsed version
6499 --------------------------------------------------------------------------------
6500 | amd_smi:::uuid_hash                                                          |
6501 |            Hash of Device 0 UUID string 'f6ff74a0-0000-1000-808e-1753dc12b715|
6502 |            '                                                                 |
6503 |            Hash of Device 1 UUID string '5bff74a0-0000-1000-80d3-148dc76b831a|
6504 |            '                                                                 |
6505 |            Hash of Device 2 UUID string 'a5ff74a0-0000-1000-807d-04507c371187|
6506 |            '                                                                 |
6507 |            Hash of Device 3 UUID string 'cdff74a0-0000-1000-80cf-b00895cff553|
6508 |            '                                                                 |
6509 |     :device=0                                                                |
6510 |            Mandatory device qualifier [0,1,2,3]                              |
6511 --------------------------------------------------------------------------------
```

__*__ - `papi_component_avail`, `papi_native_avail`, `papi_command_line`

PR #551, opened 2026-02-02
ci: add scripts for amd_smi component

dbarry9:
## Pull Request Description

This pull request proposes the following:
- updates the CI to use ROCm 7.0.1 instead of 6.3.2 for the `rocp_sdk` and `amd_smi` components, and
- adds CI support for the `amd_smi` component.

These changes have been tested using ROCm 7.0.1 on the AMD Radeon VII (gfx906) architecture.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/rocm_smi/Rules.rocm_smi
+++ b/src/components/rocm_smi/Rules.rocm_smi
@@ -88,6 +88,7 @@ CFLAGS += -I$(PAPI_ROCMSMI_ROOT)/../include/rocm_smi
 CFLAGS += -I$(PAPI_ROCMSMI_ROOT)/../include
 CFLAGS += -I$(PAPI_ROCMSMI_ROOT)/include/rocm_smi
 CFLAGS += -I$(PAPI_ROCMSMI_ROOT)/include
+CFLAGS += -I/opt/amdgpu/include
 CFLAGS += -g
 LDFLAGS += $(LDL) -g
 
--- a/src/components/sysdetect/Rules.sysdetect
+++ b/src/components/sysdetect/Rules.sysdetect
@@ -24,6 +24,7 @@ CFLAGS += -I$(PAPI_ROCM_ROOT)/include                  \
           -I$(PAPI_ROCM_ROOT)/rocm_smi/include         \
           -I$(PAPI_ROCM_ROOT)/rocm_smi/include/rocm_smi
 endif
+CFLAGS += -I/opt/amdgpu/include
 
 ifneq ($(PAPI_CUDA_ROOT),)
 CFLAGS += -I$(PAPI_CUDA_ROOT)/include


.github/workflows/papi_framework_workflow.yml
```diff
@@ -30,7 +30,7 @@ jobs:
   papi_components_amd:
     strategy:
       matrix:
-        components: [rocp_sdk rocm_smi]
+        components: [rocm rocm_smi]
```
Treece-Burgess: This should remain `rocp_sdk` and `rocm_smi` should become `amd_smi`.

Before Anthony left, we agreed that swapping the older components for the newer components for the framework tests took precedent. They still will have tests rans for component level updates.

.github/workflows/papi_framework_workflow.yml
```diff
@@ -39,7 +39,7 @@ jobs:
     timeout-minutes: 60
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
-      - name: rocp_sdk and rocm_smi component tests
+      - name: rocm and rocm_smi component tests
```
Treece-Burgess: Change to "rocp_sdk and amd_smi component tests"

src/components/rocm_smi/Rules.rocm_smi
```diff
@@ -88,6 +88,7 @@ CFLAGS += -I$(PAPI_ROCMSMI_ROOT)/../include/rocm_smi
 CFLAGS += -I$(PAPI_ROCMSMI_ROOT)/../include
 CFLAGS += -I$(PAPI_ROCMSMI_ROOT)/include/rocm_smi
 CFLAGS += -I$(PAPI_ROCMSMI_ROOT)/include
+CFLAGS += -I/opt/amdgpu/include
```
Treece-Burgess, line 91: Q: Why was this change needed?

src/components/sysdetect/Rules.sysdetect
```diff
@@ -24,6 +24,7 @@ CFLAGS += -I$(PAPI_ROCM_ROOT)/include                  \
           -I$(PAPI_ROCM_ROOT)/rocm_smi/include         \
           -I$(PAPI_ROCM_ROOT)/rocm_smi/include/rocm_smi
 endif
+CFLAGS += -I/opt/amdgpu/include
```
Treece-Burgess, line 27: Q: Why was this change needed?

src/components/rocm_smi/Rules.rocm_smi
dbarry9: Some systems only have this drm header in this location, and libdrm/drm.h is placed in this location by default (from the AMD software stack) for Redhat-based Linux distributions.

src/components/sysdetect/Rules.sysdetect
dbarry9: Some systems only have this drm header in this location, and libdrm/drm.h is placed in this location by default (from the AMD software stack) for Redhat-based Linux distributions.

.github/workflows/papi_framework_workflow.yml
dbarry9: I understand; I will make this change.

dbarry9: Line 29 states:
> build PAPI only with rocm and rocm_smi components, as they will not be active in the above comprehensive job

However, I believe this was not true prior to this PR because the comprehensive test had `rocm_smi` included, correct? The reason I ask: I think I should remove or change this comment because, as stated, it implies that `papi_components_comprehensive` is intended to test the most recent AMD components (_e.g.,_ `rocp_sdk`, and now `amd_smi`), but the `papi_components_amd` is intended to test the legacy AMD components (_e.g.,_ `rocm` and `rocm_smi`).

Treece-Burgess: The comprehensive test is designed to compile as many components as possible and make sure they will either be active if they should be or inactive if they should be. The test runs on a machine with NVIDIA devices; therefore, the AMD components would not be active. Due to this you have the `papi_components_amd` that will then run on a machine with AMD devices such that we can make sure those components are compiling and active as well.

The comment can be updated to swap rocm with rocp_sdk and rocm_smi with amd_smi.

dbarry9: I see. Thank you!

PR #555, opened 2026-02-05
amd_smi: fail on unexpected add/start errors

djwoun:
## Pull Request Description
Updates EXIT_WARNING_ON_ADD to treat unexpected error codes (e.g., PAPI_EISRUN, PAPI_EINVAL) as test failures. Previously, non-listed errors were silently ignored, causing tests to pass incorrectly.

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/components/amd_smi/tests/test_harness.h
+++ b/src/components/amd_smi/tests/test_harness.h
@@ -218,11 +218,14 @@ static inline int harness_eval_result(const char *file, int line, HarnessOpts op
  *        limits, exit as "PASSED with WARNING".
  *
  * Recognizes PAPI_ENOEVNT, PAPI_ECNFLCT, PAPI_EPERM, and PAPI_ENOSUPP.
+ * Any other non-OK return code is treated as a test failure.
  */
 #define EXIT_WARNING_ON_ADD(rc, evname) do { \
     if ((rc) == PAPI_ENOEVNT || (rc) == PAPI_ECNFLCT || (rc) == PAPI_EPERM || (rc) == PAPI_ENOSUPP) { \
         EXIT_WARNING("Event unavailable (%s): %s", PAPI_strerror(rc), (evname)); \
     } \
+    fprintf(stderr, "Unexpected error encountered (%d) when adding the event %s.\n", (rc), (evname)); \
+    exit(1); \
 } while (0)
 
 /**


src/components/amd_smi/tests/test_harness.h
```diff
@@ -218,11 +218,15 @@ static inline int harness_eval_result(const char *file, int line, HarnessOpts op
  *        limits, exit as "PASSED with WARNING".
  *
  * Recognizes PAPI_ENOEVNT, PAPI_ECNFLCT, PAPI_EPERM, and PAPI_ENOSUPP.
+ * Any other non-OK return code is treated as a test failure.
  */
 #define EXIT_WARNING_ON_ADD(rc, evname) do { \
     if ((rc) == PAPI_ENOEVNT || (rc) == PAPI_ECNFLCT || (rc) == PAPI_EPERM || (rc) == PAPI_ENOSUPP) { \
         EXIT_WARNING("Event unavailable (%s): %s", PAPI_strerror(rc), (evname)); \
     } \
+    char _papi_call[PAPI_MAX_STR_LEN + 64] = {0}; \
```
Treece-Burgess: I would suggest just condensing down `_papi_call`, `snprintf`, and `test_fail` into:
```
fprintf(stderr, "Unexpected error encountered (%d) when adding the event %s.\n", rc, (evname)); \
exit(1); \
```

A big reason why is that the output for the call to `test_fail` does not make the most sense unless you put a PAPI API call where you pass your string:
```
[tburgess@odyssey tests]$ ./amdsmi_set_test 
FAILED!!!
Line # 119 Error in Unexpected error adding event amd_smi:::power_cap: Bad combination of features
Some tests require special hardware, permissions, OS, compilers
or library versions. PAPI may still function perfectly on your 
system without the particular feature being tested here.    

```

Treece-Burgess (changes requested):

src/components/amd_smi/tests/test_harness.h
djwoun: Good point. test_fail reads oddly when the call isn’t an exact PAPI API(it returns from a series of PAPI calls). I’ve simplified EXIT_WARNING_ON_ADD to fprintf(stderr)+ exit(1) for unexpected add return codes.

Treece-Burgess: Looks good!

PR #568, opened 2026-02-25
amd_smi: add gpu_metrics events, refine descriptions, and suppress sentinel-valued fields

djwoun:
## Pull Request Description
Expands metric coverage, updates descriptions for select existing GPU metrics events, suppresses registration of struct-backed events when AMD SMI returns sentinel values.

In particular, it:
- adds and expands GPU metrics events
- updates descriptions for select existing GPU metrics events where better AMD SMI documentation is available
- filters out sentinel-backed fields at registration time for struct-based event groups so unsupported metrics are not exposed as garbage values such as `-1`

## Author Checklist
- [x] **Description**
- [x] **Commits**
- [x] **Tests**

### DIFF
--- a/src/components/amd_smi/amds.c
+++ b/src/components/amd_smi/amds.c
@@ -1153,128 +1153,121 @@ static int init_event_table(void) {
     // PCIe information events
     if (amdsmi_get_pcie_info_p) {
       amdsmi_pcie_info_t pcie_info;
-      if (amdsmi_get_pcie_info_p(device_handles[d], &pcie_info) ==
-          AMDSMI_STATUS_SUCCESS) {
-        CHECK_EVENT_IDX(idx);
-        CHECK_SNPRINTF(name_buf, sizeof(name_buf), "pcie_max_width:device=%d", d);
-        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf),
-                 "Device %d maximum PCIe link width (lanes)", d);
-        if (add_event(&idx, name_buf, descr_buf, d, 0, 0, PAPI_MODE_READ,
-                      access_amdsmi_pcie_info) != PAPI_OK)
-          return PAPI_ENOMEM;
+      memset(&pcie_info, 0, sizeof(pcie_info));
+      if (amdsmi_get_pcie_info_p(device_handles[d], &pcie_info) == AMDSMI_STATUS_SUCCESS) {
+        if (pcie_info.pcie_static.max_pcie_width != UINT16_MAX) {
+          CHECK_EVENT_IDX(idx);
+          CHECK_SNPRINTF(name_buf, sizeof(name_buf), "pcie_max_width:device=%d", d);
+          CHECK_SNPRINTF(descr_buf, sizeof(descr_buf), "Device %d maximum PCIe link width (lanes)", d);
+          if (add_event(&idx, name_buf, descr_buf, d, 0, 0, PAPI_MODE_READ, access_amdsmi_pcie_info) != PAPI_OK)
+            return PAPI_ENOMEM;
+        }
 
-        CHECK_EVENT_IDX(idx);
-        CHECK_SNPRINTF(name_buf, sizeof(name_buf), "pcie_max_speed:device=%d", d);
-        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf),
-                 "Device %d maximum PCIe link speed (GT/s)", d);
-        if (add_event(&idx, name_buf, descr_buf, d, 1, 0, PAPI_MODE_READ,
-                      access_amdsmi_pcie_info) != PAPI_OK)
-          return PAPI_ENOMEM;
+        if (pcie_info.pcie_static.max_pcie_speed != UINT32_MAX) {
+          CHECK_EVENT_IDX(idx);
+          CHECK_SNPRINTF(name_buf, sizeof(name_buf), "pcie_max_speed:device=%d", d);
+          CHECK_SNPRINTF(descr_buf, sizeof(descr_buf), "Device %d maximum PCIe link speed (GT/s)", d);
+          if (add_event(&idx, name_buf, descr_buf, d, 1, 0, PAPI_MODE_READ, access_amdsmi_pcie_info) != PAPI_OK)
+            return PAPI_ENOMEM;
+        }
 
-        CHECK_EVENT_IDX(idx);
-        CHECK_SNPRINTF(name_buf, sizeof(name_buf), "pcie_interface_version:device=%d",
-                 d);
-        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf),
-                 "Device %d PCIe interface version", d);
-        if (add_event(&idx, name_buf, descr_buf, d, 2, 0, PAPI_MODE_READ,
-                      access_amdsmi_pcie_info) != PAPI_OK)
-          return PAPI_ENOMEM;
+        if (pcie_info.pcie_static.pcie_interface_version != UINT32_MAX) {
+          CHECK_EVENT_IDX(idx);
+          CHECK_SNPRINTF(name_buf, sizeof(name_buf), "pcie_interface_version:device=%d", d);
+          CHECK_SNPRINTF(descr_buf, sizeof(descr_buf), "Device %d PCIe interface version", d);
+          if (add_event(&idx, name_buf, descr_buf, d, 2, 0, PAPI_MODE_READ, access_amdsmi_pcie_info) != PAPI_OK)
+            return PAPI_ENOMEM;
+        }
 
-        CHECK_EVENT_IDX(idx);
-        CHECK_SNPRINTF(name_buf, sizeof(name_buf), "pcie_slot_type:device=%d", d);
-        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf),
-                 "Device %d PCIe slot type", d);
-        if (add_event(&idx, name_buf, descr_buf, d, 3, 0, PAPI_MODE_READ,
-                      access_amdsmi_pcie_info) != PAPI_OK)
-          return PAPI_ENOMEM;
+        if (pcie_info.pcie_static.slot_type >= AMDSMI_CARD_FORM_FACTOR_PCIE && pcie_info.pcie_static.slot_type <= AMDSMI_CARD_FORM_FACTOR_UNKNOWN) {
+          CHECK_EVENT_IDX(idx);
+          CHECK_SNPRINTF(name_buf, sizeof(name_buf), "pcie_slot_type:device=%d", d);
+          CHECK_SNPRINTF(descr_buf, sizeof(descr_buf), "Device %d PCIe slot type", d);
+          if (add_event(&idx, name_buf, descr_buf, d, 3, 0, PAPI_MODE_READ, access_amdsmi_pcie_info) != PAPI_OK)
+            return PAPI_ENOMEM;
+        }
 
 #if AMDSMI_LIB_VERSION_MAJOR >= 25
-        if (amdsmi_lib_major >= 25) {
+        if (amdsmi_lib_major >= 25 && pcie_info.pcie_static.max_pcie_interface_version != UINT32_MAX) {
           CHECK_EVENT_IDX(idx);
-          CHECK_SNPRINTF(name_buf, sizeof(name_buf),
-                   "pcie_max_interface_version:device=%d", d);
-          CHECK_SNPRINTF(descr_buf, sizeof(descr_buf),
-                   "Device %d maximum PCIe interface version", d);
-          if (add_event(&idx, name_buf, descr_buf, d, 4, 0, PAPI_MODE_READ,
-                        access_amdsmi_pcie_info) != PAPI_OK)
+          CHECK_SNPRINTF(name_buf, sizeof(name_buf), "pcie_max_interface_version:device=%d", d);
+          CHECK_SNPRINTF(descr_buf, sizeof(descr_buf), "Device %d maximum PCIe interface version", d);
+          if (add_event(&idx, name_buf, descr_buf, d, 4, 0, PAPI_MODE_READ, access_amdsmi_pcie_info) != PAPI_OK)
             return PAPI_ENOMEM;
         }
 #endif
 
-        CHECK_EVENT_IDX(idx);
-        CHECK_SNPRINTF(name_buf, sizeof(name_buf), "pcie_width:device=%d", d);
-        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf),
-                 "Device %d current PCIe link width (lanes)", d);
-        if (add_event(&idx, name_buf, descr_buf, d, 5, 0, PAPI_MODE_READ,
-                      access_amdsmi_pcie_info) != PAPI_OK)
-          return PAPI_ENOMEM;
+        if (pcie_info.pcie_metric.pcie_width != UINT16_MAX) {
+          CHECK_EVENT_IDX(idx);
+          CHECK_SNPRINTF(name_buf, sizeof(name_buf), "pcie_width:device=%d", d);
+          CHECK_SNPRINTF(descr_buf, sizeof(descr_buf), "Device %d current PCIe link width (lanes)", d);
+          if (add_event(&idx, name_buf, descr_buf, d, 5, 0, PAPI_MODE_READ, access_amdsmi_pcie_info) != PAPI_OK)
+            return PAPI_ENOMEM;
+        }
 
-        CHECK_EVENT_IDX(idx);
-        CHECK_SNPRINTF(name_buf, sizeof(name_buf), "pcie_speed:device=%d", d);
-        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf),
-                 "Device %d current PCIe link speed (MT/s)", d);
-        if (add_event(&idx, name_buf, descr_buf, d, 6, 0, PAPI_MODE_READ,
-                      access_amdsmi_pcie_info) != PAPI_OK)
-          return PAPI_ENOMEM;
+        if (pcie_info.pcie_metric.pcie_speed != UINT32_MAX) {
+          CHECK_EVENT_IDX(idx);
+          CHECK_SNPRINTF(name_buf, sizeof(name_buf), "pcie_speed:device=%d", d);
+          CHECK_SNPRINTF(descr_buf, sizeof(descr_buf), "Device %d current PCIe link speed (MT/s)", d);
+          if (add_event(&idx, name_buf, descr_buf, d, 6, 0, PAPI_MODE_READ, access_amdsmi_pcie_info) != PAPI_OK)
+            return PAPI_ENOMEM;
+        }
 
-        CHECK_EVENT_IDX(idx);
-        CHECK_SNPRINTF(name_buf, sizeof(name_buf), "pcie_bandwidth:device=%d", d);
-        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf),
-                 "Device %d instantaneous PCIe bandwidth (Mb/s)", d);
-        if (add_event(&idx, name_buf, descr_buf, d, 7, 0, PAPI_MODE_READ,
-                      access_amdsmi_pcie_info) != PAPI_OK)
-          return PAPI_ENOMEM;
+        if (pcie_info.pcie_metric.pcie_bandwidth != UINT32_MAX) {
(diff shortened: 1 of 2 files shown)


Treece-Burgess:
I am reviewing this PR.

src/components/amd_smi/amds.c
```diff
@@ -3032,100 +2902,147 @@ static int init_event_table(void) {
...
+        CHECK_EVENT_IDX(idx);
+        CHECK_SNPRINTF(name_buf, sizeof(name_buf), "socket_thm_residency_acc:device=%d", d);
+        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf), "Device %d accumulated socket thermal throttler residency", d);
+        if (add_event(&idx, name_buf, descr_buf, d, 14, 0, PAPI_MODE_READ,
+                      access_amdsmi_gpu_metrics) != PAPI_OK)
+          return PAPI_ENOMEM;
+
+        CHECK_EVENT_IDX(idx);
+        CHECK_SNPRINTF(name_buf, sizeof(name_buf), "vr_thm_residency_acc:device=%d", d);
+        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf), "Device %d accumulated voltage regulator thermal throttler residency", d);
```
Treece-Burgess: From `papi_native_avail` on Illyad, this metric outputs a newline as follows:
```
5530 --------------------------------------------------------------------------------
5531 | amd_smi:::vr_thm_residency_acc                                               |
5532 |            Device 0 accumulated voltage regulator thermal throttler residency|
5533 |                                                                              |
5534 |     :device=0                                                                |
5535 |            Mandatory device qualifier [0,1]                                  |
5536 --------------------------------------------------------------------------------
```
Is this from `papi_native_avail` formatting or is this coming from `amd_smi`?

src/components/amd_smi/amds.c
```diff
@@ -3032,100 +2902,147 @@ static int init_event_table(void) {
...
+        CHECK_EVENT_IDX(idx);
+        CHECK_SNPRINTF(name_buf, sizeof(name_buf), "mem_activity_acc:device=%d", d);
+        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf), "Device %d memory activity accumulated counter (percent)", d);
+        if (add_event(&idx, name_buf, descr_buf, d, 18, 0, PAPI_MODE_READ,
+                      access_amdsmi_gpu_metrics) != PAPI_OK)
+          return PAPI_ENOMEM;
+
+        CHECK_EVENT_IDX(idx);
+        CHECK_SNPRINTF(name_buf, sizeof(name_buf), "num_partition:device=%d", d);
+        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf), "Device %d number of current partitions", d);
```
Treece-Burgess: Q: Are you able to list in the description all the devices that are present for that event? For example. `num_partitions` show:
```
5552 --------------------------------------------------------------------------------
5553 | amd_smi:::num_partition                                                      |
5554 |            Device 0 number of current partitions                             |
5555 |     :device=0                                                                |
5556 |            Mandatory device qualifier [0,1]                                  |
5557 --------------------------------------------------------------------------------
```

The description only makes note of Device 0, but you also have support for device 1 as well. So would it be possible to do: `Device 0,1 number of current partitions`? I

src/components/amd_smi/amds.c
```diff
@@ -3032,100 +2902,147 @@ static int init_event_table(void) {
         /* Register throttle status */
         CHECK_EVENT_IDX(idx);
         CHECK_SNPRINTF(name_buf, sizeof(name_buf), "gpu_throttle_status:device=%d", d);
-        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf),
-                 "Device %d throttle status", d);
+        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf), "Device %d current throttle status bitmask", d);
```
Treece-Burgess: nit: There are a few of the events that have been reformatted. Could you revert these changes as this PR is for adding the new counters?

src/components/amd_smi/amds.c
```diff
@@ -3032,100 +2902,147 @@ static int init_event_table(void) {
...
-        CHECK_SNPRINTF(name_buf, sizeof(name_buf),
-                 "pcie_nak_rcvd_count_acc:device=%d", d);
-        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf),
-                 "Device %d PCIe NAK received count", d);
+        CHECK_SNPRINTF(name_buf, sizeof(name_buf), "pcie_nak_rcvd_count_acc:device=%d", d);
+        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf), "Device %d PCIe NAK received accumulated count", d);
         if (add_event(&idx, name_buf, descr_buf, d, 10, 0, PAPI_MODE_READ,
                       access_amdsmi_gpu_metrics) != PAPI_OK)
           return PAPI_ENOMEM;
+
```
Treece-Burgess, line 2976: The new events all check out on Odyssey, but on Illyad with ROCm 7.1.1, the events:
```
amd_smi:::accumulation_counter
amd_smi:::prochot_residency_acc
amd_smi:::ppt_residency_acc 
amd_smi:::socket_thm_residency_acc
amd_smi:::vr_thm_residency_acc
amd_smi:::hbm_thm_residency_acc
```

will show a counter value of -1 when ran with `papi_command_line`. Do you know why this is occurring?

src/components/amd_smi/amds_accessors.c
```diff
@@ -1248,6 +1248,33 @@ int access_amdsmi_gpu_metrics(int mode, void *arg) {
   case 10:
     event->value = metrics.pcie_nak_rcvd_count_acc;
     break;
+  case 11:
+    event->value = (int64_t)metrics.accumulation_counter;
```
Treece-Burgess: Q: Why do you need to type cast `accumulation_counter`, `prochot_residency_acc`, `ppt_residency_acc`, `socket_thm_residency_acc`, `vrm_thm_residency_acc`, and `hbm_thm_residency_acc` to `int64_t`? Based on the docs they are already `uint64_t` and `event->value` is `uint64_t` as well.

src/components/amd_smi/amds.c
djwoun: There is a small problem where amdsmi_get_gpu_metrics_info_p will return true if it returns just one true metric out of the 20 metrics it queries for. 

I'm thinking maybe I should add additional checks to see if it returns sentinel values?

src/components/amd_smi/amds_accessors.c
djwoun: Wait yeah, the casting is completely unecessary. I'll remove it.

src/components/amd_smi/amds.c
djwoun: Maybe I should rename the PR to include the modified descriptions detail? I found better documentation on what these metrics exactly do and they are under the same function(amdsmi_get_gpu_metrics_info_p) that queries these metrics, so I wanted to group it in under this PR.

src/components/amd_smi/amds.c
djwoun: This is how the generic event descriptions are for AMD SMI, where it exposes device 0's event description.
Like:
"amd_smi:::L1_dcache_size_type_0"
"Device 0 L1 data cache size (bytes)"
It may be worth in future to add capabilities to have "0,1" listings like the hasing event descriptions.

src/components/amd_smi/amds.c
djwoun: yeah, I think so that line of text must be exactly 66 characeters and papi_native_avail just cuts it right there

src/components/amd_smi/amds.c
Treece-Burgess: Yes, if you could update the PR to include that you are updating the descriptions of select events that would be good!

src/components/amd_smi/amds.c
Treece-Burgess: I see, thanks for the clarification. Yes that could be a future feature to add and is out of scope of this PR.

src/components/amd_smi/amds.c
Treece-Burgess: I think that would be a good idea! What were you thinking the check could be?

djwoun: AMD SMI calls can return AMDSMI_STATUS_SUCCESS even when only part of the output struct is actually valid, and the unsupported fields are left at sentinel values. For the residency and accumulation counters that sentinel is UINT64_MAX, which PAPI ends up showing as -1. But the same pattern shows up in other struct-returning calls too, with UINT16_MAX, UINT32_MAX, UINT64_MAX, and in a few cases things like UINT8_MAX.

So the fix in this PR is the generic one: zero-init the struct, probe it in amds.c, and only register an event if the specific backing field is not its width-appropriate sentinel. That is why the PR applies the same check to other struct-based events.

src/components/amd_smi/amds.c
djwoun: yeah

src/components/amd_smi/amds.c
```diff
@@ -1279,128 +1153,121 @@ static int init_event_table(void) {
...
+        }
 
-        CHECK_EVENT_IDX(idx);
-        CHECK_SNPRINTF(name_buf, sizeof(name_buf), "pcie_speed:device=%d", d);
-        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf),
-                 "Device %d current PCIe link speed (MT/s)", d);
-        if (add_event(&idx, name_buf, descr_buf, d, 6, 0, PAPI_MODE_READ,
-                      access_amdsmi_pcie_info) != PAPI_OK)
-          return PAPI_ENOMEM;
+        if (pcie_info.pcie_metric.pcie_speed != UINT32_MAX) {
```
Treece-Burgess: Why the added conditional check?

src/components/amd_smi/amds.c
```diff
@@ -1279,128 +1153,121 @@ static int init_event_table(void) {
...
 
-        CHECK_EVENT_IDX(idx);
-        CHECK_SNPRINTF(name_buf, sizeof(name_buf),
-                 "pcie_other_end_recovery_count:device=%d", d);
-        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf),
-                 "Device %d PCIe other-end recovery count", d);
-        if (add_event(&idx, name_buf, descr_buf, d, 13, 0, PAPI_MODE_READ,
-                      access_amdsmi_pcie_info) != PAPI_OK)
-          return PAPI_ENOMEM;
+        if (pcie_info.pcie_metric.pcie_lc_perf_other_end_recovery_count != UINT32_MAX) {
```
Treece-Burgess: Why the added conditional check?

src/components/amd_smi/amds.c
```diff
@@ -2165,32 +2032,31 @@ static int init_event_table(void) {
...
+      }
       // Instantaneous socket power (W)
-      CHECK_EVENT_IDX(idx);
-      CHECK_SNPRINTF(name_buf, sizeof(name_buf), "power_current:device=%d", d);
-      CHECK_SNPRINTF(descr_buf, sizeof(descr_buf),
-               "Device %d current socket power (W)", d);
-      if (add_event(&idx, name_buf, descr_buf, d, 1, 0, PAPI_MODE_READ,
-                    access_amdsmi_power_average) != PAPI_OK)
-        return PAPI_ENOMEM;
+      if (dummy_power.current_socket_power != UINT32_MAX && dummy_power.current_socket_power != UINT16_MAX) {
```
Treece-Burgess: Why the added conditional check?

src/components/amd_smi/amds.c
```diff
@@ -2319,27 +2185,29 @@ static int init_event_table(void) {
...
-      CHECK_SNPRINTF(name_buf, sizeof(name_buf), "mm_activity:device=%d", d);
-      CHECK_SNPRINTF(descr_buf, sizeof(descr_buf),
-               "Device %d MM engine activity (%%)", d);
-      if (add_event(&idx, name_buf, descr_buf, d, 2, 0, PAPI_MODE_READ,
-                    access_amdsmi_gpu_activity) != PAPI_OK)
-        return PAPI_ENOMEM;
+    memset(&dummy_usage, 0, sizeof(dummy_usage));
+    if (amdsmi_get_gpu_activity_p && amdsmi_get_gpu_activity_p(device_handles[d], &dummy_usage) == AMDSMI_STATUS_SUCCESS) {
+      if (dummy_usage.gfx_activity != UINT32_MAX && dummy_usage.gfx_activity != UINT16_MAX) {
+        CHECK_EVENT_IDX(idx);
```
Treece-Burgess, line 2191: Was the `CHECK_EVENT_IDX` here and the two below originally forgotten when `amd_smi` was merged into master? From the diff it appears that way.

src/components/amd_smi/amds.c
```diff
@@ -2449,17 +2316,16 @@ static int init_event_table(void) {
           "deep sleep frequency (MHz)"};
       for (int t = 0; t < 2; ++t) {
         amdsmi_clk_info_t info;
-        if (amdsmi_get_clock_info_p(device_handles[d], clk_types[t], &info) !=
-            AMDSMI_STATUS_SUCCESS)
+        memset(&info, 0, sizeof(info));
+        if (amdsmi_get_clock_info_p(device_handles[d], clk_types[t], &info) != AMDSMI_STATUS_SUCCESS)
           continue;
         for (int f = 0; f < 5; ++f) {
+          if (f == 4 && info.clk_deep_sleep == UINT8_MAX)
```
Treece-Burgess, line 2323: What is the reasoning for continuing here? Could you add a comment above the for loop header for documentation.

src/components/amd_smi/amds.c
```diff
@@ -2919,83 +2782,74 @@ static int init_event_table(void) {
...
         /* Register GFX voltage */
-        CHECK_EVENT_IDX(idx);
-        CHECK_SNPRINTF(name_buf, sizeof(name_buf),
-                 "power_sensor_gfx_voltage_mv_sensor=%u:device=%d", s, d);
-        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf),
-                 "Device %d power sensor %u GFX voltage (mV)", d, s);
-        if (add_event(&idx, name_buf, descr_buf, d, 3, s, PAPI_MODE_READ,
-                      access_amdsmi_power_sensor) != PAPI_OK)
-          return PAPI_ENOMEM;
+        if (pinfo.gfx_voltage != UINT32_MAX && pinfo.gfx_voltage != UINT16_MAX) {
```
Treece-Burgess: Why the added conditional check?

src/components/amd_smi/amds.c
```diff
@@ -2919,83 +2782,74 @@ static int init_event_table(void) {
...
         /* Register MEM voltage */
-        CHECK_EVENT_IDX(idx);
-        CHECK_SNPRINTF(name_buf, sizeof(name_buf),
-                 "power_sensor_mem_voltage_mv_sensor=%u:device=%d", s, d);
-        CHECK_SNPRINTF(descr_buf, sizeof(descr_buf),
-                 "Device %d power sensor %u MEM voltage (mV)", d, s);
-        if (add_event(&idx, name_buf, descr_buf, d, 5, s, PAPI_MODE_READ,
-                      access_amdsmi_power_sensor) != PAPI_OK)
-          return PAPI_ENOMEM;
+        if (pinfo.mem_voltage != UINT32_MAX && pinfo.mem_voltage != UINT16_MAX) {
```
Treece-Burgess: Why the added conditional check?

src/components/amd_smi/amds.c
Treece-Burgess: I saw the use of conditionals checking UINT16_MAX, UINT32_MAX, UINT64_MAX, etc. A follow up question is: How do you know those are the correct? For example, you have the following conditional check:
```
if (pinfo.current_socket_power != UINT32_MAX && pinfo.current_socket_power != UINT16_MAX)
```

Looking at [documentation](https://rocm.docs.amd.com/projects/amdsmi/en/latest/doxygen/docBin/html/structamdsmi__power__info__t.html#a862241df891d3436554a34774165b719) `current_socket_power` only has UINT32_t. However, you also check UINT16_MAX.

src/components/amd_smi/amds.c
Treece-Burgess: This has been done, thank you!

src/components/amd_smi/amds.c
```diff
@@ -1279,128 +1153,121 @@ static int init_event_table(void) {
     // PCIe information events
     if (amdsmi_get_pcie_info_p) {
```
Treece-Burgess, line 1154: From my counting 36 new events are added. Of the 36, 7 events do not appear on either the MI210 or MI300A they are:
```
pcie_nak_sent_count_acc
pcie_nak_rcvd_count_acc
average_vclk1_frequency
average_dclk1_frequency
vcn_activity_vcn
jpeg_activity_jpeg
xcp_gfx_below_host_limit_acc_xcp
```

Do you know the architecture these events appear on?

src/components/amd_smi/amds.c
```diff
@@ -1279,128 +1153,121 @@ static int init_event_table(void) {
     // PCIe information events
```
Treece-Burgess, line 1153: When running `./papi_component_avail` on the master branch, `amd_smi` shows 342 native events on Odyssey at Oregon with ROCm 7.2.0. Doing the same for this PR, I see 390 native events. An increase is expected, but from my count 71 new events show in the output of `papi_native_avail`. Meaning we should see 413 output instead of 390.

src/components/amd_smi/amds_accessors.c
```diff
@@ -1228,26 +1228,161 @@ int access_amdsmi_gpu_metrics(int mode, void *arg) {
...
+  case 40: {
+    uint32_t xcp_index = event->subvariant >> 16;
+    uint32_t xcc_index = event->subvariant & 0xFFFF;
+    if (xcp_index >= AMDSMI_MAX_NUM_XCP || xcc_index >= AMDSMI_MAX_NUM_XCC)
+      return PAPI_ENOSUPP;
+    event->value = metrics.xcp_stats[xcp_index].gfx_busy_inst[xcc_index];
+    break;
+  }
+  case 41: {
+    uint32_t xcp_index = event->subvariant >> 16;
```
Treece-Burgess, line 1355: Q: I am not familiar enough with the workflow, so what will `event->subvariant` hold? I know from one of the header files it is a `uint32_t`, but how is both right shifting and ANDing going to give us the correct indices?

src/components/amd_smi/amds.c
Treece-Burgess: Yes, it is exactly 66 characters. Due to the condition in `while ( j <= strlen( info->long_descr ) ) {` in `papi_native_avail.c` we end up entering the body a second time and padding an entire second line.

So this is not from this PR. Thank you!

src/components/amd_smi/amds.c
djwoun: I added the extra UINT16_MAX check because I was seeing some uint32_t AMD SMI fields come back as 65535. After printing the values from amdsmi_get_power_info_p(), that does appear to be what is happening for average_socket_power.

src/components/amd_smi/amds.c
djwoun: Yeah, looks like it was missing before.

src/components/amd_smi/amds.c
djwoun: That continue is only there to skip adding the deep_sleep event when clk_deep_sleep is UINT8_MAX.

src/components/amd_smi/amds.c
djwoun: I am not sure but I would guess some older generations of AMD gpus because this function was created with ROCm SMI.

src/components/amd_smi/amds.c
djwoun: This PR adds new events, but it also adds registration-time sentinel filtering to some existing struct-based events. So the expected total is not just 342 + 71; it is 342 + added - filtered. On Odyssey that appears to be why the total is 390 instead of 413.

src/components/amd_smi/amds_accessors.c
djwoun: subvariant is being used here to carry the indices needed for the XCP array lookups. These events need two indices, not one: the XCP index and the inner element index. Since native_event_t only has one subvariant field, both are packed into that uint32_t during registration and unpacked in the accessor.

PR #570, opened 2026-03-09
Make net refresh latency configurable

ajanicijamd:
## Author Checklist
- [*] **Description**
Until now, minimum net refresh latency has been hard-coded to 1 second. This PR makes it configurable via
environment variable PAPI_NET_REFRESH_LATENCY.

- [*] **Commits**

- [*] **Tests**
The change has been tested in AMD's [ROCm profiler](https://github.com/ROCm/rocm-systems/tree/develop/projects/rocprofiler-systems). The reason for this change is that 1 second is too long for the purpose of the profiler's automated test, which was causing the test to fail, because it was often finishing too soon, without collecting any network statistics. With this change, the minimum net refresh latency can be made shorter, letting the test pass reliably.

### DIFF
--- a/src/components/net/CHANGES
+++ b/src/components/net/CHANGES
@@ -1,5 +1,9 @@
 Net component changelog:
 
+2026-03-10 Aleksandar Janicijevic
+
+* Add support for environment variable PAPI_NET_REFRESH_LATENCY
+
 2011-11-07 Jose Pedro Oliveira
 
  * Dynamically detects the network interfaces
@@ -15,4 +19,3 @@ Net component changelog:
  * Adds support for PAPI_event_name_to_code()
 
  * Adds a couple of small tests/examples
-
--- a/src/components/net/linux-net.c
+++ b/src/components/net/linux-net.c
@@ -22,6 +22,7 @@
 #include <ctype.h>
 #include <string.h>
 #include <net/if.h>
+#include <errno.h>
 
 /* Headers required by PAPI */
 #include "papi.h"
@@ -39,7 +40,7 @@ papi_vector_t _net_vector;
  ********************************************************************/
 
 /* Network stats refresh latency in usec (default: 1 sec) */
-#define NET_REFRESH_LATENCY   1000000
+static long long NET_REFRESH_LATENCY = 1000000;
 
 #define NET_PROC_FILE          "/proc/net/dev"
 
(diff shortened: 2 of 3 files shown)


src/components/net/CHANGES
```diff
@@ -16,3 +16,5 @@ Net component changelog:
 
  * Adds a couple of small tests/examples
 
+ * now NET_REFRESH_LATENCY can be modifed to the value of environment variable
```
Treece-Burgess: Here you are adding to the list of changes from 2011-11-07 by Jose Pedro Oliveira.

Can you create your own section above his list of changes? For example,
```
Net component changelog:

2026-03-10 your name

* Add support for environment variable PAPI_NET_REFRESH_LATENCY

2011-11-07 Jose Pedro Oliveira
.
.
.
```

src/components/net/CHANGES
```diff
@@ -16,3 +16,5 @@ Net component changelog:
```
Treece-Burgess, line 20: The README.md should be updated to disclose the addition of `PAPI_NET_REFRESH_LATENCY`.  I can do this myself as a final push or if you would like to you can as well. I would add "Configuring Refresh Latency" below the section "Enabling the NET Component":

\## Enabling the NET Component
.
.
.

\## Configuring Refresh Latency

By default the net refresh latency is set to 1000000 microseconds. To update this value set \`PAPI_NET_REFRESH_LATENCY\`, i.e:
\```
export PAPI_NET_REFRESH_LATENCY=your_value (in microseconds)
\```

src/components/net/CHANGES
```diff
@@ -16,3 +16,5 @@ Net component changelog:
 
  * Adds a couple of small tests/examples
```
Treece-Burgess: Can you squash commits and add a commit message indicating what this change does? If I run `git log` all I see is:
```
commit f7a5da6fcd58cf202ab4ba6fe8c0567faaca956f (HEAD -> master, origin/master, origin/HEAD)
Merge: e42911de4 a6bc4ab6e
Author: ajanicijamd <Aleksandar.Janicijevic@amd.com>
Date:   Mon Mar 9 18:42:26 2026 -0400

    Merge branch 'icl-utk-edu:master' into master
```

src/components/net/linux-net.c
```diff
@@ -367,6 +367,10 @@ _net_init_component( int cidx  )
     /* Export the component id */
     _net_vector.cmp_info.CmpIdx = cidx;
 
+    /* Set the net refresh latency */
+    char* refresh_latency = getenv("PAPI_NET_REFRESH_LATENCY");
+    if (refresh_latency != NULL) NET_REFRESH_LATENCY = atoll(refresh_latency);
```
Treece-Burgess: An issue I have with using `atoll` here is that if an error is encountered then `NET_REFRESH_LATENCY` will be set to 0. Further, there could be cases a user wants to set `PAPI_NET_REFRESH_LATENCY` equal to 0; therefore, not being able to guard against this.

I believe the better option is to use `strtoll` as so:
```c
    /* Set the net refresh latency */
    char *refresh_latency = getenv("PAPI_NET_REFRESH_LATENCY");
    if (refresh_latency != NULL) {
        char *endptr;
        int base = 10; 
        errno = 0;
        long long buffer_refresh = strtoll(refresh_latency, &endptr, base);
        if (errno != 0) {
            SUBDBG("strtoll failed with error code %d. Net refresh latency is the default (%lld).\n", errno, NET_REFRESH_LATENCY);
        }   
        else if (*endptr != '\0') {
            SUBDBG("PAPI_NET_REFRESH_LATENCY was not set properly. Net refresh latency is the default (%lld).\n", NET_REFRESH_LATENCY);
        }   
        else {
            NET_REFRESH_LATENCY = buffer_refresh;
        }   
    }
```

src/components/net/CHANGES
ajanicijamd: Done.

src/components/net/CHANGES
ajanicijamd: Done.

src/components/net/linux-net.c
ajanicijamd: Done. Thanks, I was not aware of SUBDBG for logging!

src/components/net/CHANGES
ajanicijamd: Done.

PR #590, opened 2026-04-16
rocp_sdk: honor ROCTX pause & resume

dbarry9:
## Pull Request Description

Prior to these changes, calls to roctxProfilerPause() and roctxProfilerResume() would be ignored by the rocp_sdk component. These changes acknowledge these calls by stopping counting on an event set following a pause and resuming counting following a resume.

These changes have been tested with the AMD MI250X and MI300A devices using ROCm 7.2.0.

Thank you to @adanalis-amd for his extensive help on this.

Useful Reference Materials:
- https://rocm.docs.amd.com/projects/rocprofiler-sdk/en/latest/how-to/using-rocprofiler-sdk-roctx.html
- https://rocm.docs.amd.com/projects/rocprofiler-sdk/en/latest/api-reference/rocprofiler-sdk-roctx_api_reference.html
- https://rocm.docs.amd.com/projects/rocprofiler-sdk/en/latest/api-reference/callback_services.html

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/rocp_sdk/sdk_class.cpp
+++ b/src/components/rocp_sdk/sdk_class.cpp
@@ -57,6 +57,8 @@ static std::condition_variable agent_cond_var = {};
 static bool data_is_ready = false;
 static std::string _rocp_sdk_error_string;
 static long long int *_counter_values = NULL;
+static long long int *_counter_values_savestate = NULL;
+static int _num_events_internal = 0;
 static int rpsdk_profiling_mode = RPSDK_MODE_DEVICE_SAMPLING;
 
 static agent_map_t gpu_agents = agent_map_t{};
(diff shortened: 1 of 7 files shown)


src/components/rocp_sdk/sdk_class.cpp
```diff
@@ -530,7 +539,33 @@ buffered_callback(rocprofiler_context_id_t,
...
+                      void* client_data) {
+    rocprofiler_context_id_t *ctx = static_cast<rocprofiler_context_id_t*>(client_data);
+
+    if(record.phase == ROCPROFILER_CALLBACK_PHASE_ENTER &&
+       record.kind == ROCPROFILER_CALLBACK_TRACING_MARKER_CONTROL_API &&
+       record.operation == ROCPROFILER_MARKER_CONTROL_API_ID_roctxProfilerPause)
+    {
+        ROCPROFILER_CALL(rocprofiler_stop_context_FPTR(*ctx), "pausing profiling context");
+        SUBDBG("Received ROCTX call to pause context.\n");
+        active_event_set_ctx->state = RPSDK_AES_STOPPED;
```
dbarry9, line 559: @adanalis-amd I wanted to double-check with you: Is setting the active event set's state necessary here?

adanalis-amd: active_event_set_ctx->state is only checked by record_callback() and read_sample(). The former won't be called anyway if you have stopped the context, so you don't need to care. However, read_sample() is triggered by PAPI_read(). If you don't guard with something like this flag, it might error further down. Note though, that the way the check in read_sample() is right now, it will also return an error code. Whether this is intended behavior is debatable. If the user code tries to PAPI_read() when profiling is paused you could legitimately return an error. If you don't want PAPI_read() to return an error, introduce a RPSDK_AES_PAUSED define, set "state" to that when pausing, and check for that in read_sample() and return success.

dbarry9: Thank you for your suggestion. I have addressed this in the latest push.

dbarry9:
@adanalis-amd This PR is ready for your review.

adanalis-amd (approved):

Treece-Burgess (approved):

PR #603, opened 2026-05-20
rocp_sdk: thread safety and multithreaded component test

dbarry9:
## Pull Request Description

Add a component test that uses multiple threads.

These changes have been tested on the AMD MI300A architecture using ROCm version 7.2.0.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/rocp_sdk/rocp_sdk.c
+++ b/src/components/rocp_sdk/rocp_sdk.c
@@ -18,6 +18,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <dirent.h>
+#include <stdatomic.h>
 #include "papi.h"
 #include "papi_internal.h"
 #include "papi_vector.h"
@@ -25,6 +26,7 @@
 #include "extras.h"
 #include "sdk_class.h"
 #include "rocprofiler-sdk/hsa.h"
+#include "rocprofiler-sdk/fwd.h"
 
 #define ROCPROF_SDK_MAX_COUNTERS (64)
 #define RPSDK_CTX_RUNNING (1)
@@ -41,6 +43,8 @@
 /* Utility functions */
 static int check_for_available_devices(char *err_msg);
 
+atomic_int tid_lock = 0;
+static int occupied = 0;
 unsigned int _rocp_sdk_lock;
 
 /* Init and finalize */
@@ -331,29 +335,46 @@ rocp_sdk_start(hwd_context_t *ctx, hwd_control_state_t *ctl)
     rocp_sdk_context_t *rocp_sdk_ctx = (rocp_sdk_context_t *) ctx;
     rocp_sdk_control_t *rocp_sdk_ctl = (rocp_sdk_control_t *) ctl;
 
-    if (0 == rocp_sdk_ctl->num_events){
-        SUBDBG("Error! Cannot PAPI_start an empty eventset.");
-        return PAPI_ENOSUPP;
-    }
+    // Only one thread can have an active profiling context, so we atomically
+    // set this to be the first thread that calls PAPI_start().
+    rocprofiler_thread_id_t my_tid;
+    rocprofiler_sdk_get_thread_id(&my_tid);
 
-    if (rocp_sdk_ctx->state & RPSDK_CTX_RUNNING) {
-        SUBDBG("Error! Cannot PAPI_start more than one eventset at a time for every component.");
-        return PAPI_EINVAL;
-    }
+    int set = atomic_compare_exchange_strong_explicit(&tid_lock, &occupied, my_tid, memory_order_acq_rel, memory_order_acquire);
 
-    if ( !(rocp_sdk_ctl->vendor_ctx) ) {
-        papi_errno = rocprofiler_sdk_ctx_open(rocp_sdk_ctl->events_id, rocp_sdk_ctl->num_events, &rocp_sdk_ctl->vendor_ctx);
+    if(tid_lock == my_tid) {
+
+        // Internally track the thread that has an active profiling context.
+        rocprofiler_sdk_set_control_thread_id(my_tid);
+
+        if (0 == rocp_sdk_ctl->num_events){
+            SUBDBG("Error! Cannot PAPI_start an empty eventset.");
+            return PAPI_ENOSUPP;
+        }
+
+        if (rocp_sdk_ctx->state & RPSDK_CTX_RUNNING) {
+            SUBDBG("Error! Cannot PAPI_start more than one eventset at a time for every component.");
+            return PAPI_EINVAL;
+        }
+
+        if ( !(rocp_sdk_ctl->vendor_ctx) ) {
+            papi_errno = rocprofiler_sdk_ctx_open(rocp_sdk_ctl->events_id, rocp_sdk_ctl->num_events, &rocp_sdk_ctl->vendor_ctx);
+            if (papi_errno != PAPI_OK) {
+                goto fn_fail;
+            }
+        }
+
+        papi_errno = rocprofiler_sdk_start(rocp_sdk_ctl->vendor_ctx);
         if (papi_errno != PAPI_OK) {
             goto fn_fail;
         }
+
+        rocp_sdk_ctx->state |= RPSDK_CTX_RUNNING;
+        goto fn_exit;
     }
 
-    papi_errno = rocprofiler_sdk_start(rocp_sdk_ctl->vendor_ctx);
-    if (papi_errno != PAPI_OK) {
-        goto fn_fail;
-    }
-
-    rocp_sdk_ctx->state |= RPSDK_CTX_RUNNING;
+    // The thread that could not get an active profiling context gets an error.
+    return PAPI_ENOSUPP;
 
   fn_exit:
     return papi_errno;
@@ -369,16 +390,30 @@ rocp_sdk_stop(hwd_context_t *ctx, hwd_control_state_t *ctl)
     rocp_sdk_context_t *rocp_sdk_ctx = (rocp_sdk_context_t *) ctx;
     rocp_sdk_control_t *rocp_sdk_ctl = (rocp_sdk_control_t *) ctl;
 
-    papi_errno = rocprofiler_sdk_stop(rocp_sdk_ctl->vendor_ctx);
-    if (papi_errno != PAPI_OK) {
-        goto fn_fail;
+    // Get thread ID and compare to the thread ID of active profiling context.
+    rocprofiler_thread_id_t my_tid;
+    rocprofiler_sdk_get_thread_id(&my_tid);
+
+    if(my_tid == tid_lock) {
+        papi_errno = rocprofiler_sdk_stop(rocp_sdk_ctl->vendor_ctx);
+        if (papi_errno != PAPI_OK) {
+            goto fn_fail;
+        }
+
+        rocp_sdk_ctl->vendor_ctx = NULL;
     }
 
-    rocp_sdk_ctl->vendor_ctx = NULL;
-
   fn_exit:
-    rocp_sdk_ctx->state = 0;
-    return papi_errno;
+    if(my_tid == tid_lock) {
+        rocp_sdk_ctx->state = 0;
+
+        // Allow for other threads to control the profiling context.
+        occupied = 0;
+        return papi_errno;
+    }
+
+    // The thread that does not have an active profiling context gets an error.
+    return PAPI_ENOTRUN;
   fn_fail:
     goto fn_exit;
 }
@@ -386,8 +421,16 @@ rocp_sdk_stop(hwd_context_t *ctx, hwd_control_state_t *ctl)
 int
 rocp_sdk_read(hwd_context_t *ctx __attribute__((unused)), hwd_control_state_t *ctl, long long **val, int flags __attribute__((unused)))
 {
-    rocp_sdk_control_t *rocp_sdk_ctl = (rocp_sdk_control_t *) ctl;
-    return rocprofiler_sdk_ctx_read(rocp_sdk_ctl->vendor_ctx, val);
+    rocprofiler_thread_id_t my_tid;
+    rocprofiler_sdk_get_thread_id(&my_tid);
+
+    if(my_tid == tid_lock) {
+        rocp_sdk_control_t *rocp_sdk_ctl = (rocp_sdk_control_t *) ctl;
+        return rocprofiler_sdk_ctx_read(rocp_sdk_ctl->vendor_ctx, val);
+    }
+
+    // The thread that does not have an active profiling context gets an error.
+    return PAPI_ENOTRUN;
 }
 
 int
--- a/src/components/rocp_sdk/tests/kernel.cpp
+++ b/src/components/rocp_sdk/tests/kernel.cpp
@@ -7,11 +7,13 @@ gemm(double *A, double *B, double *C, int N)
 {
     int colIdx = blockDim.x*blockIdx.x + threadIdx.x;
     int rowIdx = blockDim.y*blockIdx.y + threadIdx.y;
+    const double extraVal = 1.0002301;
 
     if( rowIdx < N && colIdx < N ) {
         for(int k = 0; k < N; ++k) {
             C[rowIdx*N + colIdx] += A[rowIdx*N + k] * B[k*N + colIdx];
         }
+        C[rowIdx*N + colIdx] += extraVal;
     }
 }
 
--- a/src/components/rocp_sdk/tests/kernel.h
+++ b/src/components/rocp_sdk/tests/kernel.h
@@ -10,11 +10,27 @@ do {                                                                           \
     }                                                                          \
 } while (0)
 
+#define PAPI_CALL_THD(call)                                                    \
+do {                                                                           \
+    int _status = call;                                                        \
+    if (_status != PAPI_OK) {                                                  \
+        status[id] = _status;                                                  \
+    }                                                                          \
+} while (0)
+
 #define HIP_CALL(call)                                                         \
 do {                                                                           \
     hipError_t err = call;                                                     \
     if(err != hipSuccess) {                                                    \
         test_fail(__FILE__, __LINE__, hipGetErrorString(err), PAPI_EMISC);     \
+    }                                                                          \
+} while(0)
+
+#define HIP_CALL_THD(call)                                                     \
+do {                                                                           \
+    hipError_t err = call;                                                     \
+    if(err != hipSuccess) {                                                    \
+        status[id] = PAPI_EMISC;                                               \
     }                                                                          \
 } while(0)
 
--- a/src/components/rocp_sdk/tests/multi_pthreads.cpp
+++ b/src/components/rocp_sdk/tests/multi_pthreads.cpp
@@ -1,0 +1,336 @@
+// Standard library headers
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h> 
+#include <pthread.h>
+
+// Internal headers
+#include "kernel.h"
+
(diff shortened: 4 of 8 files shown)


src/components/rocp_sdk/tests/multi_pthreads.cpp
```diff
@@ -0,0 +1,319 @@
...
+    if( 0 == total_event_count ) {
+        enumerate_and_store_rocp_sdk_native_events(&rocp_sdk_native_event_names, &total_event_count);
+    }
+
+    // Initialize barrier and lock.
+    pthread_barrier_init(&barrier, NULL, ONT);
+    pthread_mutex_init(&exclusive, NULL);
+
+    // Set thread function's arguments.
+    for(int thdIdx = 0; thdIdx < ONT; ++thdIdx) {
```
Treece-Burgess: Declare `thdIdx` outside of the for loop header.

src/components/rocp_sdk/tests/multi_pthreads.cpp
```diff
@@ -0,0 +1,319 @@
...
+        }
+    }
+
+    // Check if there was an error in any of the threads.
+    if( 0 != overall_status ) {
+        test_fail(__FILE__, __LINE__, "Encountered an error in one of the threads", PAPI_EMISC);
+    }
+
+    // Free dynamically allocated memory.
+    for (int i = 0; i < total_event_count; i++) {
```
Treece-Burgess: Declare `i` outside of the for loop header.

src/components/rocp_sdk/tests/multi_pthreads.cpp
```diff
@@ -0,0 +1,319 @@
...
+    for(int thdIdx = 0; thdIdx < ONT; ++thdIdx) {
+        pthread_join(tid[thdIdx], NULL);
+        if( status[thdIdx] != 0 ) {
+            overall_status = -1;
+        }
+    }
+
+    // Check if there was an error in any of the threads.
+    if( 0 != overall_status ) {
+        test_fail(__FILE__, __LINE__, "Encountered an error in one of the threads", PAPI_EMISC);
```
Treece-Burgess: nit: For myself, I avoid using `test_fail` unless it is associated with a PAPI function call, i.e.:
```
int retval = PAPI_add_named_event( ... )
if (retval != PAPI_OK) {
    test_fail(__FILE__, __LINE__, "PAPI_add_named_event", retval);
}
```

As currently the output appears as:
```
Line # 81 Error in Encountered an error in one of the threads: Unknown error code
```

I tend to use, `fprintf(stderr, "..."); exit(EXIT_FAILURE)`.

src/components/rocp_sdk/tests/multi_pthreads.cpp
```diff
@@ -0,0 +1,319 @@
...
+    HIP_CALL_THD(hipHostMalloc(&hostC, probSize, 0));
+
+    // Allocate device arrays.
+    HIP_CALL_THD(hipMalloc(&devA, probSize));
+    HIP_CALL_THD(hipMalloc(&devB, probSize));
+    HIP_CALL_THD(hipMalloc(&devC, probSize));
+
+    // Initialize arrays.
+    srandom(1);
+    for( int i = 0; i < N; i++ ) {
```
Treece-Burgess: Declare `i` and the below `j` outside of the for loop header.

src/components/rocp_sdk/tests/multi_pthreads.cpp
```diff
@@ -0,0 +1,319 @@
...
+
+    // Parse command-line arguments.
+    if (argc > 1) {
+        parse_and_assign_args(argc, argv, &rocp_sdk_native_event_names, &total_event_count);
+    }
+
+    // PAPI front matter.
+    int retval = PAPI_library_init(PAPI_VER_CURRENT);
+    if( retval != PAPI_VER_CURRENT ) {
+        test_fail(__FILE__, __LINE__, "PAPI_library_init() failed", retval);
```
Treece-Burgess, line 44: nit: I would drop failed. As it would read better due to the format shown in the comment on line 81.

src/components/rocp_sdk/tests/kernel.h
```diff
@@ -10,6 +10,14 @@ do {                                                                           \
     }                                                                          \
 } while (0)
 
+#define PAPI_CALL_THD(call)                                                    \
```
Treece-Burgess, line 13: `PAPI_CALL_THD` does not seem to be used.

src/components/rocp_sdk/tests/multi_pthreads.cpp
```diff
@@ -0,0 +1,319 @@
...
+    int N = 16;
+    dim3 threads_per_block( 1, 1, 1 );
+    dim3 blocks_in_grid( ceil( float(N) / threads_per_block.x ), ceil( float(N) / threads_per_block.y ), 1 );
+    size_t probSize = N*N*sizeof(double);
+
+    double *hostA=NULL,  *hostB=NULL,  *hostC=NULL;
+    double  *devA=NULL,   *devB=NULL,   *devC=NULL;
+
+    // Allocate host arrays.
+    HIP_CALL_THD(hipHostMalloc(&hostA, probSize, 0));
```
Treece-Burgess, line 209: Q: For `hipHostMalloc` and `hipMalloc` do you need to check if the variable you allocate memory for is equal to `NULL` or does the function call handle it for you?

src/components/rocp_sdk/tests/kernel.h
```diff
@@ -18,6 +26,14 @@ do {                                                                           \
     }                                                                          \
 } while(0)
 
+#define HIP_CALL_THD(call)                                                     \
```
Treece-Burgess, line 29: As stands, PAPI call's are "manually" checked and if the call fails `fprintf` is called with a message. Therefore, you can pinpoint quickly where the error is occurring. 

For HIP calls, the same is not done and we only collect the status. Consequently, if a HIP call failed we would have to go back and add `printf` calls to determine where the error occurred. 

You could always generalize these define's more such that they are:
```
#define HIP_CALL_THD(apiCall, threadId)
do {
    hipError_t _status = apiCall;
    if (_status != hipSuccess) {
        fprintf(stderr, "Thread ID %d: Call to %s failed with error code %d.\n", threadId, #apiCall, _status);
       status[threadId] = _status;
    }
} while (0)
```

src/components/rocp_sdk/tests/multi_pthreads.cpp
```diff
@@ -0,0 +1,319 @@
...
+        }
+        else
+        {
+            print_help_message(argv);
+            exit(EXIT_FAILURE);
+        }
+    }
+}
+
+void *thread_work(void *data) {
```
Treece-Burgess, line 163: Q: If a memory allocation, PAPI API call, or HIP API call fails, why continue on? As this leads to the previous error code being possibly overwritten as subsequent calls will fail.

Would it not be better to either entirely exit the program (does not cleanup everything) or in each thread if a call fails:
-> store status
-> goto cleanup
-> cleanup and return

src/components/rocp_sdk/tests/multi_pthreads.cpp
```diff
@@ -0,0 +1,319 @@
...
+
+    PAPI_CALL(PAPI_thread_init(pthread_self));
+
+    // If a user does not provide events, add them by enumerating available ones.
+    if( 0 == total_event_count ) {
+        enumerate_and_store_rocp_sdk_native_events(&rocp_sdk_native_event_names, &total_event_count);
+    }
+
+    // Initialize barrier and lock.
+    pthread_barrier_init(&barrier, NULL, ONT);
```
Treece-Burgess: Q: Should you not check the return value of `pthread_barrier_init`, `pthread_mutex_init`, `pthread_barrier_destroy`, `pthread_mutex_destroy`, and `pthread_barrier_wait`? From docs, they have the ability to fail.

src/components/rocp_sdk/tests/multi_pthreads.cpp
```diff
@@ -0,0 +1,319 @@
...
+
+    int EventSet = PAPI_NULL;
+    long long *values = (long long*)malloc(total_event_count/ONT*sizeof(long long));
+    if( NULL == values ) {
+        status[id] = PAPI_ENOMEM;
+    }
+
+    // Create event set.
+    int stat;
+    pthread_barrier_wait(&barrier);
```
Treece-Burgess: Q: For the PAPI calls, you setup `pthread_barrier_wait`, why do we need to wait at the barriers until all threads reach the barrier? Each thread appears that it will be self-contained.

src/components/rocp_sdk/tests/multi_pthreads.cpp
```diff
@@ -0,0 +1,319 @@
...
+    stat = PAPI_start(EventSet);
+    if( PAPI_OK != stat ) {
+        fprintf(stdout, "Thread %d: Failed to start counting. [%d]\n", id, stat);
+        status[id] = stat;
+    }
+    pthread_mutex_unlock(&exclusive);
+    pthread_barrier_wait(&barrier);
+
+    // Launch the GEMM five times.
+    for(int g = 0; g < NUMGEMMS; ++g) {
```
Treece-Burgess: Declare `g` outside of the for loop header.

src/components/rocp_sdk/tests/multi_pthreads.cpp
```diff
@@ -0,0 +1,319 @@
...
+
+    // Fork the threads.
+    for(int thdIdx = 0; thdIdx < ONT; ++thdIdx) {
+        pthread_create(&(tid[thdIdx]), NULL, thread_work, &(wkrdata[thdIdx]));
+    }
+
+    // Join the threads.
+    for(int thdIdx = 0; thdIdx < ONT; ++thdIdx) {
+        pthread_join(tid[thdIdx], NULL);
+        if( status[thdIdx] != 0 ) {
```
Treece-Burgess, line 77: nit: I would suggest differentiating `status` to be `papi_status` and `hip_status` (or `rocm_status`).

src/components/rocp_sdk/tests/kernel.cpp
```diff
@@ -7,11 +7,13 @@ gemm(double *A, double *B, double *C, int N)
 {
     int colIdx = blockDim.x*blockIdx.x + threadIdx.x;
     int rowIdx = blockDim.y*blockIdx.y + threadIdx.y;
+    const double extraVal = 1.0002301;
 
     if( rowIdx < N && colIdx < N ) {
         for(int k = 0; k < N; ++k) {
             C[rowIdx*N + colIdx] += A[rowIdx*N + k] * B[k*N + colIdx];
         }
+        C[rowIdx*N + colIdx] += extraVal;
```
Treece-Burgess, line 16: Q: What is the purpose of `extraVal`?

src/components/rocp_sdk/tests/multi_pthreads.cpp
```diff
@@ -0,0 +1,319 @@
...
+        // Call PAPI_read() after the GEMM.
+        pthread_barrier_wait(&barrier);
+        pthread_mutex_lock(&exclusive);
+        stat = PAPI_read(EventSet, values);
+        fprintf(stdout, "------ Thread %d ------  PAPI_read()\n", id);
+        if( PAPI_OK != stat ) {
+            fprintf(stdout, "\t<Error: Failed to read! %d>\n", stat);
+            status[id] = stat;
+        } else {
+            for(int i = id; i < total_event_count; i+=ONT) {
```
Treece-Burgess: Declare `i` outside of the for loop header.

Treece-Burgess (changes requested):

src/components/rocp_sdk/tests/kernel.cpp
dbarry9: This triggers an additional floating-point 'add' for each element in the matrix, as opposed to the GEMM only doing FMA operations, it will now have the FMAs and ADDs. This allows testing the two-threaded case with known events:
- rocp_sdk:::SQ_INSTS_VALU_FMA_F64:device=0
- rocp_sdk:::SQ_INSTS_VALU_FADD_F64:device=0

In short, it is for the sake of evaluating accuracy of this component test.

src/components/rocp_sdk/rocp_sdk.c
```diff
@@ -331,29 +335,46 @@ rocp_sdk_start(hwd_context_t *ctx, hwd_control_state_t *ctl)
...
+    // Only one thread can have an active profiling context, so we atomically
+    // set this to be the first thread that calls PAPI_start().
+    rocprofiler_thread_id_t my_tid;
+    rocprofiler_sdk_get_thread_id(&my_tid);
 
-    if (rocp_sdk_ctx->state & RPSDK_CTX_RUNNING) {
-        SUBDBG("Error! Cannot PAPI_start more than one eventset at a time for every component.");
-        return PAPI_EINVAL;
-    }
+    int set = atomic_compare_exchange_strong_explicit(&tid_lock, &occupied, my_tid, memory_order_acq_rel, memory_order_acquire);
```
adanalis-amd, line 343: Critical: the lock is never actually released, so it only works once

  In rocp_sdk_stop the release does:

  occupied = 0;

  but it never resets tid_lock:

  - Initial: tid_lock=0, occupied=0.
  - Thread A start: CAS(&tid_lock, &occupied, A) succeeds → tid_lock=A.
  - Thread A stop: sets occupied=0, returns. tid_lock is still A.
  - Thread B start later: CAS(&tid_lock, &occupied=0, B) compares tid_lock(A) vs expected 0 → fails → returns
  PAPI_ENOSUPP.

  After the first acquisition, no other thread can ever acquire the context. Only the original owner (whose tid
  is still stuck in tid_lock) can re-acquire. The release needs atomic_store(&tid_lock, 0) (leaving occupied as
  the constant-0 sentinel), not occupied = 0.

Occupied is misused as the CAS "expected" operand

  atomic_compare_exchange_strong writes the current value back into the expected pointer on failure. Since
  occupied is a shared global passed as expected:

  int set = atomic_compare_exchange_strong_explicit(&tid_lock, &occupied, my_tid, ...);

  every failed CAS clobbers occupied with the winner's tid. And because it's a plain int shared across threads,
  concurrent CAS calls read/write it non-atomically -> a genuine data race (UB in C11). The standard pattern is a
  local int expected = 0; per call. occupied shouldn't exist as a shared variable at all.

src/components/rocp_sdk/rocp_sdk.c
```diff
@@ -331,29 +335,46 @@ rocp_sdk_start(hwd_context_t *ctx, hwd_control_state_t *ctl)
...
+    if(tid_lock == my_tid) {
 
-    if ( !(rocp_sdk_ctl->vendor_ctx) ) {
-        papi_errno = rocprofiler_sdk_ctx_open(rocp_sdk_ctl->events_id, rocp_sdk_ctl->num_events, &rocp_sdk_ctl->vendor_ctx);
+        // Internally track the thread that has an active profiling context.
+        rocprofiler_sdk_set_control_thread_id(my_tid);
+
+        if (0 == rocp_sdk_ctl->num_events){
+            SUBDBG("Error! Cannot PAPI_start an empty eventset.");
+            return PAPI_ENOSUPP;
```
adanalis-amd, line 352: The thread has already claimed the lock, but several paths (like this return) bail out without releasing. The same for the goto fn_fail further down.

src/components/rocp_sdk/tests/multi_pthreads.cpp
```diff
@@ -0,0 +1,297 @@
...
+    HIP_CALL_THD(hipMalloc(&devB, probSize));
+    HIP_CALL_THD(hipMalloc(&devC, probSize));
+
+    // Initialize arrays.
+    srandom(1);
+    for( i = 0; i < N; i++ ) {
+        for( j = 0; j < N; j++ ) {
+            hostA[i*N + j] = ((double)random())/RAND_MAX + 1.1;
+            hostB[i*N + j] = ((double)random())/RAND_MAX + 1.1;
+            devC[i*N + j] = 0.0;
```
adanalis-amd: Host dereference of a device pointer. It should be hostC[]

src/components/rocp_sdk/rocp_sdk.c
```diff
@@ -41,6 +43,8 @@
 /* Utility functions */
 static int check_for_available_devices(char *err_msg);
 
+atomic_int tid_lock = 0;
```
adanalis-amd: tid_lock can be static. It doesn't need to be a visible symbol with external linkage.

src/components/rocp_sdk/tests/multi_pthreads.cpp
dbarry9: I have removed the barrier and mutex, as they are not needed.

src/components/rocp_sdk/tests/multi_pthreads.cpp
dbarry9: I have removed the barrier and mutex, as they are not needed.

src/components/rocp_sdk/tests/kernel.h
dbarry9: It is now removed.

src/components/rocp_sdk/rocp_sdk.c
dbarry9: I have updated this accordingly in the latest push.

src/components/rocp_sdk/tests/multi_pthreads.cpp
dbarry9: I have now addressed this in the latest push.

src/components/rocp_sdk/tests/multi_pthreads.cpp
dbarry9: Addressed in latest push.

src/components/rocp_sdk/tests/multi_pthreads.cpp
dbarry9: Addressed in latest push.

src/components/rocp_sdk/tests/multi_pthreads.cpp
dbarry9: This would work, too. In the current workflow, the exact error code isn't the important part because I print the nature of the error as it occurs. To avoid having to make multiple calls to a cleanup segment, I simply evaluate the presence of _some_ error after the threads have merged. We could do what you suggest, but I think either way would be okay.

src/components/rocp_sdk/tests/multi_pthreads.cpp
dbarry9: Per [this thread](https://github.com/icl-utk-edu/papi/pull/603#discussion_r4019788459), I think it wouldn't be necessary to have a different status for each.

src/components/rocp_sdk/tests/multi_pthreads.cpp
dbarry9: The function call handles it, and it will return `hipErrorOutOfMemory` (`!= hipSuccess`) in this case.

src/components/rocp_sdk/tests/kernel.h
dbarry9: Per this guidance, I have made a change that does something similar in the latest push.

src/components/rocp_sdk/rocp_sdk.c
dbarry9: I have addressed this in the latest push.git dif

src/components/rocp_sdk/rocp_sdk.c
dbarry9: I have addressed this in the latest push.

PR #608, opened 2026-05-21
rocp_sdk: sync event count before read

dbarry9:
## Pull Request Description

Previously, there was no guarantee that the record_callback() function completed before the counters are accessed in rocp_sdk_read(). These changes introduce a lock to ensure this sequence is honored. This change only applies to dispatch mode.

These changes have been tested on the AMD MI300A architecture using ROCm version 7.2.0.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/rocp_sdk/sdk_class.cpp
+++ b/src/components/rocp_sdk/sdk_class.cpp
@@ -453,6 +453,8 @@ record_callback(rocprofiler_dispatch_counting_service_data_t dispatch_data,
 
     _papi_hwi_unlock(_rocp_sdk_lock);
 
+    agent_mutex.unlock();
+
     return;
 }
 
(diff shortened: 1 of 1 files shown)


src/components/rocp_sdk/sdk_class.cpp
```diff
@@ -1292,17 +1294,22 @@ rocprofiler_sdk_ctx_read(vendorp_ctx_t ctx, long long **counters)
         if (papi_errno != PAPI_OK) {
             return papi_errno;
         }
+    } else {
```
adanalis-amd: since further down you check if the mode is RPSDK_MODE_DISPATCH (instead of !SAMPLING), change this to else if() for consistency. This way, if you introduce a third mode in the future, you won't end up deadlocking.

adanalis-amd (changes requested):

src/components/rocp_sdk/sdk_class.cpp
dbarry9: I have now addressed this in the latest push.

adanalis-amd (approved):

Treece-Burgess (approved):

PR #626, opened 2026-06-26
intel_gpu: update README with Xe drivers

dbarry9:
## Pull Request Description
Update the README to include the paranoid flag for the Xe drivers in addition to the i915 drivers.

Reference: [Intel Documentation](https://www.intel.com/content/www/us/en/docs/vtune-profiler/user-guide/2025-1/set-up-system-gpu-analysis.html)

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/intel_gpu/README.md
+++ b/src/components/intel_gpu/README.md
@@ -15,7 +15,7 @@ To enable intel_gpu component, the PAPI library is built with configure option a
 * [oneAPI Level Zero loader (libze_loader.so)](https://github.com/oneapi-src/level-zero)
 * [Intel(R) Metrics Discovery Application Programming Interface (libigdmd.so)](https://github.com/intel/metrics-discovery)
 * [Intel(R) Metrics Library Application Programming Interface (libigdml.so)](https://github.com/intel/metrics-library)
-* /proc/sys/dev/i915/perf_stream_paranoid is set to `0`
+* `/proc/sys/dev/i915/perf_stream_paranoid` or `/proc/sys/dev/xe/observation_paranoid` is set to `0`
 * User needs to be added into Linux render/video groups
 
 ## Environment Variables


src/components/intel_gpu/README.md
```diff
@@ -15,7 +15,7 @@ To enable intel_gpu component, the PAPI library is built with configure option a
 * [oneAPI Level Zero loader (libze_loader.so)](https://github.com/oneapi-src/level-zero)
 * [Intel(R) Metrics Discovery Application Programming Interface (libigdmd.so)](https://github.com/intel/metrics-discovery)
 * [Intel(R) Metrics Library Application Programming Interface (libigdml.so)](https://github.com/intel/metrics-library)
-* /proc/sys/dev/i915/perf_stream_paranoid is set to `0`
+* `/proc/sys/dev/i915/perf_stream_paranoid` or `/proc/sys/dev/xe/observation_paranoid` is set to `0`
```
Treece-Burgess, line 18: I am able to verify the `/proc/sys/dev/i915/perf_stream_paranoid` path exists; however, I am unable to do the same for `/proc/sys/dev/xe/observation_paranoid` does. Do you know of a machine we have access to that this can be verified?

Irrespective of the above, would following Intel's wording be better? In their documentation they state:
- Set `dev.i915.perf_stream_paranoid` to 0.

Then the same could be said for Xe driver:
- Set `dev.xe.observation_paranoid` to 0.

So all in all the README change would be:
- If the i915 driver is in use set `dev.i915.perf_stream_paranoid` to 0, alternatively if the Xe driver is in use set `dev.xe.observation_paranoid` to 0.

Treece-Burgess (changes requested):

src/components/intel_gpu/README.md
dbarry9: I have confirmed the existence of the `/proc/sys/dev/xe/observation_paranoid` on Mage@Oregon (Intel Battlemage B580 device).

I have updated the PR according to your comment.

As for the documenting changing the variables, the same could be said for `perf_event_paranoid` for the CPU event access (_e.g._, `/proc/sys/kernel/perf_event_paranoid` can be set with `sysctl kernel.perf_event_paranoid`). Likewise, `/proc/sys/dev/i915/perf_stream_paranoid` and `/proc/sys/dev/xe/observation_paranoid` can be set _via_ `dev.i915.perf_stream_paranoid` and `dev.xe.observation_paranoid`, respectively. However, I think that to be consistent with how we document `perf_event_paranoid` at configure time, we might consider having similar configure-time warning messages for the `intel_gpu`, `cuda`, _etc._ components.

Treece-Burgess: Thank you for the confirmation of the existence of `/proc/sys/dev/xe/observation_paranoid`. I have verified this on Mage at Oregon now as well.

Do you mean the large banner that outputs if `perf_event_paranoid` is >= 3? If so, I agree.

src/components/intel_gpu/README.md
```diff
@@ -15,8 +15,9 @@ To enable intel_gpu component, the PAPI library is built with configure option a
 * [oneAPI Level Zero loader (libze_loader.so)](https://github.com/oneapi-src/level-zero)
 * [Intel(R) Metrics Discovery Application Programming Interface (libigdmd.so)](https://github.com/intel/metrics-discovery)
 * [Intel(R) Metrics Library Application Programming Interface (libigdml.so)](https://github.com/intel/metrics-library)
-* /proc/sys/dev/i915/perf_stream_paranoid is set to `0`
-* User needs to be added into Linux render/video groups
+* If the i915 driver is in use, then `/proc/sys/dev/i915/perf_stream_paranoid` must be set to `0` (`sysctl dev.i915.perf_stream_paranoid=0`).
```
Treece-Burgess: Adding comment for documentation:

You do not need `sysctl -w` to write the value. All is needed is elevated privileges (tested in a docker container with the `--privileged` flag):
```
sudo sysctl dev.i915.perf_stream_paranoid=0
```

Treece-Burgess (approved):

src/components/intel_gpu/README.md
dbarry9: Yes, that is correct.

PR #628, opened 2026-07-02
libpfm4: fix const-correctness with strpbrk()

kvernet:
Newer glibc headers provide const-correct overloads for strpbrk(). In pfmlib_parse_event(), the return value was assigned to a char * even though it was only used to test for NULL, causing a build failure with recent GCC/glibc versions. This change removes the unnecessary assignment while preserving the existing behavior.


dbarry9:
Hi @kvernet, PAPI pulls down the latest libpfm4 changes from upstream, so your changes in this pull request would eventually be overwritten if we merged it. Changes to libpfm4 must be submitted to the maintainers of libpfm4 [here](https://sourceforge.net/p/perfmon2/patches/).

kvernet:
Thanks for the clarification.

PR #637, opened 2026-08-05
cat compiler warn err

### LINKED ISSUE #636: [Issue]: compiler warning and error for CAT
### Issue Description

GCC produces the following warning:
```
dcache.c: In function ‘varyBufferSizes’:
dcache.c:382:30: warning: initializer-string for array of ‘char’ truncates NUL terminator but destination lacks ‘nonstring’ attribute (5 chars into 4 available) [-Wunterminated-string-initialization]
  382 |             char symbol[4] = "|/-\\";
      |                              ^~~~~~~
```
and the following error:
```
main.c: In function ‘parseArgs’:
main.c:1002:13: error: too many arguments to function ‘print_usage’; expected 0, have 1
 1002 |             print_usage(name);
      |             ^~~~~~~~~~~ ~~~~
In file included from main.c:10:
driver.h:31:6: note: declared here
   31 | void print_usage();
      |      ^~~~~~~~~~~
main.c:1081:9: error: too many arguments to function ‘print_usage’; expected 0, have 1
 1081 |         print_usage(name);
      |         ^~~~~~~~~~~ ~~~~
driver.h:31:6: note: declared here
   31 | void print_usage();
      |      ^~~~~~~~~~~
main.c:1107:9: error: too many arguments to function ‘print_usage’; expected 0, have 1
 1107 |         print_usage(name);
      |         ^~~~~~~~~~~ ~~~~
driver.h:31:6: note: declared here
   31 | void print_usage();
      |      ^~~~~~~~~~~
main.c: At top level:
main.c:1146:6: error: conflicting types for ‘print_usage’; have ‘void(char *)’
 1146 | void print_usage(char* name)
      |      ^~~~~~~~~~~
driver.h:31:6: note: previous declaration of ‘print_usage’ with type ‘void(void)’
   31 | void print_usage();
      |      ^~~~~~~~~~~
```

This is because the prototype for the function `print_usage` does not have any arguments.

### Operating System

Ubuntu 24.04.4 LTS (Noble Numbat)

### Architecture

x86_64

### CPU Architecture

Intel Xeon 6767P CPU (Granite Rapids arch.)

### Compiler

GCC 16.1.1

### PAPI Version

master

### Component

CAT

### Steps to Reproduce

1. ./configure --prefix=${INSTDIR} --with-components=""
2. make && make install
3. cd counter_analysis_toolkit
4. make ARCH=X86 PAPIDIR=${INSTDIR}

### Additional Information

_No response_

dbarry9:
## Pull Request Description
This pull request addresses a couple of compile-time warnings and errors, as well as a commit that makes consistent use of a macro. This addresses Issue #636.

- cat: fix compiler warning about string without NUL
- cat: replace hard-coded value with proper macro
- cat: add missing argument to function prototype

These changes have been tested on a system containing the Intel Granite Rapids architecture.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/counter_analysis_toolkit/dcache.c
+++ b/src/counter_analysis_toolkit/dcache.c
@@ -353,7 +353,7 @@ int varyBufferSizes(long long *values, double **rslts, double **counter, cat_par
              * All other upper bounds are set to the capacity of the cache, as observed per core.
              */
             if( llc_idx+1 == j ) {
-                nextCacheSize = 16LL*(hw_desc->dcache_size[llc_idx])/hw_desc->mmsplit;
+                nextCacheSize = FACTOR*(hw_desc->dcache_size[llc_idx])/hw_desc->mmsplit;
                 ptsToNextCache = hw_desc->pts_per_mm+1;
             } else {
                 nextCacheSize = hw_desc->dcache_size[j]/hw_desc->split[j];
--- a/src/counter_analysis_toolkit/dcache.c
+++ b/src/counter_analysis_toolkit/dcache.c
@@ -378,7 +378,7 @@ int varyBufferSizes(long long *values, double **rslts, double **counter, cat_par
 
         cnt=0;
         for(j=0; j<len; j++){
-            char symbol[4] = "|/-\\";
+            char symbol[4] = {'|','/','-','\\'};
             active_buf_len = bufSizes[j]/sizeof(uintptr_t);
             if( params.show_progress ){
                 printf("%c\b",symbol[j%4]);
--- a/src/counter_analysis_toolkit/driver.h
+++ b/src/counter_analysis_toolkit/driver.h
@@ -28,7 +28,7 @@ void trav_evts(evstock* stock, int pk, int* cards, int nevts, int selexnsize, in
 int perm(int n, int k);
 int comb(int n, int k);
 void testbench(char** allevts, int cmbtotal, hw_desc_t *hw_desc, cat_params_t params, int myid, int nprocs);
-void print_usage();
+void print_usage(char *name);
 static int parse_line(FILE *input, char **key, long long *value);
 static void read_conf_file(char *conf_file, hw_desc_t *hw_desc);
 static hw_desc_t *obtain_hardware_description(char *conf_file_name);


src/counter_analysis_toolkit/dcache.c
```diff
@@ -353,7 +353,7 @@ int varyBufferSizes(long long *values, double **rslts, double **counter, cat_par
              * All other upper bounds are set to the capacity of the cache, as observed per core.
              */
             if( llc_idx+1 == j ) {
-                nextCacheSize = 16LL*(hw_desc->dcache_size[llc_idx])/hw_desc->mmsplit;
+                nextCacheSize = FACTOR*(hw_desc->dcache_size[llc_idx])/hw_desc->mmsplit;
```
Treece-Burgess, line 356: Q: dcache.h sets the value of `FACTOR` to be 12LL. Therefore, this PR changes the multiplied value from 16LL to 12LL. Why would this change not lead to any adverse behaviors?

Treece-Burgess (changes requested):

src/counter_analysis_toolkit/dcache.c
dbarry9: FACTOR is only used to probe buffer sizes that exceed the last-level cache. In this case, it means that the largest buffer size that CAT will use exceeds the LLC by a factor of 12 instead of 16, but both are reasonably large-enough to well exceed the size of the LLC.  In other words, 12 is a "good-enough" value.

Treece-Burgess (approved):
Tested on Stravinsky at Oregon (Intel(R) Xeon(R) 6767P) with GCC 16.0.0. Both the compilation warning and compilation error now do not occur. Along with this, I have verified the wheel correctly proceeds with no segmentation fault occurring. Test output:

```
tburgess@stravinsky:~/daniel_fork/src/counter_analysis_toolkit$ ./cat_collect -in event_list.txt -out ./ -dcr -verbose
D-Cache Latencies:100%
D-Cache Read Benchmarks: Total:100%  Current test:100%
tburgess@stravinsky:~/daniel_fork/src/counter_analysis_toolkit$ git log
```

PR #644, opened 2026-08-11
rocp_sdk: guard against no devices due to driver

### LINKED ISSUE #643: [Issue]: rocp_sdk: inactive driver leads to segmentation fault
### Issue Description

If the `amdgpu` driver is not active, then no devices are detected by the PAPI `rocp_sdk` component. This results in a segmentation fault when the component tries to access a device.

I am using an AMD Developer Cloud instance that contains a single AMD MI300X GPU device, and I am using ROCm 7.14.0.

### Operating System

Ubuntu 26.04 LTS (Resolute Raccoon)

### Architecture

x86_64

### CPU Architecture

Intel Xeon Platinum 8568Y+

### Compiler

GCC 15.2.0

### PAPI Version

master

### Component

rocp_sdk

### Steps to Reproduce

1. git clone https://github.com/icl-utk-edu/papi.git
2. cd papi/src/
3. export PAPI_ROCP_SDK_ROOT=/opt/rocm
4. ./configure --prefix=/root/apps/papi --with-components="rocp_sdk"
5. make
6. make install

### Additional Information

_No response_

dbarry9:
## Pull Request Description
If the driver is not active, then no devices will be detected. This leads to an invalid reference, which results in a segmentation fault. This PR remedies this issue by checking for how many agents the ROCProfiler-SDK detected.

This PR addresses Issue #643.

These changes have been tested using ROCm 7.14.0 on a system containing the MI300X GPU architecture.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**

### DIFF
--- a/src/components/rocp_sdk/sdk_class.cpp
+++ b/src/components/rocp_sdk/sdk_class.cpp
@@ -1185,6 +1185,13 @@ rocprofiler_sdk_init(void)
         goto fn_fail;
     }
 
+    if( papi_rocpsdk::gpu_agents.empty() ) {
+        if( papi_rocpsdk::get_error_string().empty() ){
+            papi_rocpsdk::set_error_string("No compatible devices detected. Please ensure that the driver is active.");
+        }
+        papi_errno = PAPI_ECMP;
+        goto fn_fail;
+    }
     papi_rocpsdk::populate_event_list();
 
   fn_exit:


src/components/rocp_sdk/sdk_class.cpp
```diff
@@ -1185,6 +1185,13 @@ rocprofiler_sdk_init(void)
         goto fn_fail;
     }
 
+    if( papi_rocpsdk::gpu_agents.empty() ) {
```
Treece-Burgess, line 1188: This solution works.

However, Anthony already created the function`check_for_available_devices` in `rocp_sdk.c`. In this function, we explicitly check if `/sys/class/kfd/kfd/topology/nodes` exists. If it does not exist then no AMD devices are on the system.

When I triggered the segmentation fault in the docker container, I did not pass `--device /dev/kfd` and `--device /dev/dri` which from [documentation ](https://rocm.docs.amd.com/projects/install-on-linux/en/docs-6.4.0/how-to/docker.html) appears to be a requirement.

In the case of not providing `--device` (i.e. `docker run -it rocm/dev-ubuntu-22.04:7.14.0-full)` then `ls -l /dev/kfd` and `ls -l /dev/dri` shows:
```
ls: cannot access '/dev/kfd': No such file or directory
ls: cannot access '/dev/dri': No such file or directory
```

In the case of providing `--device` (i.e. `docker run -it --device /dev/kfd --device /dev/dri  rocm/dev-ubuntu-22.04:7.14.0-full`) then `ls -l /dev/kfd` and `ls -l /dev/dri` shows:
```
crw-rw-rw- 1 root 105 237, 0 Aug 11 16:38 /dev/kfd
total 0
crw-rw---- 1 root irc 226,   0 Aug 11 16:38 card0
crw-rw---- 1 root irc 226,   1 Aug 11 16:38 card1
crw-rw---- 1 root irc 226,   2 Aug 11 16:38 card2
crw-rw-rw- 1 root 105 226, 128 Aug 11 16:38 renderD128
crw-rw-rw- 1 root 105 226, 129 Aug 11 16:38 renderD129
```

Instead of checking if `gpu_agents` is empty should we extend `check_for_available_devices` to include the paths `/dev/kfd` and `/dev/dri`?

Possible implementation:
```
#include <stdio.h>
#include <stdlib.h>
#include <sys/stat.h>

int main()
{
    const char *directories[] = {"/sys/class/kfd/kfd/topology/nodes", "/dev/kfd", "/dev/dri", NULL};

    for (int i = 0; directories[i] != NULL; i++) {
        struct stat buffer;
        int retval = stat(directories[i], &buffer);
        if (retval != 0) {
            printf("The directory %s does not exist.\n", directories[i]);
            return PAPI_ECMP;
        }   
    }   

    return 0;
}
```

Treece-Burgess (changes requested):

src/components/rocp_sdk/sdk_class.cpp
dbarry9: > Instead of checking if gpu_agents is empty should we extend check_for_available_devices to include the paths /dev/kfd and /dev/dri?

That's a good point. To be sure, I just spun up the cloud instance with the `amdgpu` driver missing, and here is what I see:

**ls -l /dev/kfd**
```
crw-rw---- 1 root render 236, 0 Aug 12 00:31 /dev/kfd
```

**ls -l /dev/dri**
```
total 0
drwxr-xr-x 2 root root         80 Aug 12 00:31 by-path
crw-rw---- 1 root video  226,   0 Aug 12 00:31 card0
crw-rw---- 1 root render 226, 128 Aug 12 00:31 renderD128
```

Based on this, I am not sure that `card0` existing here is a strong-enough condition to check (as it _does_ exist, but the segfault still occurs).

PR #671, opened 2026-09-16
validation tests: prevent division by zero and fix asm

### LINKED ISSUE #456: testcase papi_ld_ins hang indefinitely on Neoverse N1
After updating to commit 7710a20ed72b the validation tests papi_ld_ins hang indefinitely when run on an ARM Neoverse N1 system.

This occurs in the execute_loads and execute_stores assembly routines introduced by that commit.  It appears that an infinite loop is caused by the loop counters that are part of the test not progressing as expected on Neoverse N1.

System:
    Architecture: ARMv8 (AArch64)
    Microarchitecture: Neoverse N1
    OS: Fedora rawhide (Linux 6.16.0-0.rc5.16.fc43.aarch64)

dbarry9:
## Pull Request Description

- validation tests: fix type mismatch

    The asm uses 'x4', which is a 64-bit register, but it is given the
    variable 'n', which is an int (only 32 bits). Due to this, only the
    lower 32 bits of the register are guaranteed to be valid.

- validation tests: prevent division by zero

    If the event counts from PAPI_stop() are zero, then there will be a
    division by zero.

These changes were tested on the ARM Cortex-A72 architecture.

Although I was not able to reproduce Issue #456, this pull request addresses a what could be the underlying issue.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**


src/validation_tests/load_store_testcode.c
```diff
@@ -1,7 +1,7 @@
 #include "testcode.h"
 
 /* Execute n stores */
-int execute_stores(int n) {
+int execute_stores(long long n) {
```
Treece-Burgess, line 4: Question: The pull request description states the `asm` uses `x4` and x* is for 64 bit wide access. Therefore, the argument type is changed to `long long`; however, could it be that willow intended it to be 32 bit wide access in the original pull request? Meaning the x* should be w*. (Essentially what leads you to believe it should be 64 over 32).

src/validation_tests/papi_ld_ins.c
```diff
@@ -177,6 +177,10 @@ int main(int argc, char **argv) {
 	}
```
Treece-Burgess, line 178: I want to mention here if `expected` is equal to 0 before either `execute_stores` or `execute_loads` then we hang as well.  This is due to how the do while loop in the assembly code is constructed. As the comparison check is done after the increment.

`expected` being 0 should only occur if a user or us outright change the value to 0, but it is something I wanted to note.

src/validation_tests/papi_ld_ins.c
```diff
@@ -177,6 +177,10 @@ int main(int argc, char **argv) {
...
 
 	expected=mmm_count*REPITITIONS;
+    if( expected <= 0 ) {
```
Treece-Burgess: Opening the actual source file the indention will be off:
```
        expected=mmm_count*REPITITIONS;
    if( expected <= 0 ) {
        fprintf(stderr, "%s:%d: Error: Expected event counts must be positive!\n", __FILE__, __LINE__);
        exit(EXIT_FAILURE);
    }
```
Tabs are needed instead of spaces.

src/validation_tests/papi_sr_ins.c
```diff
@@ -178,6 +178,10 @@ int main(int argc, char **argv) {
 	}
 
 	expected=mmm_count*REPITITIONS;
+    if( expected <= 0 ) {
```
Treece-Burgess: Likewise as above.

Treece-Burgess (changes requested):
## Setup
PR 671 was tested on Hopper1 (ARM Neoverse-V2) at Oregon. The PAPI configure used was: `./configure --prefix=$PWD/test-install`.

## Master branch testing
I am unable to induce an infinite loop (n = 20) for either `papi_ld_ins.c` or `papi_sr_ins.c`. Both tests pass without issue.

## Inducing the infinite loop
If I set `expected <= 0`

src/validation_tests/papi_ld_ins.c
dbarry9: Done.

src/validation_tests/papi_sr_ins.c
dbarry9: Done.

src/validation_tests/load_store_testcode.c
dbarry9: This is a good question. @willowec Can you weigh-in on this? I just want to make sure I don't gloss-over anything.

dbarry9: >(Essentially what leads you to believe it should be 64 over 32).

@Treece-Burgess I re-wrote the asm code using 32-bit registers:

```
/* Execute n stores */
int execute_stores(int n) {

#if defined(__aarch64__)

    __asm(  ".data\n"
        "stvar: .word 1 /* stvar in memory (32-bit) */\n"
        ".text\n"
        "   ldr x2, =stvar /* address of stvar (address regs stay 64-bit) */\n"
        "   mov w4, %w0    /* n, 32-bit */\n"
        "   mov w1, #0\n"
        "str_loop:\n"
        "   str w1, [x2] /* store 32-bit value into stvar */\n"
        "   add w1, w1, #1\n"
        "   cmp w1, w4\n"
        "   bne str_loop\n"
        :
        : "r" (n)
        : "cc" /* clobbered */
    );

    return 0;

#endif
    (void) n;
    return CODE_UNIMPLEMENTED;
}

/* Execute n loads */
int execute_loads(int n) {

#if defined(__aarch64__)

    __asm(  ".data\n"
        "ldvar: .word 1 /* ldvar in memory (32-bit) */\n"
        ".text\n"
        "   ldr x2, =ldvar /* address of ldvar (address regs stay 64-bit) */\n"
        "   mov w4, %w0    /* n, 32-bit */\n"
        "   mov w1, #0\n"
        "ldr_loop:\n"
        "   ldr w3, [x2] /* load 32-bit value from ldvar */\n"
        "   add w1, w1, w3\n"
        "   cmp w1, w4\n"
        "   bne ldr_loop\n"
        :
        : "r" (n)
        : "cc" /* clobbered */
    );

    return 0;

#endif
    (void) n;
    return CODE_UNIMPLEMENTED;
}
```

Comparing the results across 10 executions for both the `papi_ld_ins` and `papi_sr_ins`, there are no observable differences (0.00% error) between 32-bit to 64-bit in the output values for the presets `PAPI_LD_INS` and `PAPI_SR_INS`.

Testing was done on a Raspberry Pi 4 system, containing the ARM Cortex-A72 architecture.

**Conclusion:** From these results, there is no meaningful difference between 32-bit and 64-bit asm for the purposes of this validation test. Therefore, I think we can proceed with this PR.

PR #673, opened 2026-09-17
rocp_sdk: rm redundant -pthread and improve error checking

dbarry9:
## Pull Request Description

- rocp_sdk: remove redundant -pthread flag from test
    
    The flag -pthread is already used in this compiler command, making this
    second appearance redundant.

- rocp_sdk: improve error checking

    Invoke test_fail() strictly for failed PAPI library calls, print the
    name of the roctx call that failed, and differentiate between which
    buffer could not be allocated.
    
These changes were tested on a system containing the AMD MI300A architecture using ROCm 7.2.4.

## Author Checklist
- [ ] **Description**
- [ ] **Commits**
- [ ] **Tests**


src/components/rocp_sdk/tests/kernel.h
```diff
@@ -23,6 +23,8 @@ do {                                                                           \
     int _status = call;                                                        \
     if(_status != 0) {                                                         \
         test_fail(__FILE__, __LINE__, #call, _status);                         \
```
Treece-Burgess: In the pull request description it is stated to "rm test_fail in check for non-PAPI error"; therefore, should the `test_fail` here not be removed? As this `#define` is strictly for `roctx*` API calls.

src/components/rocp_sdk/tests/kernel.h
```diff
@@ -23,6 +23,8 @@ do {                                                                           \
...
+        fprintf(stderr, "%s:%d: Error: ROCTX call failed.\n", __FILE__, __LINE__); \
```
Treece-Burgess: I would suggest passing `#call` and also `_status` to `fprintf` such that you know the exact API call failing and what status code was produced by the API call.

src/components/rocp_sdk/tests/roctx_pause_resume.cpp
```diff
@@ -112,7 +112,8 @@ int main(int argc, char *argv[]) {
     long long *values = (long long*)malloc(total_event_count*sizeof(long long));
     long long *prevValues = (long long*)malloc(total_event_count*sizeof(long long));
     if( NULL == values || NULL == prevValues ) {
-        test_fail(__FILE__, __LINE__, "Failed to allocate memory for counter values.", PAPI_ENOMEM);
+        fprintf(stderr, "%s:%d: Error: Failed to allocate memory for counter values.\n", __FILE__, __LINE__);
```
Treece-Burgess: nit: An issue I see here is you check both `values` and `prevValues` against `NULL`; therefore, it would not be immediately clear if `values` or `prevValues` was `NULL` and so more debugging would be required.

I would suggest individually checking them against `NULL` or you could create a `CHECK_MEMORY_ALLOCATION` `#define` too.

src/components/rocp_sdk/tests/Makefile
```diff
@@ -51,7 +51,7 @@ two_eventsets: two_eventsets.o kernel.o helper_rocp_sdk.o
 	$(AMDCXX) -O2 -g -DNDEBUG $(GPUFLAGS) two_eventsets.o kernel.o helper_rocp_sdk.o -o two_eventsets $(LDFLAGS)
 
 simple_sampling: simple_sampling.o kernel.o helper_rocp_sdk.o
-	$(AMDCXX) -O2 -g -DNDEBUG $(GPUFLAGS) simple_sampling.o kernel.o helper_rocp_sdk.o -o simple_sampling $(LDFLAGS) -pthread
+	$(AMDCXX) -O2 -g -DNDEBUG $(GPUFLAGS) simple_sampling.o kernel.o helper_rocp_sdk.o -o simple_sampling $(LDFLAGS)
```
Treece-Burgess, line 54: Q: When you state the compiler command, I assume you mean `amdclang++`.  Is that correct?

If so on Odyssey if I look at the output for `amdclang++ --version` (/opt/rocm-7.2.4/lib/llvm/bin/amdclang++ --version)  and look at the linked configuration file `/opt/rocm-7.2.4/lib/llvm/bin/clang++.cfg` I see:
```
@rocm.cfg"
```

If I then look at `@rocm.cfg`, I do not see `-pthread`:
```
-Wl,--enable-new-dtags
--rocm-path='<CFGDIR>/../../..'
-frtlib-add-rpath
```
***
`-pthread` does appear to be added at configure to `LDFLAGS`:
```
7028   # check for intel_gpu or rocp_sdk component to determine if we need -lstdc++ in LDFLAGS
7029   if (test "x$comp" = "xintel_gpu" || test "x$comp" = "xrocp_sdk"); then
7030     LDFLAGS="$LDFLAGS -lstdc++ -pthread"
7031   fi
```

Which then `$(LDFLAGS)` is appended to the end of the compilation line for the `simple_sampling` target. So I do see it being redundant; however, I think it would be better long term to make these flags more modular based on component.

Treece-Burgess (changes requested):

src/components/rocp_sdk/tests/kernel.h
dbarry9: Looks like I forgot to remove this line in my haste. Thank you for catching this.

src/components/rocp_sdk/tests/Makefile
dbarry9: I agree with this. I think we should keep this in mind when refactoring the build system.

src/components/rocp_sdk/tests/roctx_pause_resume.cpp
dbarry9: Resolved in latest push.

Treece-Burgess (approved):
I tested this PR on Odyssey at Oregon with ROCm 7.2.4, the results are as follows:

- PAPI build: ✅ 
- `simple_sampling` compilation: ✅ (no redundant `-pthread` only comes from `LDFLAGS`)
- `simple_sampling` component test: ✅ (Additionally, all `rocp_sdk` component tests pass)
- PAPI utilities*: ✅

__*__  `papi_component_avail`, `papi_native_avail`, and `papi_command_line`

